PR #144, opened 2024-02-27
Added Script to create the CWSPR client

shailay:
## Problem
Currently we are manually copying the CWSPR client from VS code to LS. This is error prone.  By merging this change we can reduce the development time since every time we need to add new api, we need to copy paste it from VS code
## Solution
- I have added a script which generates clients for SigV4 and BearerToken using respective json file. 
- These are getting generated by running `npm run generateClients` or `npm compile`. 
- Also added these client to .gitignore since it gets generate during setup. 
- Updated the CONTRIBUTING.md for adding future client . 
-  All the apis in BearerToken.json has been taken from [here](https://prod.artifactbrowser.brazil.aws.dev/packages/AWSVectorConsolasRuntimeServiceModel/versions/1.0.3674.0/platforms/AL2_x86_64/flavors/DEV.STD.PTHREAD/smithyprojections/AWSVectorConsolasRuntimeServiceModel/aws-sdk-external-2022-11-11/c2j/codewhispererruntime-2022-11-11.normal.json) which is the smithy model for CWSPR
- Referred this script from VS code as they also follow the same approach- [link](https://github.com/aws/aws-toolkit-vscode/blob/ac14f3975eaed65e2f646290291ad7c67593a253/packages/toolkit/scripts/build/generateServiceClient.ts)

## License

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,4 @@
 **/node_modules/**
 **/out/**
-server/aws-lsp-codewhisperer/src/client/codewhispererclient.d.ts
-server/aws-lsp-codewhisperer/src/client/token/codewhispererclient.d.ts
+server/aws-lsp-codewhisperer/src/client/sigv4/codewhisperersigv4client.d.ts
+server/aws-lsp-codewhisperer/src/client/token/codewhispererbearertokenclient.d.ts
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,6 +4,6 @@ node_modules/
 out/
 **/bin/
 **/obj/
-server/aws-lsp-codewhisperer/src/client/sigv4/codewhispererclient.d.ts
-server/aws-lsp-codewhisperer/src/client/token/codewhispererclient.d.ts
+server/aws-lsp-codewhisperer/src/client/sigv4/codewhisperersigv4client.d.ts
+server/aws-lsp-codewhisperer/src/client/token/codewhispererbearertokenclient.d.ts
 **/*.md
--- a/server/aws-lsp-codewhisperer/.prettierignore
+++ b/server/aws-lsp-codewhisperer/.prettierignore
@@ -4,6 +4,6 @@ node_modules/
 out/
 **/bin/
 **/obj/
-src/client/codewhispererclient.d.ts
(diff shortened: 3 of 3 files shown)


server/aws-lsp-codewhisperer/script/generateServiceClient.ts
```diff
@@ -0,0 +1,242 @@
...
+
+/** When cloning aws-sdk-js, we want to pull the version actually used in package-lock.json. */
+function getJsSdkVersion(): string {
+    const json = fs.readFileSync(path.resolve(repoRoot, 'package-lock.json')).toString()
+    const packageLock = JSON.parse(json)
+
+    return packageLock['packages']['node_modules/aws-sdk']['version']
+}
+
+async function cloneJsSdk(dir: string): Promise<void> {
```
viktorsaws, line 42: Why is it needed to pull github repo in order to build the clients? Why can't package installed in node_modules can be used directly?

package.json
```diff
@@ -15,7 +15,8 @@
         "format": "prettier --write .",
         "format-staged": "npx pretty-quick --staged",
         "clean": "ts-node ./script/clean.ts",
-        "compile": "tsc --build --verbose && npm run postcompile --workspaces --if-present",
+        "compile": "npm run generateClients && tsc --build --verbose && npm run postcompile --workspaces --if-present",
+        "generateClients": " ts-node server/aws-lsp-codewhisperer/script/generateServiceClient.ts",
```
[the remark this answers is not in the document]
viktorsaws: I don't think `generateClients` belongs in root of the repository, since it's specific to CodeWhisperer server implementation. Could you move it to codewhisperer-server package please?

server/aws-lsp-codewhisperer/script/generateServiceClient.ts
justinmk3: > Why is it needed to pull github repo in order to build the clients? Why can't package installed in node_modules can be used directly?

It probably could. Using git (in aws-toolkit-vscode's build) might be some tech debt that never got cleaned up. Just needs some human attention.

PR #157, opened 2024-03-08
add diagnostics, cancel handler and telemetry for security scan

zixlin7:
## Problem
need to publish diagnostics, handle hover for security scan findings, add handler for cancel scan and publish the security scan telemetry event
## Solution
verified that diagnostics, hover and cancel works and telemetry event being sent as expected
Screenshot (10)

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
@@ -1,21 +1,40 @@
 import { Server } from '@aws/language-server-runtimes'
 import { CredentialsProvider } from '@aws/language-server-runtimes/out/features'
-import { SecurityScanRequestParams } from '@aws/language-server-runtimes/out/features/securityScan/securityScan'
 import { pathToFileURL } from 'url'
 import { CancellationToken, ExecuteCommandParams } from 'vscode-languageserver'
 import { ArtifactMap } from '../client/token/codewhispererbearertokenclient'
 import { CodeWhispererServiceToken } from './codeWhispererService'
 import { DependencyGraphFactory } from './dependencyGraph/dependencyGraphFactory'
-import { SecurityScanHandler } from './securityScan/securityScanHandler'
-import { SecurityScanResponseParams } from './securityScan/types'
+import { getSupportedLanguageId, supportedSecurityScanLanguages } from './languageDetection'
+import SecurityScanDiagnosticsProvider from './securityScan/securityScanDiagnosticsProvider'
+import { SecurityScanCancelledError, SecurityScanHandler } from './securityScan/securityScanHandler'
+import { SecurityScanRequestParams, SecurityScanResponseParams } from './securityScan/types'
 import { parseJson } from './utils'
+import { SecurityScanEvent } from './telemetry/types'
 
 export const SecurityScanServerToken =
     (service: (credentialsProvider: CredentialsProvider) => CodeWhispererServiceToken): Server =>
-    ({ credentialsProvider, workspace, logging, lsp }) => {
+    ({ credentialsProvider, workspace, logging, lsp, telemetry }) => {
         const codewhispererclient = service(credentialsProvider)
+        const diagnosticsProvider = new SecurityScanDiagnosticsProvider(lsp, logging)
+        const scanHandler = new SecurityScanHandler(codewhispererclient, workspace, logging)
         let jobStatus: string
-        const runSecurityScan = async (params: SecurityScanRequestParams) => {
+        const securityScanStartTime = performance.now()
+        let serviceInvocationStartTime = 0
+        const securityScanTelemetryEntry: Partial<SecurityScanEvent> = {
+            codewhispererCodeScanSrcPayloadBytes: 0,
+            codewhispererCodeScanSrcZipFileBytes: 0,
+            codewhispererCodeScanLines: 0,
+            duration: 0,
+            contextTruncationDuration: 0,
+            artifactsUploadDuration: 0,
+            codeScanServiceInvocationsDuration: 0,
+            result: 'Succeeded',
+            codewhispererCodeScanTotalIssues: 0,
+            codewhispererCodeScanIssuesWithFixes: 0,
+            credentialStartUrl: credentialsProvider.getConnectionMetadata()?.sso?.startUrl ?? undefined,
+        }
+        const runSecurityScan = async (params: SecurityScanRequestParams, token: CancellationToken) => {
             logging.log(`Starting security scan`)
             try {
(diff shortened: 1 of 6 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
```diff
@@ -67,12 +95,17 @@ export const SecurityScanServerToken =
...
+                scanHandler.throwIfCancelled(token)
                 /**
                  * Step 3:  Create scan job
                  */
+                serviceInvocationStartTime = performance.now()
                 const scanJob = await scanHandler.createScanJob(artifactMap, document.languageId.toLowerCase())
-                logging.log(`Created security scan job.`)
+                logging.log(`Created security scan job id: ${scanJob.jobId}`)
+                securityScanTelemetryEntry.codewhispererCodeScanJobId = scanJob.jobId
+                scanHandler.throwIfCancelled(token)
```
mj632, line 108: Do we need this here? We are already checking on line 117 just after jobStatus check.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
```diff
@@ -96,45 +130,83 @@ export const SecurityScanServerToken =
...
-                // Todo: notify client about scan job failure.
+                securityScanTelemetryEntry.result = 'Failed'
                 return {
                     result: {
                         status: 'Failed',
                     },
                     error,
                 } as SecurityScanResponseParams
+            } finally {
+                securityScanTelemetryEntry.duration = performance.now() - securityScanStartTime
```
mj632, line 169: We won't reach to this part since we are returning in `try/catch` only. Can we pull this up or create return object and return it after this?

zixlin7: we will reach finally even if we return from try/catch

server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
mj632: Yes, missed the pollScanJobStatus method call.

mj632 (approved):

server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
```diff
@@ -1,21 +1,40 @@
...
 export const SecurityScanServerToken =
     (service: (credentialsProvider: CredentialsProvider) => CodeWhispererServiceToken): Server =>
-    ({ credentialsProvider, workspace, logging, lsp }) => {
+    ({ credentialsProvider, workspace, logging, lsp, telemetry }) => {
         const codewhispererclient = service(credentialsProvider)
+        const diagnosticsProvider = new SecurityScanDiagnosticsProvider(lsp, logging)
+        const scanHandler = new SecurityScanHandler(codewhispererclient, workspace, logging)
         let jobStatus: string
-        const runSecurityScan = async (params: SecurityScanRequestParams) => {
+        const securityScanStartTime = performance.now()
```
ege0zcan, line 22: Why are we creating this constant here? Shouldn't this be created each time a security scan gets started and not when the server is initialized?

zixlin7: yep you are right, these need to be moved inside runSecurityScan, will do

zixlin7: updated

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    async validateDiagnostics(uri: string, e: TextDocumentContentChangeEvent) {
+        const path = URI.parse(uri).path
+        const currentDiagnostics = this.diagnostics.get(path)
+        if (!currentDiagnostics) {
+            return
+        }
+        const nextDiagnostics: Diagnostic[] = currentDiagnostics.map(diagnostic => {
+            if ((e as any).range.start.line > diagnostic.range.end.line) {
```
ege0zcan, line 46: How do we ensure the TextDocumentContentChangeEvent we receive is not of type {text:string} here? Is this IDE specific behaviour?

zixlin7: seems like if there is only text, that implies the range is the entire document. Will make some updates to cover this case. Thanks for catching

ege0zcan: If this is IDE specific and we don't want to handle cases where server receives the full updated text document contents rather than a range, we can maybe use a typecheck function like this:

```
function isTextDocumentContentChangeEventWithRange(
    e: TextDocumentContentChangeEvent
): e is TextDocumentContentChangeEvent & { range: Range } {
    return 'range' in e;
}
```

so the code would start with a type check and we would returns the original diagnostic if the change event doesn't contain the range:

```
if (!isTextDocumentContentChangeEventWithRange(e)) {
   return diagnostic
}

if (e.range.start.line > diagnostic.range.end.line) {...}
else{...}
...
```

ege0zcan:
```
seems like if there is only text, that implies the range is the entire document. Will make some updates to cover this case. Thanks for catching

```

That makes it easier then, nice

zixlin7: updated

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    async validateDiagnostics(uri: string, e: TextDocumentContentChangeEvent) {
+        const path = URI.parse(uri).path
+        const currentDiagnostics = this.diagnostics.get(path)
+        if (!currentDiagnostics) {
+            return
+        }
+        // range will be omitted if the change is for entire document
+        const changeStartLine = (e as any).range?.start?.line
```
ege0zcan, line 46: nit: to me this approach(https://github.com/aws/language-servers/pull/157#discussion_r1519751853) still looks cleaner but this is an opinion, not a blocker

ege0zcan (approved):

PR #159, opened 2024-03-12
Dotnet Transformation feature changes

shailay:
## Problem
This PR includes the logic for dotnet transformation feature. 

## Solution
Using CWSPR APIs for :
- CreateUploadURL
- StartTransform, 
 - GetTransform, 
 - CancelTransform. 
 
Logging for testing
Screenshot 2024-03-12 at 3 10 43 PM

## License


client/vscode/src/sampleCommandActivation.ts
```diff
@@ -14,3 +18,59 @@ export function logCommand(languageClient: LanguageClient) {
...
     }
 }
+
+export function qNetCommand(languageClient: LanguageClient) {
+    return async () => {
+        const request = {
+            command: 'aws/qNetTransform/startTransform',
+            SolutionRootPath: 'D:\\TestProjects-master\\TestProjects-master\\net40\\mvc\\MvcMusicStore.sln',
+            TargetFramework: 'net8.0',
+            SourceCodeFilePaths: [
```
kmile, line 28: Do these have to be relative to a `workspaceFolder` (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders)

I'm not sure the server will be able to get them otherwise? Or at least, the server shouldn't access the filesystem outside of the workspace to prevent unintended or unauthorized access to the user's filesystem.

client/vscode/src/sampleCommandActivation.ts
```diff
@@ -14,3 +18,59 @@ export function logCommand(languageClient: LanguageClient) {
         languageClient.info(`Client: The log command has been executed`)
     }
 }
+
+export function qNetCommand(languageClient: LanguageClient) {
+    return async () => {
+        const request = {
+            command: 'aws/qNetTransform/startTransform',
```
kmile, line 25: Why is this namespaced `qNetTransform`? Is this materially different than other transformations in shape or intent?

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+    QNetCancelTransformResponse,
+    QNetGetTransformPlanRequest,
+    QNetGetTransformPlanResponse,
+    QNetGetTransformRequest,
+    QNetGetTransformResponse,
+    QNetStartTransformRequest,
+    QNetStartTransformResponse,
+} from './models'
+import { cleanup, createZip, getSha256 } from './utils'
+import fetch = require('node-fetch')
```
kmile, line 25: This won't be able to be bundled in the browser. As a rule of thumb, we don't want to pin any server to a specific platform or architecture.

We can use https://github.com/sindresorhus/ky or other cross-platform fetch wrapper to handle this in a cross platform way.

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -0,0 +1,255 @@
+import path = require('path')
+import { Workspace } from '@aws/language-server-runtimes/out/features'
+import * as fs from 'fs'
+import * as os from 'os'
```
kmile, line 4: We have a `fs` wrapper in the `workspace` feature that will allow you to do this in a cross platform way:

https://github.com/aws/language-server-runtimes/blob/main/src/features/workspace.ts#L21

server/aws-lsp-codewhisperer/src/language-server/netTransform/utils.ts
```diff
@@ -0,0 +1,158 @@
+import * as archiver from 'archiver'
+import * as crypto from 'crypto'
+import * as fs from 'fs'
+import * as path from 'path'
```
kmile, line 4: I don't think any of these are cross platform (node, browser, etc). We should not import dependencies that will limit the whether this server or components of it can be bundled to target different runtimes.

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
kmile: Same goes for `path`

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+    constructor(client: CodeWhispererServiceToken, workspace: Workspace) {
+        this.client = client
+        this.workspace = workspace
+    }
+
+    async startTransformation(userInputrequest: QNetStartTransformRequest): Promise<QNetStartTransformResponse> {
+        try {
+            const uploadId = await this.preTransformationUploadCode(userInputrequest)
+            const request = getCWStartTransformRequest(userInputrequest, uploadId)
+            console.log('send request to start transform api: ' + JSON.stringify(request))
```
kmile, line 39: This will break the server since it's listening on `STDOUT` of the process. And no-one will be able to read it since the output is not directed to any terminal or file.

You can use `features.logging` instead (https://github.com/aws/language-server-runtimes/blob/main/src/features/logging.ts)

kmile, line 39: I think this will work in most clients, but I'm not actually sure whether this is a client or a server concern. Looking at this (https://github.com/microsoft/vscode-languageserver-node/issues/402#issuecomment-432609554) it looks like it's the client.

When in doubt, we should probably use the `logging` feature: https://github.com/aws/language-server-runtimes/blob/main/src/features/logging.ts to ensure this output goes to the client and can be inspected in the output channel.

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+            const errorMessage = (error as Error).message ?? 'Error in StartTransformation API call'
+            console.log(errorMessage)
+            throw new Error(errorMessage)
+        }
+    }
+
+    async preTransformationUploadCode(userInputrequest: QNetStartTransformRequest): Promise<string> {
+        let uploadId = ''
+        const randomPath = uuidv4().substring(0, 4)
+        const workspacePath = path.join(os.tmpdir(), randomPath)
```
kmile, line 53: There is a `workspace.fs.getTempDirPath` method that does this in a cross-platform way: 

https://github.com/aws/language-server-runtimes/blob/main/src/features/workspace.ts#L25C5-L25C19

server/aws-lsp-codewhisperer/src/language-server/netTransform/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+    const externalReferences = request.ProjectMetadata.flatMap(r => r.ExternalReferences)
+    return externalReferences
+        .filter(reference => reference.IncludedInArtifact)
+        .filter(
+            (reference, index) =>
+                index === externalReferences.findIndex(other => reference.AssemblyFullPath === other.AssemblyFullPath)
+        )
+}
+
+export async function zipArtifact(basePath: string): Promise<string> {
```
kmile, line 83: I don't think this will work in a cross platform way. Here is an example in the same package of zipping the workspace in a (mostly) platform-agnostic way. One of the key bits here is we can write the stream directly to S3 instead of going to disk first:

https://github.com/mj632/language-servers/blob/c32400b0345d31e79ef21570ce6f3d1069559638/server/aws-lsp-codewhisperer/src/language-server/dependencyGraph/dependencyGraph.ts#L151

I don't think the workspace zip from that server can be readily reused here (since we have the dependency graph built in, and I don't think we do that for transform), but it might be worth looking and potentially aligning with the filesystem handling there.

server/aws-lsp-codewhisperer/src/language-server/netTransform/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+            .directory(sourceDir, false)
+            .on('error', err => reject(err))
+            .pipe(stream)
+
+        stream.on('close', () => resolve())
+        archive.finalize()
+    })
+}
+
+async function writeReuqirmentJsonAsync(dir: string, fileContent: string) {
```
kmile, line 139: Typo?

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
```diff
@@ -18,7 +25,48 @@ export const NetTransformServerFactory: (
...
+                        const userInputrequest = params as QNetStartTransformRequest
+                        logging.log('prepare artifact for solution: ' + userInputrequest.SolutionRootPath)
+                        return await transformHandler.startTransformation(userInputrequest)
+                    }
+                    case 'aws/qNetTransform/getTransform': {
+                        const request = params as QNetGetTransformRequest
+                        logging.log('Calling getTransform request with job Id: ' + request.TransformationJobId)
+                        return await transformHandler.getTransformation(request)
+                    }
+                    case 'aws/qNetTransform/pollTransform': {
```
kmile, line 42: Reading the rest of the CR, but why would the client poll the transform? The server can report progress and partial results when the the server gets status, and that logic is owned by the server. The client doesn't have to worry about that.

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
```diff
@@ -18,7 +25,48 @@ export const NetTransformServerFactory: (
...
+                    case 'aws/qNetTransform/getTransformPlan': {
+                        const request = params as QNetGetTransformPlanRequest
+                        logging.log('Calling getTransformPlan request with job Id: ' + request.TransformationJobId)
+                        const transformationPlan = await transformHandler.getTransformationPlan(request)
+                        logging.log(
+                            'Transformation plan for job Id' + request.TransformationJobId + ' is ' + transformationPlan
+                        )
+                        return transformationPlan
+                    }
+                    case 'aws/qNetTransform/cancelTransform': {
```
kmile, line 60: LSP has built in cancellation, I don't think we need to add another request type for it: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#cancelRequest

Unless the cancellation might need some extra metadata to track?

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
```diff
@@ -18,7 +25,48 @@ export const NetTransformServerFactory: (
...
+            try {
+                const client = createService(credentialsProvider)
+                const transformHandler = new TransformHandler(client, workspace)
+                switch (params.command) {
+                    case 'aws/qNetTransform/startTransform': {
+                        const userInputrequest = params as QNetStartTransformRequest
+                        logging.log('prepare artifact for solution: ' + userInputrequest.SolutionRootPath)
+                        return await transformHandler.startTransformation(userInputrequest)
+                    }
+                    case 'aws/qNetTransform/getTransform': {
```
kmile, line 37: What is the purpose of `getTransform`? Is there a different response from `start` that are not the results?

server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export function getCWStartTransformResponse(
+    response: PromiseResult<CodeWhispererTokenUserClient.StartTransformationResponse, AWSError>,
+    uploadId: string
+): QNetStartTransformResponse {
+    return {
+        UploadId: uploadId,
+        TransformationJobId: response.transformationJobId,
+    }
```
kmile, line 86: Why does the client need to know these? Could the result of this method be the transformation result?

client/vscode/package.json
```diff
@@ -40,6 +40,10 @@
             {
                 "command": "helloWorld.log",
                 "title": "Hello World LSP - Log Command"
+            },
+            {
+                "command": "aws/qNetTransform/startTransform",
```
volodkevych, line 45: nit: can it follow the same pattern as above, with dots?

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
shailay: Ack. Will replace all console log with `logging`

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
volodkevych, line 1: It would be good to have tests.

client/vscode/package.json
huawenm: sure

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
shailay: Do you know whats the best way to use `path` which is cross platform ?

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
shailay: So the getTransformation is the polling api. This will give the status of transformation as Inprogess, Failed, etc . Based on that we show different progress on the client side.

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
shailay: This is the cwspr cancel apis. This will be used to cancel the ongoing transformation job using the `TransformationJobId`

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
shailay: When you are saying client does that mean Toolkit and server mean Language server?

We do need to poll the status of the job do display different status on the UI

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -0,0 +1,255 @@
...
+        return ''
+    }
+
+    async uploadArtifactToS3Async(fileName: string, resp: CreateUploadUrlResponse) {
+        const sha256 = getSha256(fileName)
+        const headersObj = this.getHeadersObj(sha256, resp.kmsKeyArn)
+        try {
+            const response = await fetch(resp.uploadUrl, {
+                method: 'PUT',
+                body: fs.readFileSync(fileName),
```
shailay, line 109: @kmile @ege0zcan i need a way to use ReadFileSync, Currently i dont see `workspace.fs.readFileSync` available. It only has `workspace.fs.readFile` read file. Any workaround for that?

kmile: Why does this need to be a synchronous call?

Wouldn't 

```
const body = await fs.readFile(fileName),
const response = await fetch(resp.uploadUrl, {
                method: 'PUT',
                body: body,
                 headers: headersObj,
            })
```

Have the same effect?

client/vscode/src/sampleCommandActivation.ts
huawenm: that is the official feature name we get from PM, there's another java transform but probably not using this LSP.

client/vscode/src/sampleCommandActivation.ts
huawenm: is lsp server process act as subprocess in the visual studio? the source code files are subset of workspace folder but we also need "references", its not necessary to be in the same folder (customer can chose the nuget source from anyplace on the local). Visual studio would definitely have the access. Is LSP process have same permission as visual studio?

server/aws-lsp-codewhisperer/src/language-server/netTransform/utils.ts
huawenm: will do

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
huawenm: how this partial results work? is there any example how client get notification/result from server side?

ege0zcan: One important note is that custom commands do not have support for partial results but they do have support for work done progress reporting. 

Here (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand) you can see that `interface ExecuteCommandParams extends WorkDoneProgressParams`. `WorkDoneProgressParams` includes a `workDoneToken` which is a token of type string or number. If a client generates and provides a workDoneToken(usually achieved by creating a random uuid) when making a request to the server, the server can report progress on this token to client. This mechanism can be used instead of registering a new custom command for polling the transform status

- LSP documentation for client initiated work done progress: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#clientInitiatedProgress
- LSP documentation for partial results: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#partialResults

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
shailay: This is not working as it its. Its giving me 400 error when i use `workspace.fs.readFile`

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
huawenm: There're two polling requests in our workflow. it seems that these polling requests may not be applicable to all of our polling scenarios. Here's our current workflow:

1. once receiving an accepted response from startTransformRequest, the client initiates a call to getJobStatus to poll the status.

2. When the jobStatus retrieved from the GET API transitions from "accepted" to "transforming," the backend service generates a transformPlan.

3. the client begins polling another API called getTransformPlan to update the plan object locally.Note: the plan is not static; it undergoes dynamic changes in both status and plan steps. Each time the plan changes, the updated plan needs to be propagated to the client.

4. the client continues polling getJobStatus to monitor the job status until it's marked as completed.

I think we can use the work done progress reporting as optimization to replace the getJobStatus polling. we have to make change on both the client and server sides. Given that we currently have a different set of PRs under review on the client side, can we take this effort in a new PR (both client and server) after closing the current one? We've already documented this proposed change in our ticket for reference.

server/aws-lsp-codewhisperer/src/language-server/netTransform/utils.ts
shailay: We have lot of usage of `fs` I dont see everything available in runtime. For example for `fs.writeFileSync` , this is something needed for our functionality

client/vscode/src/sampleCommandActivation.ts
kmile: Yes, the LSP process has the same permissions. I just wanted to make sure we limit filesystem access to no more than we need for the use case.

client/vscode/src/sampleCommandActivation.ts
kmile: The Java transform will be refactored to use LSP as well so I fear there will be a whole lot of duplication if we make every server specific to a single development environment and programming language.

Are there any .NET specific cases that can be moved to a specific transformer strategy while keeping the core generic?

PR #168, opened 2024-03-22
Added support for format, hover and onDidOpenTextDocument capabilities

aderende:
## Problem
YAML JSON LSP server needs to support more capabilities.

## Solution

- Added onDidOpenTextDocument, onHover, onDidFormatDocument capabilities support 
- Added tests for new capabilities

## Testing

- Added more tests for new capabilities
- ran `npm install && npm run compile && npm test`
- ran YAML JSON server in vscode instance and tested manually

## License

### DIFF
--- a/server/aws-lsp-yaml-json/src/language-server/testUtils.ts
+++ b/server/aws-lsp-yaml-json/src/language-server/testUtils.ts
@@ -1,4 +1,5 @@
 import { TextDocument } from 'vscode-languageserver-textdocument'
+import { TextDocumentItem } from 'vscode-languageserver-types'
 
 const JSON_COMPLETION_CONTENT = `{
     "greet"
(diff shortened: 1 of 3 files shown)


server/aws-lsp-yaml-json/src/language-server/testUtils.ts
```diff
@@ -1,23 +1,58 @@
 import { TextDocument } from 'vscode-languageserver-textdocument'
+import { TextDocumentItem } from 'vscode-languageserver-types'
 
-const JSON_COMPLETION_CONTENT = `{
+const json_completion_content = `{
```
viktorsaws, line 4: nit: is there a reason for using different letters casing for constants in this file?

server/aws-lsp-yaml-json/src/language-server/yamlJsonServer.ts
```diff
@@ -59,9 +80,56 @@ export const YamlJsonServerFactory =
...
+        const onFormatHandler = async (params: DocumentFormattingParams): Promise<TextEdit[] | null> => {
+            const getTextDocumentFullRange = (textDocument: TextDocument): Range => {
+                return {
+                    start: {
+                        line: 0,
+                        character: 0,
+                    },
+                    end: {
+                        line: textDocument.lineCount,
+                        character: 0,
```
ege0zcan, line 111: does this mean we exclude the last line because we set end index to character 0?

server/aws-lsp-yaml-json/src/language-server/testUtils.ts
aderende: I've just made name in lower case for all constants that are not exported from this file. If it looks ugly, I cat revert it

viktorsaws: It's up to you, I just think consistency is better

server/aws-lsp-yaml-json/src/language-server/yamlJsonServer.ts
aderende: last line has index "textDocument.lineCount - 1". line: textDocument.lineCount is intentional here. 
This range include full document content

ege0zcan (approved):

PR #185, opened 2024-04-10
Added integration tests for Yaml/Json LSP Server

aderende:
## Problem
Required to add tests for YamlJson LSP Server to check real results for sam templates. 

## Solution

Added integration tests. Server is bundled with SAM json-schema to check what it returns for autocompletion and hover.

## License

### DIFF
--- a/app/aws-lsp-yaml-json-binary/package.json
+++ b/app/aws-lsp-yaml-json-binary/package.json
@@ -8,13 +8,22 @@
     },
     "scripts": {
         "compile": "tsc --build",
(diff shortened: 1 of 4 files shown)


app/aws-lsp-yaml-json-binary/src/tests/testFixture/completion.json
imykhai, line 1: Could we just create these empty files before the tests start and delete them afterwards?

aderende: Yep, we could
it makes sense

imykhai (approved):

PR #186, opened 2024-04-15
Add streaming client generation

ege0zcan:
## Problem
Introduces the codewhisperer streaming client as a usable module

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
         "app/*",
         "client/**",
         "core/*",
-        "server/*"
+        "server/*",
+        "server/**"
     ],
     "scripts": {
         "prepare": "husky install .husky",
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
@@ -4,6 +4,7 @@ import {
     ExecuteCommandParams,
     Server,
 } from '@aws/language-server-runtimes/server-interface'
+import { performance } from 'perf_hooks'
 import { pathToFileURL } from 'url'
 import { ArtifactMap } from '../client/token/codewhispererbearertokenclient'
 import { CodeWhispererServiceToken } from './codeWhispererService'
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ build
 .idea
 **/*.tgz
 .testresults/**
+server/aws-lsp-codewhisperer/src.gen/@amzn/codewhisperer-streaming/dist*
 
 # Visual Studio Related
 **/.vs
--- a/client/vscode/src/sso/builderId.ts
+++ b/client/vscode/src/sso/builderId.ts
@@ -122,7 +122,10 @@ export interface SsoConnection extends SsoProfile {
 }
 
 export class OidcClient {
-    public constructor(private readonly client: SSOOIDC, private readonly clock: { Date: typeof Date }) {}
+    public constructor(
+        private readonly client: SSOOIDC,
+        private readonly clock: { Date: typeof Date }
+    ) {}
 
     public async registerClient(request: RegisterClientRequest) {
         const response = await this.client.registerClient(request)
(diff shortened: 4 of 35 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
```diff
@@ -4,6 +4,7 @@ import {
     ExecuteCommandParams,
     Server,
 } from '@aws/language-server-runtimes/server-interface'
+import { performance } from 'perf_hooks'
```
volodkevych, line 7: Seems to be unused?

package.json
```diff
@@ -7,15 +7,17 @@
...
     ],
     "scripts": {
         "prepare": "husky install .husky",
         "lint": "eslint app/ client/ core/ script/ server/ --ext .ts,.tsx && prettier --check .",
         "format": "prettier --write .",
         "format-staged": "npx pretty-quick --staged",
         "clean": "ts-node ./script/clean.ts",
-        "compile": "tsc --build --verbose && npm run postcompile --workspaces --if-present",
+        "compile": "npm run steamingClient &&  tsc --build --verbose && npm run postcompile --workspaces --if-present",
+        "steamingClient": "npm run build -w server/aws-lsp-codewhisperer/src.gen/@amzn/codewhisperer-streaming",
```
volodkevych, line 20: Can we do `compile-streamingClient` (or something similar) here to expose the action?

volodkevych (approved):

package.json
```diff
@@ -7,15 +7,17 @@
         "app/*",
         "client/**",
         "core/*",
-        "server/*"
+        "server/*",
+        "server/aws-lsp-codewhisperer/src.gen/@amzn/codewhisperer-streaming"
```
viktorsaws, line 11: Does streaming client have to be defined as separate workspace in the top-level of monorepo? If the purpose is to reuse it across several sub-packages, I think it should be better stored in a separate dedicated location, rathen than CodeWhisperer LSP server.

package.json
viktorsaws, line 20: Can this build scripts be contained in the streaming client package and conform with setup on this monorepo? I'd prefer us to make packages introduced consistent with whole project setup, instead of introducing custom setup like this.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
ege0zcan: Actually it is the opposite, it is used but was not imported previously

package.json
ege0zcan: I think this is needed to ensure that the streaming client gets built before everything else given others will depend on it

package.json
ege0zcan: Currently all different codewhisperer servers live under `server/aws-lsp-codewhisperer` directory so I don't see it as a problem for now. 

Also given this is a non-ideal way of creating the streaming client that would hopefully be replaced with a proper way, keeping the mess contained here doesn't look that bad to me

viktorsaws: Seems like nested workspace could help here (I mean define `codewhisperer-streaming` as a workspace inside of `server/aws-lsp-codewhisperer` instead of root of monorepo), but idk if that works properly. There're few but reports open about it, but perhaps we could verify? 
https://github.com/npm/cli/issues/4774

package.json
viktorsaws: Looking at `compile` script, currently it does not trigger `compile` script in nested workspaces, only `postcompile` script does that, so whole project is not compiled by running `npm run compile` at root level.
Also I think just enabling `npm run compile -w` can become problematic, as npm does not build dependencies tree for workspaces, according to documentation.

But as you mentioned in thread above, if `codewhisperer-streaming` client is only meant to be used in `aws-lsp-codewhisperer` server, we don't need to expose it and it's commands at root level of this monorepo. Can this commands be moved to that package then?

server/aws-lsp-codewhisperer/src.gen/@amzn/codewhisperer-streaming/dist-cjs/CodeWhispererStreaming.js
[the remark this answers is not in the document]
rahmaniaam: I see if we run `npm run compile` in the CW server we're going to generate these folders again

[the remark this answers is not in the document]
ege0zcan: Good call out, let me update the gitignore so we don't commit these dist folders

PR #194, opened 2024-04-16
Added Chat client package

volodkevych:
## Problem
Created basic Chat client package with MynahUI in it. All the needed commands will be implemented in the following CRs. This CR implements only few events/commands, mainly as an example:
- `send to prompt`
- `focus`
- `tabs added/removed/updated`
- `ui ready`

Types for all the events/commands will be moved to separate package in following CRs.

`vscode` client was modified to load Chat client and trigger events/commands - mainly for testing purposes.

## License

### DIFF
--- a/chat-client/src/contracts/uiContracts.ts
+++ b/chat-client/src/contracts/uiContracts.ts
@@ -7,8 +7,8 @@ export const ERROR_MESSAGE = 'error-message'
 
 export type UiMessageCommand =
     | typeof UI_IS_READY
-    | typeof UI_FOCUS // TODO: Move to server
-    | typeof TAB_ID_RECEIVED // TODO: Move to server
+    | typeof UI_FOCUS
+    | typeof TAB_ID_RECEIVED
     | typeof SEND_TO_PROMPT
     | typeof AUTH_NEEDED_EXCEPTION
     | typeof ERROR_MESSAGE
@@ -23,7 +23,6 @@ export interface UiMessage extends Message {
 
 export type UiMessageParams = FocusParams | TabIdReceivedParams
 
-// TODO: This needs to be event to server as well, as it records metrics
 export interface FocusParams {
     type: string
 }
@@ -48,7 +47,6 @@ export interface SendToPromptParams {
     triggerId: string
 }
 
-// TODO: Use generics?
 export interface SendToPromptMessage {
     command: typeof SEND_TO_PROMPT
     params: SendToPromptParams


package.json
```diff
@@ -22,7 +23,7 @@
         "test-integ": "npm run compile && npm run test-integ --workspaces --if-present",
         "test-unit": "npm run compile && npm run test-unit --workspaces --if-present",
         "update-runtimes": "cp ../language-server-runtimes/aws-*.tgz bin/ && npm update @aws/language-server-runtimes",
-        "package": "npm install && npm run compile && npm run package-x64 --workspaces --if-present"
+        "package": "npm install && npm run compile && npm run package-x64 --workspaces --if-present && npm run package --workspaces --if-present"
```
rahmaniaam, line 26: nit: we should refactor all the packages to just use `npm run package`, just makes the script too long here

chat-client/src/client/chat.ts
```diff
@@ -0,0 +1,82 @@
...
+    UI_FOCUS,
+    UI_IS_READY,
+    UiMessage,
+} from '../contracts/uiContracts'
+import { Messager, OutboundChatApi } from './messager'
+import { InboundChatApi, createMynahUI } from './mynahUi'
+import { TabFactory } from './tabs/tabFactory'
+import { TabStorage } from './tabs/tabStorage'
+
+export const createChat = (clientApi: { postMessage: (msg: UiMessage | ServerMessage) => void }) => {
```
rahmaniaam, line 20: Not limited to this file, but can we add unit tests? does it make sense to have any for the chat client?

rahmaniaam (approved):

package.json
volodkevych: Maybe we should: I did not see any usage of `package` & `package-x64` in one package. But it's a very low priority. But it' a low priority.

chat-client/src/client/chat.ts
volodkevych: Yes, will do it as a followup.

chat-client/src/client/chat.ts
```diff
@@ -0,0 +1,82 @@
...
+                command: UI_IS_READY,
+            })
+
+            window.addEventListener('message', handleMessage)
+            window.addEventListener('focus', handleApplicationFocus)
+            window.addEventListener('blur', handleApplicationFocus)
+        },
+    }
+
+    const connector = new Messager(chatApi)
```
viktorsaws, line 78: super minor nit: naming could be made consistent, everywhere else this is referred to as `messager`

chat-client/src/client/chat.ts
```diff
@@ -0,0 +1,82 @@
...
+import { InboundChatApi, createMynahUI } from './mynahUi'
+import { TabFactory } from './tabs/tabFactory'
+import { TabStorage } from './tabs/tabStorage'
+
+export const createChat = (clientApi: { postMessage: (msg: UiMessage | ServerMessage) => void }) => {
+    // eslint-disable-next-line semi
+    let mynahApi: InboundChatApi
+
+    const sendMessageToClient = (message: UiMessage | ServerMessage) => {
+        clientApi.postMessage(message)
```
viktorsaws, line 25: What is referred to as `client` here?
If I read this correctly (correcty me if I'm wrong), this is connecting MynahUI Chat Client inside of webview with outside proxy over postMessage in the hosting application. If that's true, can this be renamed to something more explicit, `sendMessageToHost` perhaps?

I found it a bit confusing to understand the direction of messages flow, perhaps changing the name or adding some documentation could help here.

volodkevych: `clientApi` is `vscode` api if the extension is `vscode` extension. `postMessage` cannot be renamed to `sendMessageToHost`, `clientApi` can be renamed to whatever makes it more clear.

chat-client/src/client/tabs/tabStorage.ts
```diff
@@ -0,0 +1,53 @@
...
+    private tabs: Map<string, Tab> = new Map()
+    private lastSelectedTab?: Tab
+
+    public addTab(tab: Tab) {
+        if (this.tabs.has(tab.id)) {
+            return
+        }
+        this.tabs.set(tab.id, tab)
+        if (tab.isSelected) {
+            this.setSelectedTab(tab.id)
```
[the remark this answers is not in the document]
volodkevych: If tab with the same ID already exists, the request to add a tab (either selected or not selected) is ignored. This is the logic in current MynahUI. If we need to change it in the future, we will change it.

chat-client/src/client/mynahUi.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        messager.onSendToPrompt(params, tabId)
+    }
+
+    return {
+        sendToPrompt: sendToPrompt,
+    }
+}
+
+const uiComponentsTexts = {
```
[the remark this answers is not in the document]
volodkevych: Not sure at this point of time. If we need, we'll change that.

chat-client/package.json
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "@aws/q-chat-client",
```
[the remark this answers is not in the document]
volodkevych: I can rename if there are suggestions.

PR #200, opened 2024-04-22
Create Chat session service that manages API calls

makenneth:
## Problem

Creating `ChatSessionService` client that talks to the Q conversation API provided by `@amzn/code-whisperer-streaming`.  This service will be wrapped within a Management service that creates a mapping between tabId and a client `Map<string, ChatService>`, which will come in a separate PR

Also adding `aws-sdk-v3-proxy` package to handle http proxy for the streaming service since that uses sdk v3.

This client:
- exposes `abortingRequest` which allows LSP client to abort the API call
- stores conversation id and automatically appends to the request

## License


server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
```diff
@@ -0,0 +1,71 @@
...
+            retryStrategy: new ConfiguredRetryStrategy(0, (attempt: number) => 500 + attempt ** 10),
+        })
+
+        /**
+         * Streaming client use v3 sdk so there is no longer a global configuration for proxy
+         *
+         * addProxyToClient detects environment variables and determines whether or not to attach proxy
+         */
+
+        this.#client = addProxyToClient(steamingClient, { throwOnNoProxy: false })
```
viktorsaws, line 42: One note about using proxy: ideally servers developed in this repo should avoid using nodejs-specific libraries, to allow compiling and running the server as a  WebWorker in browser environments. `aws-sdk-v3-proxy` relies on NodeJS packages a lot, which would prevent the server to run in web.

This could be mitigated in future by introducing browser-compatible variants of this Service. 

In the past we have mitigated problem with running CodeWhisperer server behind proxy by exposing special "proxy" [variant of server](https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts).

But if that's currently not a goal to support browser, this should not be blocking from merging this PR

viktorsaws (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
makenneth: Thanks for the call out! I think I will just move the `addProxyToClient` out to the proxy variant wrapper

PR #206, opened 2024-04-24
Remove downlevel command and dependency from CodeWhisperer streaming client

viktorsaws:
## Problem
CodeWhisperer streaming client has unused dependency, which messes with our internal workflow.

## Solution

Script that uses dependency seem to be unused during build process. Remove it and problematic dependency from the project.

## License


makenneth (approved):
I tested in my local environment that removing this line doesn't impact the build at all

PR #207, opened 2024-04-24
Clear security scan findings on workspace/didChangeWorkspaceFolders event

shruti0085:
## Description
This change fixes a bug where when a project was unloaded in the IDE and a new one opened, the security scan findings(diagnostics) associated with the previous project did not get cleared.
This change adds handling for the `workspace/didChangeWorkspaceFolders` [lsp notification ](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWorkspaceFolders) sent by the client. The handler clears diagnostics associated with the previous security scan run(project).

Bonus: This change also converts the `resetDiagnostic` method to operate asynchronously.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
@@ -23,10 +23,10 @@ class SecurityScanDiagnosticsProvider {
         this.findings = []
     }
 
-    resetDiagnostics() {
-        this.findings.forEach(finding => {
-            this.logging.log(finding.filePath)
-            this.publishDiagnostics(finding.filePath, [])
+    async resetDiagnostics() {
+        this.findings.forEach(async finding => {
+            this.logging.log(`reset diagnostics for: ${finding.filePath}`)
+            await this.publishDiagnostics(finding.filePath, [])
         })
         this.diagnostics = new Map()
     }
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
```diff
@@ -23,10 +23,10 @@ class SecurityScanDiagnosticsProvider {
...
     }
 
-    resetDiagnostics() {
-        this.findings.forEach(finding => {
-            this.logging.log(finding.filePath)
-            this.publishDiagnostics(finding.filePath, [])
+    async resetDiagnostics() {
+        this.findings.forEach(async finding => {
+            this.logging.log(`reset diagnostics for: ${finding.filePath}`)
+            await this.publishDiagnostics(finding.filePath, [])
```
makenneth, line 29: What's the expected behavior of the `forEach` loop here? Is the function supposed to wait until all the publish diagnostics execution are complete?

If so, we may need to switch to `await Promise.all(this.findings.map(...))` . 
If not, the async/await keywords here doesn't change the behavior much; it will still fire multiple async call and move forward in the function

shruti0085: I don't think we want to for all findings publish diagnostic to finish but the intent is we want to wait for the resetDiagnostic call to finish before we return from the didChangeWorkspaceFolder handler

makenneth (approved):

zixlin7 (approved):

ege0zcan (approved):

PR #208, opened 2024-04-24
Chat server skeleton

makenneth:
## Description
- Create a chat server skeleton with a proxy variant
- Create session management service that manages the mapping between tabId and api client
  - Each tab can have an outgoing request at the same time (we will need to add a restriction there)
  - Each outgoing request can be cancelled separately

## Future Tasks
- Add more tests for the ChatServer as I work on the implementation
- Look into reducing the number of libraries needed for http proxy

## Testing
- Tested in vscode "Run and Debug" 
   - I can make a chat request  
   - I can make a request through a proxy 
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionManagementService.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionManagementService.ts
@@ -15,11 +15,11 @@ export class ChatSessionManagementService {
         this.#clientConfig = clientConfig
     }
 
-    hasSession(tabId: string): boolean {
+    public hasSession(tabId: string): boolean {
         return this.#sessionByTab.has(tabId)
     }
 
-    getSession(tabId: string): ChatSessionService {
+    public getSession(tabId: string): ChatSessionService {
         const maybeSession = this.#sessionByTab.get(tabId)
 
         if (!maybeSession) {
@@ -34,12 +34,12 @@ export class ChatSessionManagementService {
         return maybeSession
     }
 
-    deleteSession(tabId: string): void {
+    public deleteSession(tabId: string): void {
         this.#sessionByTab.get(tabId)?.dispose()
         this.#sessionByTab.delete(tabId)
     }
 
-    dispose(): void {
+    public dispose(): void {
         this.#sessionByTab.forEach(session => session.dispose())
     }
 }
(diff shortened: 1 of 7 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionManagementService.ts
```diff
@@ -0,0 +1,45 @@
...
+    public hasSession(tabId: string): boolean {
+        return this.#sessionByTab.has(tabId)
+    }
+
+    public getSession(tabId: string): ChatSessionService {
+        const maybeSession = this.#sessionByTab.get(tabId)
+
+        if (!maybeSession) {
+            const clientConfig = typeof this.#clientConfig === 'function' ? this.#clientConfig() : this.#clientConfig
+            const newSession = new ChatSessionService(this.#credentialsProvider, clientConfig)
```
makenneth, line 27: I am separating this part out into a createSession function in future work https://github.com/aws/language-servers/commit/1e0798053214ba66e4330ec1b6ff40f312f6c74b#diff-f6d1aac51c11d25d6c080c419e43328dee44efcd423b490c0f24b44453fcb841

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
```diff
@@ -0,0 +1,39 @@
+import {
+    CancellationToken,
+    ChatParams,
+    CredentialsProvider,
+    EndChatParams,
+    Server,
+} from '@aws/language-server-runtimes/server-interface'
+import { ChatSessionManagementService } from './chat/chatSessionManagementService'
+
+export const QChatServerToken =
```
ege0zcan, line 10: nit: for the time being we can call this QChatServer given there is no possibility to use streaming client with iam credentials

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionManagementService.ts
```diff
@@ -0,0 +1,45 @@
+import { CredentialsProvider } from '@aws/language-server-runtimes/server-interface'
+import { ChatSessionService, ChatSessionServiceConfig } from './chatSessionService'
+
+export class ChatSessionManagementService {
```
ege0zcan, line 4: Can this service be a singleton given there will probably be one chat server? 

For example for session management in inline code completions, we use a session manager singleton: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/session/sessionManager.ts#L220

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionManagementService.ts
makenneth: Yeah that's a great idea! I will make this a singleton, thank you

PR #209, opened 2024-04-25
Remove postinstall script to build dependencies

viktorsaws:
## Problem
Postinstall script in CodeWhisperer LSP runs build script of streaming client dependency package. This script fails when `@aws/lsp-codewhisperer` client is installed as a dependency due to missing transient dependencies at installation time.

This breaks workflows, that use `@aws/lsp-codewhisperer` as a direct dependency.

## Solution

Remove `postinstall` script from `server/aws-lsp-codewhisperer`. Compiling streaming client dependency is triggered in `compile` command still.

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,12 +17,14 @@
         "format": "prettier --write .",
         "format-staged": "npx pretty-quick --staged",
         "clean": "ts-node ./script/clean.ts",
-        "compile": "tsc --build --verbose && npm run postcompile --workspaces --if-present",
+        "precompile": "npm run precompile --workspaces --if-present",
+        "compile": "tsc --build --verbose && npm run compile:servers && npm run compile:apps",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -17,12 +17,14 @@
         "format": "prettier --write .",
         "format-staged": "npx pretty-quick --staged",
         "clean": "ts-node ./script/clean.ts",
-        "compile": "tsc --build --verbose && npm run postcompile --workspaces --if-present",
```
volodkevych, line 20: what about calling `postcompile` in the new version?

viktorsaws: `post` scripts are called by npm itself in nested workspaces correctly. It's only `precompile` that was causing problem.

makenneth (approved):

PR #210, opened 2024-04-25
Fix security scan duplicate hover message

rbbarad:
## Problem
Security Scan diagnostic error messages were displayed twice in the IDE hover.

## Solution

Removes the LSP hover message since the default hover displays the relevant information. 
- Future work to add additional content to this hover can redefine a hover handler - this will add, but not replace, content in the IDE's existing diagnostic hover.

Also changes the Diagnostic's Description to be constructed with the `DetectorName` rather than the `Title`. The `Title` field is a combination of the error code and the `DetectorName`, which was resulting in duplication of the error code

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
@@ -1,6 +1,6 @@
 import {
     Diagnostic,
-    Hover,
+    DiagnosticSeverity,
     Logging,
     Lsp,
     Position,
@@ -47,13 +47,15 @@ class SecurityScanDiagnosticsProvider {
     }
 
     mapScanIssueToDiagnostics(issue: CodeScanIssue): Diagnostic {
-        return Diagnostic.create(
-            this.createDiagnosticsRange(issue.startLine, issue.endLine),
-            `${issue.title} - ${issue.description.text}`,
-            2,
-            issue.relatedVulnerabilities.join(','),
-            'CodeWhisperer'
-        )
+        const diagnostic: Diagnostic = {
+            range: this.createDiagnosticsRange(issue.startLine, issue.endLine),
+            message: `${issue.detectorName} - ${issue.description.text}`,
+            severity: DiagnosticSeverity.Warning,
+            code: issue.relatedVulnerabilities.join(','),
+            source: 'CodeWhisperer',
+        }
+
+        return diagnostic
     }
 
     async validateDiagnostics(uri: string, e: TextDocumentContentChangeEvent) {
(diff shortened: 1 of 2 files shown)


shruti0085 (commented):
Please add @zixlin7  and @mj632 as reviewers on this PR

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
```diff
@@ -112,22 +114,16 @@ class SecurityScanDiagnosticsProvider {
...
 
+    /**
+     * A default hover element for diagnostics is surfaced in the IDE
+     * that displays as `${diagnostic.Code}: ${diagnostic.Description}`
+     *
+     * Creating a custom lsp hover will add content to the IDE's existing hover,
+     * but will not replace the above default display text.
+     */
     handleHover = () => {
         this.lsp.onHover(({ position, textDocument }) => {
```
shruti0085, line 125: Based on the comment, it seems like we do not need to handle this event then and want to let Visual Studio internally handle this. Can this handler be fully deleted?

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
```diff
@@ -112,22 +114,16 @@ class SecurityScanDiagnosticsProvider {
...
     }
 
+    /**
+     * A default hover element for diagnostics is surfaced in the IDE
+     * that displays as `${diagnostic.Code}: ${diagnostic.Description}`
+     *
+     * Creating a custom lsp hover will add content to the IDE's existing hover,
+     * but will not replace the above default display text.
+     */
     handleHover = () => {
```
shruti0085, line 124: Based on the comment, it seems like we do not need to handle this event then and want to let client internally handle this. Also is this behavior consistent with other clients i.e if more clients are integrated with this server?
Can this handler be fully deleted?

rbbarad: we could remove this - however I'd prefer to keep it. future work will likely involve customization here and I'd like to keep this note to avoid the case development for different clients could break one another

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
```diff
@@ -47,13 +47,15 @@ class SecurityScanDiagnosticsProvider {
...
-        )
+        const diagnostic: Diagnostic = {
+            range: this.createDiagnosticsRange(issue.startLine, issue.endLine),
+            message: `${issue.detectorName} - ${issue.description.text}`,
+            severity: DiagnosticSeverity.Warning,
+            code: issue.relatedVulnerabilities.join(','),
+            source: 'CodeWhisperer',
+        }
+
+        return diagnostic
```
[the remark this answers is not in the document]
rbbarad: Ah good point - I initially switched away from using `Diagnostic.create` to access parameters outside the scope of the method, but that's no longer necessary. I do find this syntax a bit more readable, but i'll go back to using the create method here

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
shruti0085: I will defer to DEXP team and @zixlin7 / @mj632 regarding any concerns about behaviors across diff clients.

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanDiagnosticsProvider.ts
viktorsaws: I think keeping no-op code for potential future use is typically unnecessary, unless we know for sure it will be implemented. When there is a use case for hover handling, it can be implemented.

rbbarad: okay - removed

shruti0085 (approved):

mj632 (approved):

PR #212, opened 2024-04-26
[WIP] Chat client unit tests

volodkevych:
## License


chat-client/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "@aws/q-chat-client",
+    "name": "@aws/chat-client",
     "version": "0.0.1",
     "description": "Q Chat Client",
```
kmile, line 4: Should we remove the 'Q' here or add 'AWS'?

chat-client/src/client/chat.test.ts
```diff
@@ -0,0 +1,69 @@
...
+injectJSDOM()
+
+import { afterEach } from 'mocha'
+import { assert } from 'sinon'
+import { NEW_TAB_CREATED, TAB_CHANGED, TAB_REMOVED, UI_IS_READY } from '../contracts/serverContracts'
+import { SEND_TO_PROMPT, TAB_ID_RECEIVED } from '../contracts/uiContracts'
+import { createChat } from './chat'
+import sinon = require('sinon')
+
+describe('Chat', () => {
```
kmile, line 12: Nit: the 'Chat' here gets added to the test descriptions, so you'd get output like:

```
Chat chat initialized publishes ready event
```

Which reads a bit off.

chat-client/src/client/chat.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            postMessage: sandbox.stub(),
+        }
+    })
+
+    afterEach(() => {
+        sandbox.restore()
+    })
+
+    it('chat initialized publishes ready event', () => {
+        createChat(clientApi)
```
kmile, line 27: What does `createChat` do exactly? Create the Chat UI? The Chat JSON/RPC connection? The Chat Proxy/Router/Client? Can we make this more descriptive of which component it is?

chat-client/src/client/chat.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        const tabId = mynahUi.updateStore('', {})
+        mynahUi.updateStore('', {})
+        mynahUi.selectTab(tabId!, (mynahUi as any).lastEventId)
+
+        assert.calledWithMatch(clientApi.postMessage, { command: TAB_CHANGED, params: { tabId: tabId } })
+    })
+
+    function createInboundEvent(params: any) {
+        const event = new CustomEvent('message') as any
+        event.data = params
```
kmile, line 66: Should we use `detail` for the additional data here? (https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/detail)

chat-client/src/client/chat.test.ts
volodkevych: It injects chat UI with `postMessage` listener and the logic around it into the destination. This is also the function that is exposed to destination - I did not want it to be too specific intentionally. Will `createChatClient` look better?

chat-client/src/client/chat.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    it('chat initialized publishes ready event', () => {
+        createChat(clientApi)
+        assert.calledOnceWithExactly(clientApi.postMessage, { command: UI_IS_READY })
+    })
+
+    it('send to prompt publishes telemetry event', () => {
+        createChat(clientApi)
+
+        const sendToPromptEvent = createInboundEvent({ command: SEND_TO_PROMPT, params: { prompt: 'hey' } })
+        window.dispatchEvent(sendToPromptEvent)
```
kmile, line 35: Do integrators of the `chat-client` need to dispatch messages to `window` to communicate with it? Which `window` is that in the context of `postMessage`?

chat-client/src/client/chat.test.ts
volodkevych: It did not work for me.

chat-client/src/client/mynahUi.ts
```diff
@@ -6,30 +6,23 @@ import { MynahUI, NotificationType } from '@aws/mynah-ui'
...
+export const createMynahUi = (messager: Messager, tabFactory: TabFactory): [MynahUI, InboundChatApi] => {
+    const mynahUi = new MynahUI({
         onReady: messager.onUiReady,
         onTabAdd: (tabId: string) => {
-            mynahUI.updateStore(tabId, {})
-            tabStorage.addTab({
-                id: tabId,
-                isSelected: true,
-            })
+            mynahUi.updateStore(tabId, {})
```
kmile, line 18: This doesn't do anything with the tab in the store (yet, until it's filled in with some data)

chat-client/src/client/chat.test.ts
volodkevych: For cross-origin, `postMessage` will work as usual.

chat-client/src/client/mynahUi.ts
volodkevych: I can remove it for now. In the future, it will be used for quick actions at least.

kmile (approved):

PR #213, opened 2024-04-26
Handle chat prompt request and tab notification handlers in QChatServer

makenneth:
## Description
- Add handlers for chat prompt request, tab add notification and tab remove notification
   - There are probably some inaccuracy in the event parser, which I can fix later
- Add initial error handling which can be improved when we work on telemetry

## Testing
- Added unit tests
- Tested with the mock vscode client
 

## License


client/vscode/src/chatActivation.ts
```diff
@@ -14,6 +15,16 @@ export function registerChat(languageClient: LanguageClient, extensionUri: Uri)
 
     panel.webview.onDidReceiveMessage(message => {
         languageClient.info(`vscode client: Received ${JSON.stringify(message)} from chat`)
+
+        // TODO: get server contract types from chat-client
```
volodkevych, line 19: Those specific contract types should be in `@aws/language-server-runtimes-types` already existing package.

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
```diff
@@ -1,39 +1,110 @@
...
+                logging.log('cancellation requested')
                 session.abortRequest()
             })
 
-            return {}
+            let requestInput: GenerateAssistantResponseCommandInput
+            let response: GenerateAssistantResponseCommandOutput
+
+            try {
+                requestInput = convertChatParamsToRequestInput(params)
```
volodkevych, line 67: You could return Result<T> from `convertChatParamsToRequestInput`. Then you would not need to throw error there, catch it here and transform into error. It would look a bit better.

client/vscode/src/chatActivation.ts
makenneth: Great! I have updated to use the runtime types

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
```diff
@@ -1,39 +1,110 @@
+import {
+    GenerateAssistantResponseCommandInput,
+    GenerateAssistantResponseCommandOutput,
+} from '@amzn/codewhisperer-streaming'
+import { ChatParams, ErrorCodes, ResponseError } from '@aws/language-server-runtimes/protocol'
```
viktorsaws, line 5: small note: as a server developer, you should be able to rely on types solely from `@aws/language-server-runtimes/server-interface`. `/protocol` subpackage purpose to define protocol for client communication, which is that proxied to server through `server-interface` shapes transiently.

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
volodkevych: Depending on how your other error handling logic for chat looks like and if you want to reuse that `Result` in other places, you could return `ResponseError` there or more generic error (like string, for example) and transform it to `ResponseError` in `onChatPrompt`.

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
makenneth: Great to know, I will keep that in mind. I have updated the import location.

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
makenneth: Thanks for the suggestion and the example provided offline. I have updated to use the result pattern

server/aws-lsp-codewhisperer/src/language-server/chat/chatEventParser.ts
```diff
@@ -0,0 +1,97 @@
...
+            ...reference,
+            information: ChatEventParser.getReferencedInformation(reference),
+        }
+    }
+
+    constructor(messageId: string) {
+        this.messageId = messageId
+    }
+
+    public processPartialEvent(chatEvent: ChatResponseStream): ChatResult {
```
volodkevych, line 40: You could convert `ChatResult` into `Result<T, TError>` too. That way, you would not need `error` field on the class. Usage of the class would be more intuitive.

volodkevych (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatEventParser.ts
makenneth: Agreed. I will update these to follow the same pattern in a follow up soon

PR #225, opened 2024-05-02
added patches for markdown and unsafe-eval

aderende:
## Problem
- we do not want to show footer and header in hover content
- ajv compile require `unsafe-eval` rule to be enabled. We can't enable this rule.

## Solution
Added patches for markdown hover result and removed ajv usage.

## Testing
ran `npm install && npm run test`
checked in browser that unsafe-eval is not required after applying the patch

## License

### DIFF
--- a/app/aws-lsp-yaml-json-binary/src/tests/testUtilsCF.ts
+++ b/app/aws-lsp-yaml-json-binary/src/tests/testUtilsCF.ts
@@ -420,7 +420,7 @@ Globals:
 export const HOVER_YAML = {
     contents: {
         kind: 'markdown',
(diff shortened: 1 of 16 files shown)


justinmk3:
How will other developers understand what's going on here, and how to maintain it? Is it documented somewhere?

aderende:
> How will other developers understand what's going on here, and how to maintain it? Is it documented somewhere?

Sure, it's a good idea. This patch is suppose to be temporary, because we work on merging the same changes to the package. I'll add the information to read.me and to the yaml-common package

imykhai (approved):

PR #226, opened 2024-05-03
Refactor chat handlers

makenneth:
## Description
- Create a controller logic to centralize the handler functions so it's more maintainable/readable
- Convert many of the code to use the Result pattern so we can have a more unified error handling pattern
 
## Note
- Tests in `chatController.test.ts` are moved from `QChatServer.test.ts` and are modified accordingly

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
@@ -89,7 +89,7 @@ describe('ChatController', () => {
         clock.restore()
     })
 
-    it('chatController creates a session when a tab add notifcation is received', () => {
+    it('creates a session when a tab add notifcation is received', () => {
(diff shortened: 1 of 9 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
```diff
@@ -0,0 +1,253 @@
...
+    })
+
+    afterEach(() => {
+        generateAssistantResponseStub.restore()
+        disposeStub.restore()
+        ChatSessionManagementService.reset()
+        clock.restore()
+    })
+
+    it('chatController creates a session when a tab add notifcation is received', () => {
```
volodkevych, line 92: Won't `chatController` be duplicated in test names, because it's also in `describe`?

makenneth: Good catch thanks! I removed the duplicated `chatController` text.

PR #231, opened 2024-05-06
Pass document context to Q

makenneth:
## Description
Pass editor context to `generateAssistantResponse` using the `CursorState` and `documentIdentifier` that can be passed along in the request now. I referenced the existing vscode's implementation on how they extract text: they expand selection on both sides of the cursor to give more context to Q up until character limit. However, the `TextDocument` structure is different and I was able to find a more simple approach.

I reused and slightly expanded on the language detection logic. Q is actually able to make suggestions even if `languageName` is not provided.

## Notes
- I am not including the symbol extraction logic yet until I fully explore the deployment solution for the `.wasm` files for both types of executable.
- I will move the types for `CursorState` and `ChatParams` import to server-interface. At the moment, the new types are only released in the types package.

## Testing
- Added tests
- Tested each language type manually
- Tested flow in `Run and Debug`

context-screenshot

## License


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
```diff
@@ -249,5 +254,96 @@ describe('ChatController', () => {
...
+                    end: {
+                        line: 1,
+                        character: 1,
+                    },
+                },
+            }
+
+            beforeEach(() => {
+                extractEditorStateStub = sinon.stub(DocumentContextExtractor.prototype, 'extractEditorState')
+                testFeatures.workspace.getTextDocument.resolves(typescriptDocument)
```
kmile, line 277: `TestFeatures` has an implementation of `getTextDocument` already you can leverage by using `openDocument` (https://github.com/aws/language-server-runtimes/blob/fe773e98f92d49890d6a72fd145b9192196ba51c/runtimes/testing/TestFeatures.ts#L96).

Saves you keeping track of another stub, and allows you to have multiple files open.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -73,7 +60,9 @@ export class ChatController implements ChatHandlers {
         let response: GenerateAssistantResponseCommandOutput
 
         try {
+            this.#log('Request from tab:', params.tabId, JSON.stringify(requestInput.data))
             response = await session.generateAssistantResponse(requestInput.data)
+            this.#log('Response to tab:', params.tabId, JSON.stringify(response.$metadata))
```
kmile, line 65: Do we want to keep this verbose logging or are these temporary log statements?

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -83,16 +72,64 @@ export class ChatController implements ChatHandlers {
             )
         }
 
-        const result = await this.#processAssistantResponse(response)
+        const result = await this.#processAssistantResponse(response, params.partialResultToken)
```
kmile, line 75: Is this deliberately outside the `try/catch`? Is there no way that parsing chat events from the server response can fail? Or if they can, would we need a specific exception?

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.test.ts
```diff
@@ -0,0 +1,163 @@
...
+                        end: {
+                            line: 0,
+                            character: 11,
+                        },
+                    },
+                },
+            }
+
+            const result = await documentContextExtractor.extractEditorState(mockTSDocument, {
+                // highlighing "log"
```
kmile, line 37: nit: typo

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.test.ts
```diff
@@ -0,0 +1,163 @@
...
+                        line: 1,
+                        character: 15,
+                    },
+                },
+            })
+
+            assert.deepStrictEqual(result, expected)
+        })
+
+        it('extracts editor state for collapsed poisition', async () => {
```
kmile, line 53: nit: typo

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+describe('DocumentContext', () => {
+    const mockTypescriptCodeBlock = `function test() {
+    console.log('test')
+}`
+    const mockTSDocument = TextDocument.create('file://test.ts', 'typescript', 1, mockTypescriptCodeBlock)
+
+    describe('documentContextExtractor.extractEditorState', () => {
+        it('extracts editor state for range selection', async () => {
+            const documentContextExtractor = new DocumentContextExtractor(19)
```
kmile, line 14: Does the limit always respect the nearest line boundaries or does it just go `N/2` characters back and `N/2` characters forward?

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.test.ts
kmile: I don't understand why this test returns line 2, when the cursor state is on line 1?

kmile: Nevermind, the result is transposed by taking the limit characters, so the lines that are cut off are also cut off from the returned `cursorState` to make it match.

server/aws-lsp-codewhisperer/src/language-server/languageDetection.ts
```diff
@@ -1,58 +1,96 @@
...
     | 'rust'
     | 'scala'
     | 'shell'
-    | 'shell'
     | 'sql'
-    | 'plaintext'
+    | 'tsx'
+    | 'typescript'
+
+    // These work as well
```
kmile, line 23: Should we just include these in the list above?

server/aws-lsp-codewhisperer/src/language-server/languageDetection.ts
```diff
@@ -1,58 +1,96 @@
...
+    '.sql': 'sql',
+    '.tf': 'tf',
     '.ts': 'typescript',
+    '.tsx': 'tsx',
+    '.vue': 'vue',
+    '.yaml': 'yaml',
+    '.yml': 'yaml',
+}
+
+export const additionalLanguageMapping: { [key: string]: CodewhispererLanguage } = {
```
kmile, line 65: Where do we get these language identifiers from? If it's not `getCodeWhispererLanguageIdByTextDocumentLanguageId`, `getCodeWhispererLanguageIdByExtension`, what is the source of these identifiers on the left side of this map?

kmile (approved):
Great to see this thorough test coverage! 🦄

server/aws-lsp-codewhisperer/src/language-server/languageDetection.ts
makenneth: Sure I will just include this in the list

server/aws-lsp-codewhisperer/src/language-server/languageDetection.ts
makenneth: This mapping is `{ [textDocumentLanguageId: string]: CWPRLanguageId }`. I noticed `getLanguageId` is not checking against the entire list so I decided to create a single map that has all the mapping.

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.test.ts
makenneth: No, this doesn't respect line boundaries, it will go all the way until the characters limit or document boundaries, whichever comes first. If we see an issue with sending partial lines, I can modify the logic.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
makenneth: that's great to know, I updated to use that

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
makenneth: vscode has something similar in their debug log. But for now, I trimmed down on the input logging to include only conversation id. We can revisit on this at a later point.

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.test.ts
kmile: Nope, sounds good to me!

There could be some theoretical edge cases where cutting off partial statements could result in the intent changing

`mock.deleteAllFiles()` vs `deleteAllFiles()`

but lines don't resolve that, only a context-aware parser will be able to catch that. Since the user is already selecting the relevant file contents, and we _increase_ the context that we send, this seems like a highly unlikely edge case.

PR #232, opened 2024-05-07
Upgrade typedoc version

rtarcr:
## Problem
Dependabot upgraded the typescript version and the build passed for some reason. It should trigger a version conflict instead due to `typedoc` dependency requiring typescript version 4.x 

## Solution
Upgrade typedoc to latest version

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src.gen/@amzn/codewhisperer-streaming/package.json
+++ b/server/aws-lsp-codewhisperer/src.gen/@amzn/codewhisperer-streaming/package.json
@@ -62,7 +62,7 @@
         "@types/uuid": "^9.0.8",
         "concurrently": "7.0.0",
         "rimraf": "^3.0.0",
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src.gen/@amzn/codewhisperer-streaming/package.json
```diff
@@ -63,7 +63,7 @@
         "concurrently": "7.0.0",
         "rimraf": "^3.0.0",
         "typedoc": "0.23.23",
-        "typescript": "~5.4.5"
+        "typescript": "~4.9.5"
```
dogusata, line 66: Typedoc is usually not a mandatory dependency, is it not possible to upgrade the typedoc version to use the one dependabot suggests?

dogusata (approved):

PR #249, opened 2024-05-16
Send telemetry for vote, copyCodeToClipboard and authFollowUpClicked event

rahmaniaam:
## Problem
Telemetry-only events are not yet handled

## Solution
Created missing handlers for copyCodeToClipboard and vote in from mynah-ui, and used telemetry API to send to extension.
Additional forwarding of telemetry in vscode client for testing purposes

Dependant on https://github.com/aws/language-server-runtimes/pull/115

## License

### DIFF
--- a/chat-client/src/client/messager.ts
+++ b/chat-client/src/client/messager.ts
@@ -9,6 +9,8 @@ import {
     InsertToCursorPositionParams,
     SendToPromptParams,
     TabIdReceivedParams,
+    CopyCodeToClipboardParams,
+    VoteParams,
 } from '@aws/chat-client-ui-types'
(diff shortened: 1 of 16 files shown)


chat-client/src/client/messager.ts
```diff
@@ -9,6 +9,8 @@ import {
     InsertToCursorPositionParams,
     SendToPromptParams,
     TabIdReceivedParams,
+    CopyCodeToClipboardParams,
```
volodkevych, line 12: Cannot those be chat-client internal types? Types in '@aws/chat-client-ui-types' should be the ones that clients (e.g. IDE Extension depends on).

rahmaniaam: ah I think I confused the types for _all_ communication between the chat-client and the extension. Yes in that case I can make these internal instead

PR #252, opened 2024-05-17
Extract symbols to improve Q's suggestions

makenneth:
## Description
Apologies for the big PR.  I am missing some test cases, but I will come back for them when I work on quick actions.

This PR extracts symbols from document using `@aws/fully-qualified-names` to improve Q's suggestions. The execution will be run by workers to increase resilience. Since the implementation of workers and the fqn library are platform dependent, we need to split up the entry points so build system can bundle the appropriate files.

Each environment will have its own worker file that takes care of importing fqn and running the code. The exported `FqnWorkerPool` will use this worker. Build systems are then able to locate the right bundle based on the "target" option

- [x] Browser (didn't get `TextDecoder` error, which meant it was the bundler version)
- [x] Toolkits (vscode, vs)

## Note
- I did reference the vscode logic so dedup, sort and limit are all based on that implementation 

## Next up
- Updating bundle to include chat server and the required artifacts (https://github.com/makenneth/language-servers/commit/7aaeed213f9438d34895bbebda3f138c367e13f2)

## License


core/aws-lsp-fqn/package.json
```diff
@@ -0,0 +1,41 @@
+{
+    "name": "@aws/lsp-fqn",
+    "version": "0.0.1",
+    "description": "CodeWhisperer Language Server",
```
volodkevych, line 4: nit: Description needs to be updated

core/aws-lsp-fqn/src/common/defaults.ts
```diff
@@ -0,0 +1,2 @@
+// timeout for a graceful termination of the process
+export const DEFAULT_TIMEOUT = 5000
```
volodkevych, line 2: not used?

core/aws-lsp-fqn/src/common/fqnExtractor.ts
```diff
@@ -0,0 +1,34 @@
...
+    fileText: string,
+    extent: Extent
+): Promise<FqnExtractorOutput> {
+    switch (languageId) {
+        case 'java':
+            return Java.findNamesWithInExtent(fileText, extent)
+        case 'javascriptreact':
+        case 'typescriptreact':
+            return Tsx.findNamesWithInExtent(fileText, extent)
+        case 'javascript':
```
volodkevych, line 15: I see in all other examples 'javascript' is used with `Tsx`.

core/aws-lsp-fqn/src/node/fqnExtractorWorker.ts
```diff
@@ -0,0 +1,21 @@
...
+async function start() {
+    try {
+        const symbolExtractorInput: FqnExtractorInput = workerData.input
+        console.log(`Extracting fully qualified names for ${symbolExtractorInput.languageId}`)
+
+        const data = await extract(symbolExtractorInput)
+
+        parentPort?.postMessage({ data, success: true })
+    } catch (e) {
+        console.error(`Failed to extract fully qualified names: ${e instanceof Error ? e.message : 'unknown'}`)
```
volodkevych, line 14: Do you know, will there be any metrics on CW/Toolkit side for us to know how often it works/fails?

server/aws-lsp-codewhisperer/package.json
```diff
@@ -29,7 +29,7 @@
     "dependencies": {
         "@amzn/codewhisperer-streaming": "file:./src.gen/@amzn/codewhisperer-streaming",
         "@aws/language-server-runtimes": "^0.2.5",
-        "@aws/language-server-runtimes-types": "^0.0.2",
```
volodkevych, line 32: Don't we need `types` package anymore or is it some kind of merge mistake?

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentFqnExtractor.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export interface DocumentFqnExtractorConfig {
+    nameMinLength?: number
+    nameMaxLength?: number
+    maxSymbols?: number
+    timeout?: number
+    logger?: Features['logging']
+}
+
+export type FqnSupportedLanagues =
```
volodkevych, line 14: typo

volodkevych: Cann't those be defined in 1 place, e.g. `fqn` package?

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentFqnExtractor.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    // made public for stubbing in test
+    async findNamesInRange(
+        fileText: string,
+        selection: Range,
+        languageId: FqnSupportedLanagues
+    ): Promise<Result<FqnExtractorOutput, string>> {
+        const [resultPromise, dispose] = fqnExtractor(
+            {
+                fileText: fileText.replace(/([\uE000-\uF8FF]|\uD83C[\uDF00-\uDFFF]|\uD83D[\uDC00-\uDDFF])/g, ''),
```
volodkevych, line 132: I think this requires some explanation. What is being replaced, why?

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentFqnExtractor.ts
```diff
@@ -0,0 +1,161 @@
...
+        this.#nameMinLength = nameMinLength ?? DocumentFqnExtractor.DEFAULT_CONFIG.nameMinLength
+        this.#nameMaxLength = nameMaxLength ?? DocumentFqnExtractor.DEFAULT_CONFIG.nameMaxLength
+        this.#maxSymbols = maxSymbols ?? DocumentFqnExtractor.DEFAULT_CONFIG.maxSymbols
+        this.#timeout = timeout
+        this.#logger = logger
+        this.#disposeFunctionSet = new Set()
+    }
+
+    public dispose() {
+        this.#disposeFunctionSet.forEach(disposeFunction => disposeFunction())
```
volodkevych, line 57: When will those `disposeFunction`'s be called? I see a long chain of objects creation and it's difficult to understand when stuff is going to be disposed. I just try to verify we do not create memory issues with this.

makenneth: In the way it is structured, there could be multiple threads spawned from this class at the same time. - I think I will change it so we instantiate a new instance every time so we have a more straightforward logic.

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentFqnExtractor.ts
makenneth: Yes, it should be. I will move them, thanks!

core/aws-lsp-fqn/package.json
makenneth: Oops! Missed that, I will update!

core/aws-lsp-fqn/src/common/fqnExtractor.ts
```diff
@@ -0,0 +1,34 @@
...
+        case 'typescript':
+            return TypeScript.findNamesWithInExtent(fileText, extent)
```
justinmk3, line 17: There is a known issue with the fqn package, where it hangs on various kinds of js/ts input. For that reason, we have currently disabled it for those filetypes in the Amazon Q extension: https://github.com/aws/aws-toolkit-vscode/pull/4954

volodkevych: That's why we are running it in a separate worker thread with timeout.

core/aws-lsp-fqn/src/node/runner.ts
```diff
@@ -0,0 +1,71 @@
...
+import { FqnExtractor, FqnExtractorInput, FqnExtractorOutput, Result, RunnerConfig } from '../common/types'
+import path = require('path')
+
+export const fqnExtractor: FqnExtractor = (input: FqnExtractorInput, config?: RunnerConfig) => {
+    const { logger, timeout = 5000 } = config ?? {}
+
+    let hasFulfilled = false
+
+    let timer: NodeJS.Timeout
+    const worker = new Worker(path.resolve(__dirname, './fqnExtractorWorker.js'), {
```
kmile, line 11: Do we have to create the worker thread on every invocation? Does performance of the WASM module degrade over time or instantly on malformed inputs? Does it hang or stall, or saturate the CPU?

I worry about the performance overhead given this is on the chat request path. The code itself doesn't seem to have much initialization overhead (it's just `extract`) but there is overhead is spawning the thread.

And `extract` is an `async` method so that implies to me there is some I/O wait or otherwise non-CPU bound processing happening? Could one worker thread handle multiple concurrent extractions?

core/aws-lsp-fqn/src/node/runner.ts
```diff
@@ -0,0 +1,71 @@
...
+
+            reject({ success: false, error: error.message })
+        })
+
+        timer = setTimeout(() => {
+            if (!hasFulfilled) {
+                logger?.log('Fully qualified names extraction timed out')
+
+                // this is able to terminate worker that has event loop blocked
+                worker.terminate().then(() => {
```
kmile, line 55: The docs on `terminate` say "Stop all JavaScript execution in the worker thread **as soon as possible**" (emphasis mine). Is this safe to call before `reject` or can the stalled thread still block or delay this call?

core/aws-lsp-fqn/src/node/runner.ts
kmile: Second question: you have a `browser/runner` as well that's currently inactive. Do you plan to move some of the `Promise` logic to some reusable place when that gets implemented? Or do you expect the async and timeout behavior will be different across platforms?

volodkevych: Indeed, using something like `workerpool` could improve here.

core/aws-lsp-fqn/src/common/fqnExtractor.ts
justinmk3: is that mentioned anywhere in the code, docs, etc?

core/aws-lsp-fqn/src/common/defaults.ts
makenneth: updated to use the default here

core/aws-lsp-fqn/src/node/runner.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export const fqnExtractor: FqnExtractor = (input: FqnExtractorInput, config?: RunnerConfig) => {
+    const { logger, timeout = 5000 } = config ?? {}
+
+    let hasFulfilled = false
+
+    let timer: NodeJS.Timeout
+    const worker = new Worker(path.resolve(__dirname, './fqnExtractorWorker.js'), {
+        workerData: { input },
+        // allow us to intercept the logs, otherwise, the output will cause vscode extensions to crash
```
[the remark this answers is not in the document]
makenneth: I think it should still be good to log these stdin and stderr data in the logger instance. I removed the mention of vscode though.

core/aws-lsp-fqn/src/node/runner.ts
makenneth: Using `workerpool` now, which simplified some of the things!

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentFqnExtractor.ts
makenneth: Workerpool simplified this, so no longer relevant

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentFqnExtractor.ts
makenneth: Add explanation

core/aws-lsp-fqn/src/node/runner.ts
makenneth: This code has been removed -- `workerpool` comes with timeout feature

core/aws-lsp-fqn/src/node/runner.ts
makenneth: I think so far I have only experience the module crashing instantly. I haven't experience the stalling issue so that needs to be investigated. I just implemented this with the worst case assumption that the I/O will be blocked.

Using `workerpool` as Oksana suggested will reduce the overhead, by keeping workers initialized. It also manages the task queue for us, which I wasn't sure how to do.

For the second question, the two main things that differ between environment are 1) how the worker script is bundled and 2) how the fqn library is imported. The newest refactor has reduced the number custom config requried.

core/aws-lsp-fqn/src/common/fqnExtractor.ts
makenneth: I am prepping a doc to explain the fqn usage in DEXP server, which just needs to be updated to reflect the newest changes, but I can also add to the readme here.

core/aws-lsp-fqn/src/common/CommonFqnWorkerPool.ts
volodkevych, line 1: File names introduced in this PR are inconsistent.

core/aws-lsp-fqn/src/browser/aws-lsp-fqn-worker.ts
volodkevych, line 1: This file seems to be unused.

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentFqnExtractor.ts
```diff
@@ -0,0 +1,132 @@
...
+        'typescript',
+        'javascriptreact',
+        'typescriptreact',
+    ])
+
+    #nameMinLength: number
+    #nameMaxLength: number
+    #maxSymbols: number
+
+    #workerPool: FqnWorkerPool
```
volodkevych, line 34: nit: This can be defined as `IFqnWorkerPool` here.

core/aws-lsp-fqn/src/browser/FqnWorkerPool.ts
```diff
@@ -0,0 +1,29 @@
+import { CommonFqnWorkerPool } from '../common/CommonFqnWorkerPool'
+import { ExtractorResult, FqnExtractorInput, WorkerPoolConfig } from '../common/types'
+
+// TODO: implement logic for browser/webworker environment
+export class FqnWorkerPool extends CommonFqnWorkerPool {
```
volodkevych, line 5: nit: it would look a bit cleaner here if `FqnWorkerPool` just implemented `IFqnWorkerPool`. This stub does not need common logic for any behavior.

core/aws-lsp-fqn/src/common/CommonFqnWorkerPool.ts
```diff
@@ -0,0 +1,44 @@
...
+export class CommonFqnWorkerPool implements IFqnWorkerPool {
+    #workerPool: Pool
+    #logger?: Logger
+    #timeout: number
+
+    constructor(filePath: string, { timeout = DEFAULT_TIMEOUT, logger, workerPoolOptions }: WorkerPoolConfig = {}) {
+        this.#timeout = timeout
+        this.#logger = logger
+        this.#workerPool = pool(filePath, {
+            maxWorkers: DEFAULT_MAX_WORKERS,
```
volodkevych, line 15: Should we also limit `maxQueueSize`?

makenneth: Yes we should, thanks! I set it to 10 for now

volodkevych (approved):

PR #257, opened 2024-05-21
feat: add PartiQL language aserver

akoreman:
## Problem

Adds an LSP server for the [PartiQL language](https://partiql.org/). This server offers diagnostics for PartiQL files by parsing the file using the [Rust parser maintained by the PartiQL team](https://github.com/partiql/partiql-lang-rust).

This parser is compiled to a WebAssembly binary using the bindings maintained by the PartiQL team in the [PartiQL playground package](https://github.com/partiql/partiql-rust-playground). This binary is further encoded as a base64 string in a Javascript file to make it easy to be bundled together with the other assets of the server. See the README for instructions on how to update the binary.

## Solution

Screenshot 2024-05-20 at 23 54 32

- Tested the server wrapped in the webworker runtime a web-based LSP client.
- Added unit tests covering both the server and the service.

## License

### DIFF
--- a/app/aws-lsp-partiql-binary/src/index.ts
+++ b/app/aws-lsp-partiql-binary/src/index.ts
@@ -1,6 +1,6 @@
-import { PartiQLLanguageServer } from '@aws/aws-lsp-partiql'
 import { RuntimeProps } from '@aws/language-server-runtimes/runtimes/runtime'
 import { standalone } from '@aws/language-server-runtimes/runtimes/standalone'
+import { PartiQLLanguageServer } from '@aws/lsp-partiql'
 
 const props: RuntimeProps = {
     servers: [PartiQLLanguageServer],
--- a/app/aws-lsp-partiql-binary/src/index.ts
+++ b/app/aws-lsp-partiql-binary/src/index.ts
@@ -1,0 +1,9 @@
+import { PartiQLLanguageServer } from '@aws/aws-lsp-partiql'
+import { RuntimeProps } from '@aws/language-server-runtimes/runtimes/runtime'
+import { standalone } from '@aws/language-server-runtimes/runtimes/standalone'
+
+const props: RuntimeProps = {
+    servers: [PartiQLLanguageServer],
+    name: 'AWS PartiQL',
+}
+standalone(props)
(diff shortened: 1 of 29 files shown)


justinmk3:
> This binary is further encoded as a base64 string in a Javascript file to make it easy to be bundled together with the other assets of the server.

Why can't it be bundled as a wasm file?

app/aws-lsp-partiql-binary/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import { PartiQLLanguageServer } from '@aws/aws-lsp-partiql'
```
viktorsaws, line 1: nit: `aws-` prefix in package name is redundant

server/aws-lsp-partiql/src/server/language-server.ts
```diff
@@ -0,0 +1,72 @@
+import { type DidChangeTextDocumentParams, type Server } from '@aws/language-server-runtimes/server-interface'
+import { TextDocumentSyncKind, type DidOpenTextDocumentParams } from 'vscode-languageserver'
+import { TextDocument } from 'vscode-languageserver-textdocument'
```
viktorsaws, line 3: nit: `TextDocument` interface is available through Runtimes `@aws/language-server-runtimes/server-interface`, so direct dependency on `vscode-languageserver-textdocument` in code.

`TextDocumentSyncKind` and `DidOpenTextDocumentParams` should be also available, so direct dependency on VSCode LSP packages can be minimized even more.

server/aws-lsp-partiql/src/server/language-service.ts
```diff
@@ -0,0 +1,39 @@
+import type { Diagnostic } from 'vscode-languageserver-protocol'
```
viktorsaws, line 1: nit: imports from `vscode-languageserver-protocol` should be available from  `@aws/language-server-runtimes/server-interface`

viktorsaws (approved):

app/aws-lsp-partiql-binary/src/index.ts
akoreman: yeah that `aws` there is a bit redundant, removed it

akoreman:
@justinmk3 Main reason for not bundling it as a wasm file is to make it easier to use the server as a web worker when assets are loaded from a CDN. If the binary is provided as a wasm file, the worker needs to know which CDN URL to fetch the binary from. This URL needs to be communicated by the client to the web worker which adds complexity to the integration into a web app.

viktorsaws (approved):

PR #264, opened 2024-05-27
Added configuration to produce Amazon Q server with IAM Credentials provider service

viktorsaws:
## Problem

We don't have example of building and running Amazon Q server with IAM credentials.

## Solution

Added configuration to build standalone Amazon Q server with inline completions support. Updated VSCode launch configuration to run it with sample VSCode client extension.

## License


server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -91,6 +96,14 @@ export class CodeWhispererServiceIAM extends CodeWhispererServiceBase {
         }
     }
 
+    updateAwsConfiguration = (awsConfig: AWSConfig) => {
```
volodkevych, line 99: nit: can be moved to base?

server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -50,8 +54,9 @@ export class CodeWhispererServiceIAM extends CodeWhispererServiceBase {
     private readonly codeWhispererRegion = 'us-east-1'
     private readonly codeWhispererEndpoint = 'https://codewhisperer.us-east-1.amazonaws.com/'
 
-    constructor(credentialsProvider: CredentialsProvider) {
+    constructor(credentialsProvider: CredentialsProvider, additionalAwsConfig: AWSConfig) {
```
volodkevych, line 57: Can `additionalAwsConfig` be optional so that `{}` would not be required?

.vscode/settings.json
```diff
@@ -7,6 +7,6 @@
             "source.organizeImports": true,
             "source.sortMembers": true
         },
-        "editor.formatOnSave": true
+        "editor.formatOnSave": false
```
volodkevych, line 10: Why are you changing this configuration. Should be useful to have it enabled.

viktorsaws: I find it annoying, given that my IDE config does not pick up prettier settings automatically. I will revert this config, it slipped in with other changes

PR #258, opened 2024-05-23
Update `patchYamlJsonPackage.js` to handle paths with spaces in `applyPatch`

twwildey:
## Problem

When `language-servers` is checked out under a directory that contains spaces in the full path, `npm install` will fail to complete due to `patchYamlJsonPackage.js`.

## Solution

This commit fixes this issue by ensuring file paths are quoted to support spaces in them.

### Testing

Ensured `npm install` succeeds in a directory with spaces in the full path (i.e. under a user's `Application Support` folder for Sublime Text on OS X).

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #268, opened 2024-05-29
Update `patchYamlJsonPackage.js` to handle paths with spaces in `applyPatch`

twwildey:
## Problem

When `language-servers` is checked out under a directory that contains spaces in the full path, `npm install` will fail to complete due to `patchYamlJsonPackage.js`.

## Solution

This commit fixes this issue by ensuring file paths are quoted to support spaces in them.

### Testing

Ensured `npm install` succeeds in a directory with spaces in the full path (i.e. under a user's `Application Support` folder for Sublime Text on OS X).

## License

### DIFF
--- a/app/aws-lsp-codewhisperer-binary/src/token-standalone.ts
+++ b/app/aws-lsp-codewhisperer-binary/src/token-standalone.ts
@@ -1,10 +1,11 @@
 import { standalone } from '@aws/language-server-runtimes/runtimes'
 import { RuntimeProps } from '@aws/language-server-runtimes/runtimes/runtime'
+import { NetTransformServer } from '@aws/lsp-codewhisperer/out/language-server/netTransformServer'
 import {
     CodeWhispererSecurityScanServerTokenProxy,
     CodeWhispererServerTokenProxy,
+    QChatServerProxy,
 } from '@aws/lsp-codewhisperer/out/language-server/proxy-server'
-import { NetTransformServer } from '@aws/lsp-codewhisperer/out/language-server/netTransformServer'
 
 const MAJOR = 0
 const MINOR = 1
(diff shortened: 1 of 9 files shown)


twwildey:
This is a continuation of https://github.com/aws/language-servers/pull/258.

makenneth (approved):

server/aws-lsp-yaml-json/patchYamlJsonPackage.js
```diff
@@ -35,7 +35,7 @@ const filePathToPatchPathUnsafeEval = {
 function applyPatch(filePathToPatchPath) {
     for (var filePath in filePathToPatchPath) {
         const pathToPatch = `${__dirname}/${filePathToPatchPath[filePath]}`
-        const script = `cd ${rootPackage} && patch ${filePath} ${pathToPatch}`
+        const script = `cd "${rootPackage}" && patch "${filePath}" "${pathToPatch}"`
         console.log(script)
         const output = execSync(script, { encoding: 'utf-8', timeout: 2000 })
```
justinmk3, line 40: It is almost always more robust to use the list form instead of passing a string to the exec() family of functions. That would avoid the quoting you have tried above, and avoids problems with wildcards and other shell-specific special chars that might appear in the `script` string.

justinmk3, line 40: here is an untested (but probably correct) example. 

note the use of `cwd` to set the working directory.

```suggestion
        console.log(script);
        execFile('patch', [filePath, pathToPatch], { cwd: rootPackage, encoding: 'utf-8', timeout: 2000 }, (error, stdout, stderr)
```

twwildey: So what about the logging statement?  Would we want the following instead?

```js
function applyPatch(filePathToPatchPath) {
    for (var filePath in filePathToPatchPath) {
        const pathToPatch = `${__dirname}/${filePathToPatchPath[filePath]}`
        const patchProc = execFile('patch', [filePath, pathToPatch], { cwd: rootPackage, encoding: 'utf-8', timeout: 2000 })

        console.log({
            cmd: patchProc.spawnfile,
            args: patchProc.spawnargs,
            cwd: rootPackage,
        });
    }
}
```

twwildey:
I've once again created the PR: https://github.com/aws/language-servers/pull/270

server/aws-lsp-yaml-json/patchYamlJsonPackage.js
twwildey: I've once again created the PR with the code from above here: https://github.com/aws/language-servers/pull/270

I've confirmed that `applyPatch` works as well.

justinmk3:
> I've once again created the PR: #270

Why? That loses the PR comments. Please just force-push to your existing branch instead of creating new PRs.

PR #272, opened 2024-05-30
Split Yaml Json LSP into 2 servers

aderende:
## Problem

Split of Yaml/Json LSP Server to 2 separate servers helps to avoid adding extra webpack configuration in case when integrator need only json language support

## Solution

yaml/json server has been split into 2 different servers

## License


app/aws-lsp-json-binary/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "name": "@aws/lsp-json-binary",
+    "version": "0.0.1",
+    "description": "JSON Language Server Binary",
+    "main": "out/index.js",
+    "bin": {
+        "aws-lsp-json-binary": "./out/index.js"
+    },
+    "scripts": {
+        "compile": "tsc --build",
+        "package-x64": "pkg --targets node18-linux-x64,node18-win-x64,node18-macos-x64 --out-path bin --compress GZip .",
```
imykhai, line 11: not exactly related to this PR, but we should stop relying on `pkg`. Instead we should bundle it with webpack/rollup/etc. and run it with `node`

app/aws-lsp-json-binary/src/tests/yamlJsonServerCFInteg.test.ts
imykhai, line 1: Should we rename the file ? It includes `yaml` in the filename.

server/aws-lsp-cloudformation/package.json
```diff
@@ -8,7 +8,7 @@
     },
     "dependencies": {
         "@aws/lsp-core": "^0.0.1",
-        "@aws/aws-lsp-yaml-json": "^0.0.1",
+        "@aws/aws-lsp-json": "^0.0.1",
```
imykhai, line 11: This line can cause issues when we bump version, in npm “^0.0.1” can't be resolved to “0.0.2” or “0.1.1". This is because npm considers a major version “the leftmost non zero digit”. So, “0.0.2" is a major version change 
https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004

imykhai (approved):

server/aws-lsp-cloudformation/package.json
aderende: agreed to solve in next PR

PR #273, opened 2024-05-30
Implement various chat events

rahmaniaam:
## Problem
We want to implement some of the chat events on the chat-client

## Solution
- Add needed event commands and params
- Implement event communication flow from mynah-ui to client
- Implement event logic
- Add handler in vscode client for testing

Note: unit tests in follow up PR

## License

### DIFF
--- a/chat-client/src/client/messager.ts
+++ b/chat-client/src/client/messager.ts
@@ -10,7 +10,17 @@ import {
     SendToPromptParams,
     TabIdReceivedParams,
 } from '@aws/chat-client-ui-types'
-import { ChatParams, TabAddParams, TabChangeParams, TabRemoveParams } from '@aws/language-server-runtimes-types'
+import {
+    ChatParams,
+    FeedbackParams,
+    FollowUpClickParams,
+    InfoLinkClickParams,
+    LinkClickParams,
+    SourceLinkClickParams,
+    TabAddParams,
+    TabChangeParams,
+    TabRemoveParams,
+} from '@aws/language-server-runtimes-types'
 import { TelemetryParams } from '../contracts/serverContracts'
 import { CopyCodeToClipboardParams, VoteParams } from '../contracts/telemetry'
 
@@ -23,6 +33,11 @@ export interface OutboundChatApi {
     telemetry(params: TelemetryParams): void
     insertToCursorPosition(params: InsertToCursorPositionParams): void
     authFollowUpClicked(params: AuthFollowUpClickedParams): void
+    followUpClicked(params: FollowUpClickParams): void
+    sendFeedback(params: FeedbackParams): void
+    linkClick(params: LinkClickParams): void
+    sourceLinkClick(params: SourceLinkClickParams): void
+    infoLinkClick(params: InfoLinkClickParams): void
     uiReady(): void
 }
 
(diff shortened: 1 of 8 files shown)


chat-client/src/client/messager.ts
```diff
@@ -73,6 +92,22 @@ export class Messager {
         this.chatApi.telemetry(params)
     }
 
+    onSendFeedback = (params: FeedbackParams): void => {
+        this.chatApi.sendFeedback(params)
+    }
+
+    onLinkClick = (params: LinkClickParams) => {
```
ege0zcan, line 99: nitpick: these 3 link click functions don't explicitly specify return type as void

client/vscode/src/chatActivation.ts
```diff
@@ -48,6 +54,22 @@ export function registerChat(languageClient: LanguageClient, extensionUri: Uri)
                         })
                     })
                 break
+            case feedbackNotificationType.method:
+                languageClient.sendNotification(feedbackNotificationType, message.params)
+                break
+            case followUpClickNotificationType.method:
+                if (!isValidAuthFollowUpType(message.params.followUp.type))
+                    languageClient.sendNotification(followUpClickNotificationType, message.params)
```
ege0zcan, line 62: the `else` clause in a real world use case is that the extension handles the auth right?

chat-client/src/client/mynahUi.ts
```diff
@@ -131,6 +143,60 @@ export const createMynahUi = (messager: Messager, tabFactory: TabFactory): [Myna
...
+
+                mynahUi.notify({
+                    type: NotificationType.INFO,
+                    title: 'Your feedback is sent',
+                    content: 'Thanks for your feedback.',
+                })
+            }
+        },
+        onLinkClick: (tabId, messageId, link, mouseEvent, eventId) => {
+            mouseEvent?.preventDefault()
```
ege0zcan, line 163: What is the behaviour for link clicks when we prevent default and stop propagation?

client/vscode/src/chatActivation.ts
rahmaniaam: yup, correct

chat-client/src/client/mynahUi.ts
```diff
@@ -131,6 +143,60 @@ export const createMynahUi = (messager: Messager, tabFactory: TabFactory): [Myna
             }
             messager.onVote(payload)
         },
+        onSendFeedback: (tabId, feedbackPayload, eventId) => {
+            if (feedbackPayload.comment !== undefined) {
```
ege0zcan, line 147: I think we don't need this if check and comment can be empty 

On mynahui feedback form, you choose an issue type from the dropdown and optionally type some comments https://github.com/aws/mynah-ui/blob/main/docs/PROPERTIES.md#onsendfeedback

rahmaniaam: ah, I missed that detail, will update the logic

chat-client/src/client/mynahUi.ts
rahmaniaam: I could see the chat-client having additional support for this, but I think it goes back to how destinations wants to handle them.

I'm not entirely sure if this is even needed, but it seems to be useful at least for the VSCode implementation

ege0zcan (approved):

PR #279, opened 2024-06-03
Escape backticks in user prompts

ege0zcan:
## Problem

Right now the way we create markdown code blocks for `send to prompt` is a bit too naive and is prone to breaking when a user selection contains backticks. See for example:

image

## Solution

To overcome this, we can implement a function that finds the longest backtick sequence inside a provided user selection and then creates the markdown code block with more backticks than that amount

This has the small caveat that codeblocks with backticks will now consume more of the character limit 

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -30,6 +30,16 @@ export interface InboundChatApi {
     showError(params: ErrorParams): void
 }
 
+function createMarkdownCodeBlock(input: string) {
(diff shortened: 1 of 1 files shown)


ege0zcan:
Had an offline discussion with @dogusata about this. Although this solution works, it introduces some additional computation that needs to be done by the regex and then the sorting.  We decided to update the logic in mynah-ui to let mynah handle codeblocks by using the `~~~`  syntax for codeblocks

https://github.com/aws/mynah-ui/pull/51

PR #283, opened 2024-06-04
Add script to enable bundleDependencies in monorepo workspaces

rtarcr:
## Problem
We want to publish JSON and YAML LSP servers to NPM. They depend on the core subpackage in the monorepo and to avoid publishing it we need to include the dependency in the published package. 

This is currently not possible due to https://github.com/npm/cli/issues/3466

## Solution
Create script that: 
- reads `bundleDependencies`
- Goes to root node_modules and finds the dependency
- Creates a symlink in subpackage node_modules pointing to the dependency folder
- Run this script as part `prepack` script 

## License

### DIFF
--- a/script/link_bundled_dep.sh
+++ b/script/link_bundled_dep.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# This script is needed as a workaround for https://github.com/npm/cli/issues/3466
+
+root_dir=$(dirname "$0")/..
+target_package_dir=$(pwd)
+
+if [ -z target_package_dir/package.json ]; then
+    echo "package.json not found in the $target_package_dir directory."
+    exit 1
+fi
+
+# Get the bundleDependencies array from package.json
+bundle_dependencies=$(jq -r '.bundleDependencies[]' package.json)
+
+if [ -z "$bundle_dependencies" ]; then
+    echo "No bundleDependencies found in package.json."
+    exit 0
+fi
+
+
+# Loop through each bundleDependency
+for dependency in $bundle_dependencies; do
+    dependency_path="$root_dir/node_modules/$dependency"
+    
+     # Check if the dependency exists in the root node_modules directory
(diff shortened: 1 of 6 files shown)


rahmaniaam:
Looks good! Does the `@aws/lsp-core` package live on inside `root/node_modules/@aws/aws-lsp-yaml/node_modules` after installing or will there be a `root/node_modules/@aws/lsp-core` in wherever the package is installed? I imagine it's the former so it's exclusively used only by `aws-lsp-yaml`

script/link_bundled_dep.sh
```diff
@@ -0,0 +1,42 @@
...
+        link_target=$(readlink -f "$dependency_path")
+
+        mkdir -p $link_location
+        # Clear contents or remove last folder in the path
+        rm -r $link_location
+
+        ln -s $link_target $link_location
+        echo "Symlink created for $dependency"
+    else
+        echo "ERROR: $dependency not found in the root node_modules directory."
```
rahmaniaam, line 39: is there a possibility where a dependency is already in the packages node_modules instead of root?

rtarcr: Yeah, I guess it can be possible for a bundleDependency to be already in the sub-package node_modules. In that case the script would have to be adjusted to account for this. For the current use case I think this is fine though

rtarcr:
> Looks good! Does the `@aws/lsp-core` package live on inside `root/node_modules/@aws/aws-lsp-yaml/node_modules` after installing or will there be a `root/node_modules/@aws/lsp-core` in wherever the package is installed? I imagine it's the former so it's exclusively used only by `aws-lsp-yaml`

Good question. Not sure what is the behavior tbh, but I believe it should not matter in the scope of these changes

rahmaniaam (approved):

rtarcr:
Closing this due to dependencies of bundled dependencies not getting installed automatically, which require more custom scripts for this solution

PR #299, opened 2024-06-12
Move aws-lsp-fqn as a local dependency for now

makenneth:
## Description

Releasing `@aws/lsp-fqn` means we would be releasing an incomplete package since the browser bundle is incomplete. So we can move the FQN functionality as part of CWSPR server for now for simplicity.

I was able to bundle the artifacts and run them in VS.

## License


server/aws-lsp-codewhisperer/package.json
```diff
@@ -13,11 +13,12 @@
         "node": ">=18.0.0"
     },
     "workspaces": [
-        "src.gen/@amzn/codewhisperer-streaming"
+        "src.gen/@amzn/codewhisperer-streaming",
+        "src.gen/aws-lsp-fqn"
```
rtarcr, line 17: This brings another issue because FQN package is not published to NPMJS, and it is not bundled as part of CodeWhisperer server. 

We are having the same issue with the streaming client, but are planning to publish it to NPMJS to fix it. 

I think it is better to make it a folder in the CodeWhisperer server instead of a sub-package to fix this for now

makenneth: Moving the code out of a package would break the environment split, which would break at least one of the environments. 

I could move this as a subfolder `language-server-runtimes/runtime` package and import the files specific to the environment (i.e. node for standalone) and pass down as a feature, which was what we planned to do long term. That way we don't have to maintain a separate package

PR #302, opened 2024-06-13
Define quick actions dynamically

ege0zcan:
## Problem
The chat client shouldn't have to wait for the language client initialisation to finish before rendering the UI. Chat client should render the chat UI and then once the language client initialisation is done, the chat client should update the UI configuration to reflect the new information obtained during initialisation. For now, this information is only the available quick actions 

## Solution
Created new contract between ide extension and chat client that allows the ide extension to send a message about the initialization result to the chat client. The chat client then uses this information to update the default tab data for existing and any new tabs. 

## License

### DIFF
--- a/chat-client/src/client/chat.ts
+++ b/chat-client/src/client/chat.ts
@@ -11,6 +11,8 @@ import {
     SEND_TO_PROMPT,
     SendToPromptMessage,
     UiMessage,
+    QUICK_ACTIONS_OPTIONS,
+    QuickActionsOptionsMessage,
 } from '@aws/chat-client-ui-types'
 import {
     CHAT_REQUEST_METHOD,
@@ -35,12 +37,13 @@ import {
     TabChangeParams,
     TabRemoveParams,
 } from '@aws/language-server-runtimes-types'
-import { MynahUIDataModel } from '@aws/mynah-ui'
+import { MynahUI, MynahUIDataModel } from '@aws/mynah-ui'
 import { ServerMessage, TELEMETRY, TelemetryParams } from '../contracts/serverContracts'
 import { ENTER_FOCUS, EXIT_FOCUS } from '../contracts/telemetry'
 import { Messager, OutboundChatApi } from './messager'
 import { InboundChatApi, createMynahUi } from './mynahUi'
 import { TabFactory } from './tabs/tabFactory'
+import { MynahUITabStoreModel } from '@aws/mynah-ui/dist/static'
 
 const DEFAULT_TAB_DATA = {
     tabTitle: 'Chat',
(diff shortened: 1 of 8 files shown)


chat-client/src/client/chat.ts
```diff
@@ -35,12 +37,13 @@ import {
...
     TabRemoveParams,
 } from '@aws/language-server-runtimes-types'
-import { MynahUIDataModel } from '@aws/mynah-ui'
+import { MynahUI, MynahUIDataModel } from '@aws/mynah-ui'
 import { ServerMessage, TELEMETRY, TelemetryParams } from '../contracts/serverContracts'
 import { ENTER_FOCUS, EXIT_FOCUS } from '../contracts/telemetry'
 import { Messager, OutboundChatApi } from './messager'
 import { InboundChatApi, createMynahUi } from './mynahUi'
 import { TabFactory } from './tabs/tabFactory'
+import { MynahUITabStoreModel } from '@aws/mynah-ui/dist/static'
```
viktorsaws, line 46: nit: this type should be exported from MynahUI to avoid import from nested directory.

client/vscode/src/chatActivation.ts
```diff
@@ -28,6 +28,23 @@ export function registerChat(languageClient: LanguageClient, extensionUri: Uri)
         } // Webview options
     )
 
+    // Listen for Initialize handshake from LSP server to register quick actions dynamically
+    languageClient.onDidChangeState(({ oldState, newState }) => {
+        if (oldState === State.Starting && newState === State.Running) {
```
viktorsaws, line 33: Have you verifies that this is doable in Visual Studio client?

chat-client/src/client/chat.ts
```diff
@@ -81,6 +82,17 @@ export const createChat = (
             case ERROR_MESSAGE:
                 mynahApi.showError((message as ErrorMessage).params)
                 break
+            case QUICK_ACTIONS_OPTIONS:
```
viktorsaws, line 85: As I mentioned in related PR https://github.com/aws/language-server-runtimes/pull/140/files#r1639904723, I think this should not be limited to Quick Actions only, and be made as more generic extendable object. This will allow to extend this mechanism in the future

viktorsaws: There is bunch on hard-coded string, that setup Chat panel messages specific to Amazon Q. Do we need to allow integrating clients to configure them at start time? Not sure if mechanism will be same though

chat-client/src/client/chat.ts
ege0zcan: oh yes, i missed the auto import here

chat-client/src/client/chat.ts
```diff
@@ -51,12 +57,10 @@ const DEFAULT_TAB_DATA = {
 
 type ChatClientConfig = Pick<MynahUIDataModel, 'quickActionCommands'>
 
-export const createChat = (
-    clientApi: { postMessage: (msg: UiMessage | ServerMessage) => void },
-    config?: ChatClientConfig
```
viktorsaws, line 56: Won't we need to support passing configuration from client at initialisation time? If yes, we should keep `config` param in function signature.

viktorsaws (approved):

chat-client/src/client/chat.ts
ege0zcan: Made this more generic and named it CHAT_OPTIONS to allow for more configuration to be passed in the future

chat-client/src/client/chat.ts
ege0zcan: newer versions of mynah now export this

ege0zcan:
Followed up on: https://github.com/aws/language-servers/pull/326, closing this PR

PR #305, opened 2024-06-14
fix: partiql - don't report diagnostics for quoted identifiers

akoreman:
## Problem

Queries like 

```
SELECT * 
FROM "TableName"."IndexName" 
```
should not result in diagnostics. (see https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ql-reference.select.html)

Currently these result in diagnostics with the error `InvalidInput("\")` which is coming from the extra `\` we add in `normalizeQuery`. The query is directly send to the wasm module [here](https://github.com/aws/language-servers/blob/main/server/aws-lsp-partiql/src/partiql-parser-wasm/partiql_playground.js#L181) so the extra escaping of the `"` doesn't seem to be necessary. 

Same bug was present in the partiql-playground package using the same WASM module and was fixed in the same way in https://github.com/partiql/partiql-rust-playground/pull/16.

## Solution

Removed the replacement of `\"` by `\\"` in the preparing of the document before sending it to the wasm module.

## License

### DIFF
--- a/server/aws-lsp-partiql/src/server/language-service.ts
+++ b/server/aws-lsp-partiql/src/server/language-service.ts
@@ -5,7 +5,7 @@ import { initSync, parse_as_json } from '../partiql-parser-wasm/partiql_playgrou
 import { convertObjectToParserError } from './error-parsing/parser-errors'
 
 export function normalizeQuery(data: string): string {
-    return data != null ? data.replace(/\"/g, '\\"') : ''
+    return data != null ? data : ''
 }
 
 export function createPartiQLLanguageService() {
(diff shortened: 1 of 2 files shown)


server/aws-lsp-partiql/src/server/language-service.ts
```diff
@@ -5,7 +5,7 @@ import { initSync, parse_as_json } from '../partiql-parser-wasm/partiql_playgrou
 import { convertObjectToParserError } from './error-parsing/parser-errors'
 
 export function normalizeQuery(data: string): string {
-    return data != null ? data.replace(/\"/g, '\\"') : ''
```
kmile, line 8: Is there a (now-failing) test case for the escaping of quotes and the situation for which this was necessary?

akoreman: This was copied from the partiql-playground (first party package using the same parser) which had the same code, they also removed it in https://github.com/partiql/partiql-rust-playground/pull/16 after we discussed this bug with the team maintaining that.

There wasn't a test for the escaping before, I added one to verify that quoted identifiers don't give diagnostics.

imykhai (approved):

PR #306, opened 2024-06-15
Send Generic command to new tab if current tab is busy

makenneth:
## Problem
If user has an outgoing request and use the generic commands, the chat history becomes a mess. 

Note: 
- Change `sinon.stub` -> `sandbox.stub` in `chat.test` since there is only a `sandbox.restore()`, which affected the other test file.
- Changed `mynah.test` to use `beforeEach` setup as it is difficult to debug shared setup
- Clear all tabs after each test since both test files seem to share the same UI (with the same tab limit)

## Solution
vscode current opens up a new tab if there is an outgoing request. This PR mimics their behavior.

## License

### DIFF
--- a/chat-client/src/client/chat.test.ts
+++ b/chat-client/src/client/chat.test.ts
@@ -16,29 +16,34 @@ import { createChat } from './chat'
 import sinon = require('sinon')
 import { TELEMETRY } from '../contracts/serverContracts'
 import { ERROR_MESSAGE_TELEMETRY_EVENT, SEND_TO_PROMPT_TELEMETRY_EVENT } from '../contracts/telemetry'
+import { MynahUI } from '@aws/mynah-ui'
 
 describe('Chat', () => {
     const sandbox = sinon.createSandbox()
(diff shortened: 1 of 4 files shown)


chat-client/src/client/chat.test.ts
```diff
@@ -70,17 +73,17 @@ describe('Chat', () => {
...
     it('publishes tab added event, when UI tab is added', () => {
-        const mynahUi = createChat(clientApi)
         const tabId = mynahUi.updateStore('', {})
 
         assert.calledWithMatch(clientApi.postMessage, {
             command: TAB_ADD_NOTIFICATION_METHOD,
             params: { tabId: tabId },
         })
+
+        mynahUi.getAllTabs()
```
ege0zcan, line 83: nit: why is this call needed?

ege0zcan (approved):

chat-client/src/client/chat.test.ts
makenneth: Oops, I left it in by accident

PR #321, opened 2024-06-19
Fixing Windows problems - prettier CRLF and linux cp tool

imykhai:
## Problem
The package fails builds on Windows. One of the reason is that we have `cp` in our npm scripts which only works on linux. 
Another problem was that we ran `prettier -- check` in some subpackages. In windows git automatically converts linux's `LF` line endings into Windows' `CRLF` when pulling files. It works generally, because git also converts them back to the linux's `LF` before staging or committing the files. And our husky git hooks also run prettier before committing so that prettier will convert the line endings to linux's LF even if git wouldn't. 
But it all breaks because we had npm script that ran `prettier check` on running. And the command fails on windows as it notices `CRLF` line endings which are not allowed.
## Solution
- Use shx tool to make linux-specific calls cross platform
- Run prettier check only as part of the Linux-based GitHub action and not as part of tests 
- Also used `npm ci` instead of `npm install` in the windows GitHub actions (related to https://github.com/aws/language-servers/pull/317)

## License

### DIFF
--- a/chat-client/package.json
+++ b/chat-client/package.json
@@ -11,9 +11,8 @@
     "license": "Apache-2.0",
     "scripts": {
         "compile": "tsc --build && npm run package",
-        "test:prettier": "prettier . --check",
         "test:unit": "ts-mocha -b 'src/**/*.test.ts'",
-        "test": "npm run test:prettier && npm run test:unit",
+        "test": "npm run test:unit",
         "fix:prettier": "prettier . --write",
         "package": "webpack"
     },
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "scripts": {
         "prepare": "husky install .husky",
         "lint": "eslint app/ chat-client/ client/ core/ script/ server/ --ext .ts,.tsx --ignore-path .gitignore && prettier --check .",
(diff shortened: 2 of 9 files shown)


chat-client/package.json
```diff
@@ -11,9 +11,8 @@
     "license": "Apache-2.0",
     "scripts": {
         "compile": "tsc --build && npm run package",
-        "test:prettier": "prettier . --check",
```
ege0zcan, line 14: doesn't this mean local builds and tests will ignore formatting errors? Or so we still have a mechanism against this

imykhai: We now rely on a different mechanism for these - precommit hooks. So locally you won't be able to commit unless everything is formatted and pretty.

package.json
```diff
@@ -39,6 +40,7 @@
         "husky": "^9.0.11",
         "prettier": "^3.2.5",
         "pretty-quick": "^4.0.0",
+        "shx": "^0.3.4",
```
viktorsaws, line 43: oh, this is nice find

viktorsaws (approved):

PR #323, opened 2024-06-19
Send auth follow up item back when q returns auth error

makenneth:
## Description
Return an auth follow up action when Q api returns missing token errors
- Still testing the errors out
- Also, might need to consider missing scopes

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
@@ -2,7 +2,7 @@ import {
     GenerateAssistantResponseCommandInput,
     GenerateAssistantResponseCommandOutput,
 } from '@amzn/codewhisperer-streaming'
-import { TabChangeParams, chatRequestType } from '@aws/language-server-runtimes/protocol'
+import { LSPErrorCodes, TabChangeParams, chatRequestType } from '@aws/language-server-runtimes/protocol'
 import {
     CancellationToken,
     Chat,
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -81,21 +81,23 @@ export class ChatController implements ChatHandlers {
...
+            const authFollowType = getAuthFollowUpType(err)
 
+            if (authFollowType) {
+                this.#log(`Q auth error: ${getErrorMessage(err)}`)
+
+                return createAuthFollowUpResult(authFollowType)
+            }
+
+            this.#log(`Q api request error ${err instanceof Error ? err.message : 'unknown'}`)
             return new ResponseError<ChatResult>(
```
ege0zcan, line 98: Is my following understanding correct: The expectation in this error case is that the IDE extension uses the `sendError` contract of the chat-client to show an error on the chat-ui

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -2,7 +2,7 @@ import {
     GenerateAssistantResponseCommandInput,
     GenerateAssistantResponseCommandOutput,
 } from '@amzn/codewhisperer-streaming'
-import { TabChangeParams, chatRequestType } from '@aws/language-server-runtimes/protocol'
+import { LSPErrorCodes, TabChangeParams, chatRequestType } from '@aws/language-server-runtimes/protocol'
```
ege0zcan, line 5: Why do we use LSPErrorCodes, for simplicity? If they don't provide a value I would argue we shouldn't use them

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
makenneth: This makes handling exception a lot easier at least on VS side because the LSP library convert these into meaningful exceptions

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
makenneth: Yes, that is correct

PR #324, opened 2024-06-19
Emit initial add tab event

makenneth:
## Description

If toolkit has to keep track of the tabs, it is difficult to do so currently because there isn't an event for the initial tab. So chat-client should emit an event on start up.

I think setting trigger to "click" might make sense, but unsure about this.

## License

### DIFF
--- a/chat-client/src/client/chat.test.ts
+++ b/chat-client/src/client/chat.test.ts
@@ -15,8 +15,13 @@ import { assert } from 'sinon'
 import { createChat } from './chat'
 import sinon = require('sinon')
 import { TELEMETRY } from '../contracts/serverContracts'
-import { ERROR_MESSAGE_TELEMETRY_EVENT, SEND_TO_PROMPT_TELEMETRY_EVENT } from '../contracts/telemetry'
+import {
+    ERROR_MESSAGE_TELEMETRY_EVENT,
+    SEND_TO_PROMPT_TELEMETRY_EVENT,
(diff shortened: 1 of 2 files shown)


chat-client/src/client/chat.test.ts
```diff
@@ -39,8 +44,23 @@ describe('Chat', () => {
         })
     })
 
-    it('publishes ready event, when initialized', () => {
-        assert.calledOnceWithExactly(clientApi.postMessage, { command: READY_NOTIFICATION_METHOD })
+    it('publishes ready event and initial tab add event, when initialized', () => {
+        sinon.assert.callCount(clientApi.postMessage, 3)
+
+        sinon.assert.calledWithExactly(clientApi.postMessage.getCall(0), { command: READY_NOTIFICATION_METHOD })
```
ege0zcan, line 50: nit: I think the following syntax is also possible and I personally prefer it on tests:

`clientApi.postMessage.firstCall`
`clientApi.postMessage.secondCall`
`clientApi.postMessage.thirdCall`


So the assertions would look like:   `sinon.assert.calledWithExactly(clientApi.postMessage.firstCall, { command: READY_NOTIFICATION_METHOD })`

ege0zcan:
Approved with a very small comment, the failing windows tests are being worked on right now so feel free to ignore them.

chat-client/src/client/chat.test.ts
makenneth: Yeah, that's more readable

PR #326, opened 2024-06-20
Update dependencies

ege0zcan:
## Description

This PR updates dependencies to the newly published latest versions. Most importantly this PR integrates the latest chat encryption related changes introduced in https://github.com/aws/language-server-runtimes/pull/145/files and updates the sample vscode extension to handle the necessary encryption and decryption duties 

## License


makenneth (approved):

chat-client/package.json
```diff
@@ -17,9 +17,9 @@
         "package": "webpack"
     },
     "dependencies": {
-        "@aws/chat-client-ui-types": "0.x.x",
-        "@aws/language-server-runtimes-types": "0.x.x",
-        "@aws/mynah-ui": "^4.11.x"
+        "@aws/chat-client-ui-types": "^0.0.6",
```
viktorsaws, line 20: I think we shouldn't change these values from `x.x` suffix - iirc Ivan discovered that when using caret (^) syntax in npm will result in not upgrading patch version. This effectively pins version to `0.0.6` untl manually upgraded

client/vscode/src/chatActivation.ts
```diff
@@ -207,3 +272,45 @@ function insertTextAtCursorPosition(text: string) {
...
+
+    if (!result.payload) {
+        throw new Error('JWT payload not found')
+    }
+    return result.payload as T
+}
+
+function isEncryptedMessage(message: string, algorithm: string, encoding: string) {
+    // Check if the message has five parts separated by periods
+    const parts = message.split('.')
```
viktorsaws, line 291: nit: could you add comment about used encryption mechanism here?

viktorsaws (approved):

chat-client/package.json
ege0zcan: 0.x.x has the same effect as caret since the jump from 0.0.6 to 0.0.7 is also considered a major version. We no longer need the 0.x.x type values and this makes it easier and more explicit when we want to bump  to newer versions

client/vscode/src/chatActivation.ts
ege0zcan: Yeah, this function is pretty barebones right now. I will expand and explain this as followup

chat-client/package.json
viktorsaws: ah, ok, I must have remembered that wrong

PR #327, opened 2024-06-21
Update Codewhisperer streaming client package

viktorsaws:
## Problem

Codewhisperer streaming client package used in Codewhisperer LSP is 2 months old and outdated.

## Solution

* Updated `@amzn/codewhisperer-streaming` package to latest version used in AWS Toolkit https://github.com/aws/aws-toolkit-vscode. Tested Chat functionality with sample VSCode extension.
* [`Reference.recommendationContentSpan`](https://github.com/aws/language-servers/pull/327/files#diff-21a75ff27b821e42265decb7a14a13858a5244ba32b4fb3eca9cc55ba3e4c8ecR279-R280) interface is not compatible with Chat [`ReferenceTrackerInformation`](https://github.com/aws/language-server-runtimes/blob/4480b87a81c84dda229538fed7cbf165a44dc971/types/chat.ts#L36-L39) class anymore: `start` and `end` fields are defined as optional now. I had to adjusted [`chatEventParser`](https://github.com/aws/language-servers/pull/327/files#r1648984413) to handle this change. Correctness needs to be checked.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatEventParser.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatEventParser.ts
@@ -40,6 +40,12 @@ export class ChatEventParser implements ChatResult {
     static mapReferenceData(reference: Reference): ReferenceTrackerInformation {
         return {
             ...reference,
+            recommendationContentSpan: reference.recommendationContentSpan
+                ? {
+                      start: reference.recommendationContentSpan.start ?? 0,
+                      end: reference.recommendationContentSpan.end ?? 0,
+                  }
+                : undefined,
             information: ChatEventParser.getReferencedInformation(reference),
         }
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -117,8 +117,7 @@
         },
         "app/aws-lsp-partiql-binary/node_modules/@aws/lsp-partiql": {
             "version": "0.0.1",
-            "resolved": "https://registry.npmjs.org/@aws/lsp-partiql/-/lsp-partiql-0.0.1.tgz",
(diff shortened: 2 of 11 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/chatEventParser.ts
```diff
@@ -40,6 +40,10 @@ export class ChatEventParser implements ChatResult {
     static mapReferenceData(reference: Reference): ReferenceTrackerInformation {
         return {
             ...reference,
+            recommendationContentSpan: {
+                start: reference.recommendationContentSpan?.start || 0,
+                end: reference.recommendationContentSpan?.end || 0,
```
viktorsaws, line 45: Set defaults, as `Reference.recommendationContentSpan` interface was changed, making `start` and `end` fields optional.

server/aws-lsp-codewhisperer/src/language-server/chat/chatEventParser.ts
```diff
@@ -40,6 +40,12 @@ export class ChatEventParser implements ChatResult {
     static mapReferenceData(reference: Reference): ReferenceTrackerInformation {
         return {
             ...reference,
+            recommendationContentSpan: reference.recommendationContentSpan
+                ? {
+                      start: reference.recommendationContentSpan.start ?? 0,
+                      end: reference.recommendationContentSpan.end ?? 0,
+                  }
+                : undefined,
```
viktorsaws, line 48: Set defaults, as `Reference.recommendationContentSpan` interface was changed, making `start` and `end` fields optional.

This logic needs to be verified or changed before merging

viktorsaws: @ege0zcan @makenneth could you check this approach and change it if required, since you have more knowledge about expected business logic?

makenneth: I verified it and it is fine to have default values.

makenneth (approved):

PR #328, opened 2024-06-21
Optimise packaging

viktorsaws:
## Problem

1. We use deprecated unsecure `vercel/pkg` package and mention it as main way of consuming bundled runtime servers, which also takes a lot of time (3-4 minutes per run)
2. Our github workflows take ~7 minutes in Linux and ~13 minutes on Windows runners to complete. Npm commands called in workflows are wasteful: npm commands in workflow call `npm install` and `npm run compile` under the hood 2 or 3 times, which adds up to workflow run times.

## Solution

To reduce workflow run times:
1. Removed using pkg for producing binaries
2. Removed redundant npm script calls during workflow runs

## Result
Run time for all workflows is 3x times faster now, taking ~4 minutes.

## License


ege0zcan:
Nice change, should we also consider renaming the folder names from `-binary` to something like `-standalone` or something else

app/aws-lsp-json-binary/package.json
```diff
@@ -8,9 +8,9 @@
     },
     "scripts": {
         "compile": "tsc --build",
-        "package-x64": "pkg --targets node18-linux-x64,node18-win-x64,node18-macos-x64 --out-path bin --compress GZip .",
+        "pkg": "pkg --targets node18-linux-x64,node18-win-x64,node18-macos-x64 --out-path bin --compress GZip .",
```
ege0zcan, line 11: so we still want to keep this usage of pkg right?

app/aws-lsp-yaml-json-webworker/package.json
```diff
@@ -6,8 +6,8 @@
         "compile": "tsc --build",
         "clean": "rm -rf dist && rm -rf node_modules && rm -rf build",
         "build": "tsc",
-        "prepublishOnly": "npm run build && npm run compile && npm run bundle",
-        "bundle": "webpack",
+        "prepublishOnly": "npm run build && npm run compile && npm run package",
+        "package": "webpack",
```
ege0zcan, line 10: do we want to keep the `package` name in the scripts, why not rename all usages of it with something like `bundle` instead?

ege0zcan: is it because we still have a few subpackages that use `pkg` and do actual packaging?

ege0zcan:
We also need to update integration tests that make use of executables since they will no longer work: 

https://github.com/aws/language-servers/blob/main/app/hello-world-lsp-runtimes/src/tests/helloWorldServerInteg.test.ts#L20-L40

we should instead spawn a child process like `node /build/server.js`

app/aws-lsp-json-binary/package.json
ege0zcan: I think if we are only keeping it in for testing, we can remove these and use the bundled js files for testing

app/hello-world-lsp-runtimes/README.md
```diff
@@ -24,26 +24,17 @@ npm run bundle
...
-    ```
-
-* To create only javascript bundled files with webpack, run:
+* To create javascript bundled files with webpack, run:
     ```bash
     npm run webpack
     ```
 
-Result artifacts can be found in `/out` and `/bin` directories.
+Result artifacts can be found in `/out` and `/build` directories.
```
ege0zcan, line 32: nit: the following file tree doesn't show the build folder

app/aws-lsp-yaml-json-webworker/package.json
viktorsaws: I wanted to keep naming consistent with older variants, I don't think it matters much and we used `package` for most of such commands

ege0zcan (approved):

app/aws-lsp-json-runtimes/src/tests/jsonServerCFInteg.test.ts
```diff
@@ -20,25 +19,12 @@ describe('Test JsonServer with CloudFormation schema', () => {
     let process: ChildProcessWithoutNullStreams
     let endpoint: JSONRPCEndpoint
     let client: LspClient
-    let binaryFile = 'lsp-json-binary'
+    // let runtimeFile = 'lsp-json-runtimes'
```
ege0zcan, line 22: nit: leftover comment

ege0zcan (approved):

PR #329, opened 2024-06-21
Ensure unque tab id and allow follow up text to be empty

makenneth:
## Description
This PR has two small fixes:

1. Unique Tab Id
    - VS has a case where the connection to the server would not close but the UI will get reinitialized. The first tab being constant means user can still access existing session. And we don't have time to trigger through the UI. Maybe a new reset notification could be helpful.
2. Allow follow up text to be empty if explicitly set to `""`. The default text doesn't fit auth follow up. I think we'd want to it to be empty to sort of match the expected UX (which is a button in chat). Server is already adding a default message for chat

## License


ege0zcan (approved):
Changes look good to me, I believe you need to also push your package lock changes to get the github actions to pass

ege0zcan (approved):

ege0zcan (approved):

PR #330, opened 2024-06-21
Add Chat Telemetry

makenneth:
## Description
This PR add chat telemetry handling. It can be broken into two main parts:
- 1. chat request related telemetry
- 2. client side telemetry, which are received via `onClientTelemetry`, which are forwarded by ide to chat-client to add message id then sent back.

The flow for chat request related telemetry is as follows:

- AddMessageNotification received - register trigger type (except for follow up)
- ChatRequest received 
   - for follow up - reuse previous trigger
   - for others - get new trigger
- Build input with trigger
- Make API call
- After each request, the trigger should be stored as last
- Emit metrics

## Note 
- `ModifyCode` will have to be in client side if it is P0. We have no access to the cursor position in `insertToCursorPosition` params.

## Testing (in VS)
- copy 
- insert to cursor
- conversation focus
- command (clear and help)
- link clicks
- follow up (able to use previous trigger)
- chat message

## License


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -171,35 +208,84 @@ export class ChatController implements ChatHandlers {
...
+                return {}
+        }
+    }
+
+    async #getTriggerContext(params: ChatParams, metric: Metric<CombinedConversationEvent>) {
+        const lastMessageTrigger = this.#telemetryController.getLastMessageTrigger(params.tabId)
+
+        let triggerContext: TriggerContext
+
+        // this is the only way we can detect a follow up action
```
ege0zcan, line 255: Can https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts#L132 this listener help with this?

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -171,35 +208,84 @@ export class ChatController implements ChatHandlers {
...
 
     onQuickAction(params: QuickActionParams, _cancellationToken: CancellationToken) {
         switch (params.quickAction) {
             case QuickAction.Clear: {
                 const sessionResult = this.#chatSessionManagementService.getSession(params.tabId)
 
+                this.#telemetryController.emitChatMetric({
+                    name: ChatTelemetryEventName.RunCommand,
+                    data: {
+                        // TODO verify
```
ege0zcan, line 222: Is this still a todo?

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
makenneth: I tried to, but the chat prompt message comes first now. It will also be harder to follow if it relies on orchestration of two different handlers

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
makenneth: No it is not, I will remove this

PR #316, opened 2024-06-18
Create scripts to bundle dependencies

rtarcr:
## Problem
Bundled dependencies are not included in monorepo. Also their dependencies do not get installed 

## Solution
Create scripts to bundle them and install their dependencies 

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #332, opened 2024-06-24
Add initial packing test workflow for Codewhisperer package

viktorsaws:
## Problem
- We need to test that packed npm packages can be consumed before publishing them to npm. 
- The transitive dependencies of bundled dependencies (see https://github.com/aws/language-servers/pull/316) do not get bundled or installed. Only dependencies that are present in the local node_modules get included during packing. 
- There is an issue with `@aws-sdk/util-utf8-browser` package that doesn't have dist folders included in the bundle, which fails the initial workflows of this PR. 

## Solution

- Add github workflow to test CodeWhisperer server npm package.
- Add postinstall script that goes into bundled dependencies and does npm install there. 

## License


server/aws-lsp-codewhisperer/package-lock.json
rahmaniaam, line 1: Is deleting the lockfile necessary? published packages shouldn't know anything about it all

server/aws-lsp-codewhisperer/script/install_transitive_dep.js
```diff
@@ -0,0 +1,28 @@
...
+
+for (let dependency of bundleDependencies) {
+    const dependencyPath = path.join('node_modules', dependency)
+
+    // Check if the dependency exists in node_modules to avoid monorepo setup
+    if (fs.existsSync(dependencyPath)) {
+        console.log(`Bundled dependency '${dependency}' found. Installing its dependencies...`)
+        try {
+            process.chdir(dependencyPath)
+            const npmInstallCommand = process.platform === 'win32' ? 'npm.cmd install' : 'npm install'
```
ege0zcan, line 19: what is the win32 platform? Is it used by any github action environments?

rtarcr: I added this to make the package compatible with Windows, otherwise installation of this package would not work

rahmaniaam (approved):

server/aws-lsp-codewhisperer/package-lock.json
rtarcr: I think it was added by mistake in the previous PR and now it is getting removed. We should have a single lockfile in a monorepo afaik

ege0zcan (approved):

PR #339, opened 2024-06-24
Set message can be voted; fix clear command

makenneth:
## Description

- A spinner is showing up after the screen is cleared.  We should just terminate early in the chat-client chat request logic for `/clear` command
- Thumbs up and down are not showing up due to missing `canBeVoted` - we should set `canBeVoted` to true by default unless otherwise specified

As a bonus, we will add a default `What can Amazon Q help me with?` for `/help` command like vscode does

## License


shruti0085 (approved):

ege0zcan (approved):
We already have a followup tasks to enable servers to clear chat, I will also create a followup task for us to allow for servers to add prompts on behalf of user for the `/help` scenario. This will ensure we don't have hardcoded quick action logic in the chat client

PR #340, opened 2024-06-25
Expand chat telemetry types

makenneth:
## Description

- Add chat telemetry type according to definition
- Add also definition for the client side telemetry. Ideally we can just import all of them in the future
- Add a metric class which holds the telemetry (maybe we can consolidate with the trigger that I will be adding in the other PR when we have time)

(Feel free to review https://github.com/aws/language-servers/pull/330 directly if you are feeling like it)

## License


server/aws-lsp-codewhisperer/src/language-server/telemetry/types.ts
```diff
@@ -102,23 +102,110 @@ export enum ChatTelemetryEventName {
     ExitFocusChat = 'ExitFocusChat',
     EnterFocusConversation = 'EnterFocusConversation',
     ExitFocusConversation = 'ExitFocusConversation',
+    StartConversation = 'StartConversation',
+    InteractWithMessage = 'InteractWithMessage',
+    AddMessage = 'AddMessage',
+    RunCommand = 'RunCommand',
+    MessageResponseError = 'MessageResponseError',
```
makenneth, line 109: These names are incorrect, I think they should be `amazonq_....`. I will update that in https://github.com/aws/language-servers/pull/330

ege0zcan (approved):

PR #344, opened 2024-06-25
Make sure clear quick action doesn't get stuck loading

ege0zcan:
## Problem
/clear quick action caused a `generating your answer` chat item to be added that would not go away because the server would respond with an empty object causing the chat client to create an empty chat item. This would cause mynah to create a chat item that assumed it was going to be filled with data and therefore had the `generating your answer` string

## Solution
chat client will not try to create a chat item if the server response is an empty object 
## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -46,11 +46,6 @@ export const handleChatPrompt = (
             mynahUi.updateStore(tabId, {
                 chatItems: [],
             })
-
-            mynahUi.updateStore(tabId, {
-                loadingChat: false,
-                promptInputDisabledState: false,
-            })
         } else if (prompt.command === '/help') {
             userPrompt = DEFAULT_HELP_PROMPT
         }
@@ -69,7 +64,6 @@ export const handleChatPrompt = (
         // Send chat prompt to server
         messager.onChatPrompt({ prompt, tabId }, triggerType)
     }
-
     // Add user prompt to UI
     mynahUi.addChatItem(tabId, {
         type: ChatItemType.PROMPT,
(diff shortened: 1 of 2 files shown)


makenneth (approved):

chat-client/src/client/mynahUi.ts
```diff
@@ -46,11 +46,6 @@ export const handleChatPrompt = (
             mynahUi.updateStore(tabId, {
                 chatItems: [],
             })
-
-            mynahUi.updateStore(tabId, {
-                loadingChat: false,
-                promptInputDisabledState: false,
```
makenneth, line 52: From testing in VS, I found that I need these two at least for. But it can be combined with the above `updateStore`

makenneth: I'll test it out a little more

makenneth: This doesn't seem to work for me in VS.

makenneth (approved):
Actually I tested again, this works. The json serializer in VS just keeps adding a default boolean value in, which made it not empty. Ideally, we don't have to use that empty condition, but it is fine

chat-client/src/client/mynahUi.ts
```diff
@@ -282,6 +276,11 @@ export const createMynahUi = (messager: Messager, tabFactory: TabFactory): [Myna
             return
         }
 
+        // If chat response from server is an empty object don't do anything
```
dogusata, line 279: Is it enough to also not add a card on /clear?

dogusata (approved):
If incoming data is empty, new statement check avoids adding a card. Tested and works

PR #354, opened 2024-06-27
Add Chat modify code telemetry

makenneth:
## Description 

- Track document position and inserted code when user uses the insert code to cursor functionality and insert to a queue
- When an entry has been added for at least five minutes, we use use fastest-levenshtein to calculate edit distance between the current text in the same position and the original inserted text.
 
We are able to get the E2E flow working in VS by forcing the cursor state and document identifier through, but they should be added to the interface later on.

✅ Tested E2E flow in Visual Studio 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
@@ -8,9 +8,15 @@ import {
     InteractWithMessageEvent,
 } from '../../telemetry/types'
 import { Features, KeysMatching } from '../../types'
-import { ChatUIEventName, RelevancyVoteType, isClientTelemetryEvent } from './clientTelemetry'
+import {
+    ChatUIEventName,
+    InsertToCursorPositionParams,
+    RelevancyVoteType,
+    isClientTelemetryEvent,
+} from './clientTelemetry'
 import { UserIntent } from '@amzn/codewhisperer-streaming'
 import { TriggerContext } from '../contexts/triggerContext'
+import { AcceptedSuggestionEntry, CodeDiffTracker } from '../../telemetry/codeDiffTracker'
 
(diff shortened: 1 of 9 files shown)


server/aws-lsp-codewhisperer/src/language-server/telemetry/codeDiffTracker.ts
```diff
@@ -0,0 +1,152 @@
+import { distance } from 'fastest-levenshtein'
+import { Position } from 'vscode-languageserver-textdocument'
```
ege0zcan, line 2: nit: why not import from `'@aws/language-server-runtimes-types'`

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
```diff
@@ -200,6 +224,27 @@ export class ChatTelemetryController {
         )
     }
 
+    #enqueueCodeDiffEntry(params: InsertToCursorPositionParams) {
+        const documentUri = params.textDocument?.uri
```
ege0zcan, line 228: Currently this is not part of the interface therefore this field will be empty. Is it ok to have it empty until we figure out how we want to pass this information

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/telemetry/codeDiffTracker.ts
makenneth: I can import from server interface instead. (we don't have the runtimes-types in the dependency)

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
makenneth: We are able to pass these through via telemetry. Maybe because it doesn't have defined types on the protocol side?

PR #355, opened 2024-06-27
Ensure errors are useful message for security scan

farisab1:
## Problem
Security scan error messages were vague and did not provide enough information on what the error being returned from LSP was

## Solution
Update wording on error messages to describe the issue rather than returning internal objects

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
@@ -49,23 +49,29 @@ export const SecurityScanServerToken =
                 if (!credentialsProvider.hasCredentials('bearer')) {
                     throw new Error('Credentials provider does not have bearer token credentials')
                 }
+
+                logging.log(`Parameters provided: ${JSON.stringify(params)}`)
+
                 if (!params.arguments || params.arguments.length === 0) {
-                    throw new Error(`Error: Incorrect parameters parameters provided`)
+                    throw new Error(`Error: Invalid data.`)
                 }
                 const [arg] = params.arguments
+
+                logging.log(`Arguments provided: ${JSON.stringify(arg)}`)
+
                 const { ActiveFilePath: activeFilePath, ProjectPath: projectPath } = parseJson(arg)
                 if (!activeFilePath) {
-                    throw new Error('Error: File to scan is missing')
+                    throw new Error('Error: File to scan is missing.')
                 }
 
                 if (!projectPath) {
-                    throw new Error('Error: Project is missing')
+                    throw new Error('Error: Project is missing.')
                 }
                 const activeFilePathUri = pathToFileURL(activeFilePath).href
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
```diff
@@ -47,21 +47,25 @@ export const SecurityScanServerToken =
             }
             try {
                 if (!credentialsProvider.hasCredentials('bearer')) {
-                    throw new Error('credentialsProvider does not have bearer token credentials')
+                    throw new Error('Credentials provider does not have bearer token credentials')
                 }
                 if (!params.arguments || params.arguments.length === 0) {
-                    throw new Error(`Incorrect params provided. Params: ${params}`)
+                    throw new Error(`Error: Incorrect parameters parameters provided`)
```
awschristou, line 53: Since the error text becomes user-facing, lets log (with `logging`) this error message, along with `${params}`. This will let us diagnose the problem and see what the input is.

Meanwhile, we can make this error text less "technical" by saying something like "Error: Invalid data"

server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
```diff
@@ -47,21 +47,25 @@ export const SecurityScanServerToken =
...
                 }
                 const [arg] = params.arguments
                 const { ActiveFilePath: activeFilePath, ProjectPath: projectPath } = parseJson(arg)
-                if (!activeFilePath || !projectPath) {
-                    throw new Error(`Error: file path or project path not provided. Params: ${params}`)
+                if (!activeFilePath) {
```
awschristou, line 57: In each of these failure cases, it would be good to log the params payload. As the client, we will want to diagnose by understanding what we did or didn't send over to the server.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
```diff
@@ -47,21 +47,31 @@ export const SecurityScanServerToken =
...
+
+                logging.log(`Parameters provided: ${JSON.stringify(params)}`)
+
                 if (!params.arguments || params.arguments.length === 0) {
-                    throw new Error(`Incorrect params provided. Params: ${params}`)
+                    throw new Error(`Error: Invalid data.`)
                 }
                 const [arg] = params.arguments
+
+                logging.log(`Arguments provided: ${JSON.stringify(arg)}`)
```
ege0zcan, line 60: Could this lead to sensitive data being logged or is it just path information?

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts
farisab1: This is just path information.

PR #356, opened 2024-06-28
Clean up logging and remove q data from response error

makenneth:
## Description

- Clean up some unneeded logging. 
- Remove partial response data from response error since they won't be encrypted.

## License


ege0zcan (approved):
LGTM but there seems to be failing tests,  for the `ChatController` `onChatPrompt` suite

PR #357, opened 2024-06-28
Fixing typo in /help quick action

farisab1:
## Problem
Typo in /help quick action where `selected code` was missing from the example questions.

image

## Solution
Update text to include the missing string

## License


shruti0085 (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/quickActions.ts
makenneth, line 1: Is this a regression in mynah-ui? I can't recall what it looked like before, but we should fix this for launch.

makenneth (approved):

ege0zcan (approved):

ege0zcan (approved):

PR #358, opened 2024-06-30
Filter out Git Ignored files from security scans

rbbarad:
## Problem

Security scans were not respecting .gitignore files. This is an AppSec concern, as users would not want git ignored files to be processed.

## Solution

This PR fixes the issue by adding logic to handle .gitignore files and filter out the necessary files from the security scan.

## Testing

- created unit tests
- tested behavior by running scans client side

## License


package.json
```diff
@@ -28,6 +28,7 @@
         "package": "npm run compile && npm run package --workspaces --if-present"
     },
     "dependencies": {
+        "@gerhobbelt/gitignore-parser": "^0.2.0-9",
```
ege0zcan, line 31: this shouldn't be a top level dependency

server/aws-lsp-codewhisperer/src/language-server/pathUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Returns `true` if path `p` is a descendant of directory `d` (or if they are
+ * identical).
+ *
+ * Only the logical structure is checked; the paths are not checked for
+ * existence on the filesystem.
+ *
+ * @param d  Path to a directory.
+ * @param p  Path to file or directory to test.
+ */
+export function isInDirectory(d: string, p: string): boolean {
```
ege0zcan, line 13: I guess something like the following won't work:

```
function isInDirectory(d, p) {
    // Resolve paths to absolute paths
    const resolvedDir = path.resolve(d);
    const resolvedPath = path.resolve(p);

    // Get the relative path from the directory to the path
    const relativePath = path.relative(resolvedDir, resolvedPath);

    // Check if the relative path starts with `..` or is equal to ''
    return !relativePath.startsWith('..') && !path.isAbsolute(relativePath);
}
```

But what about using this: https://github.com/domenic/path-is-inside/blob/master/lib/path-is-inside.js

server/aws-lsp-codewhisperer/src/language-server/pathUtils.ts
```diff
@@ -0,0 +1,51 @@
+import os = require('os')
+
+/**
+ * Returns `true` if path `p` is a descendant of directory `d` (or if they are
+ * identical).
+ *
+ * Only the logical structure is checked; the paths are not checked for
+ * existence on the filesystem.
+ *
+ * @param d  Path to a directory.
```
ege0zcan, line 10: nit: better to use explicit names like `potentialParent` and `pathToCheck`

server/aws-lsp-codewhisperer/src/language-server/dependencyGraph/dependencyGraph.ts
```diff
@@ -116,6 +117,25 @@ export abstract class DependencyGraph {
         return files.reduce((a, f) => a.concat(f), [])
     }
 
+    /**
+     * @param rootPath root folder to look for .gitignore files
+     * @returns list of files without those that are git ignored
+     */
+    async filterOutGitIgnoredFiles(rootPath: string, files: string[]): Promise<string[]> {
+        // Pattern to find .gitignore files with either windows or unix path styles
+        const gitIgnorePattern = /.*[\/\\]\.gitignore$/
```
ege0zcan, line 126: although not a huge risk, this regex seems to be of polynomial complexity.

I didn't test this but would something like `(?:^|/)\.gitignore(?=$|/)` work?

server/aws-lsp-codewhisperer/src/language-server/pathUtils.ts
```diff
@@ -0,0 +1,51 @@
+import os = require('os')
+
+/**
+ * Returns `true` if path `p` is a descendant of directory `d` (or if they are
```
ege0zcan, line 4: ideally as a followup it would be nice to have some unit tests for these functions

ege0zcan (approved):

PR #363, opened 2024-07-01
Add Semantic Tokens support for aws-lsp-partiql server

XuechunHHH:
## Problem
The current aws-lsp-partiql server does not support Semantic Tokens.

## Solution
This update introduces Semantic Tokens support for the aws-lsp-partiql server, enabling syntax highlighting for PartiQL files. The implementation leverages a fallible tree-sitter parser developed by the PartiQL Team.

The parser is compiled into a WebAssembly binary, which is then encoded as a base64 string within a TypeScript file for compatibility with the `web-tree-sitter` package. Additionally, the `tree-sitter.wasm`file from the `web-tree-sitter` package is similarly compiled into a base64 string and included in a TypeScript file. This setup enables differentiated instantiation in browser and Node.js environments, allowing the server to be used more efficiently as a web worker when assets are loaded from a CDN, compared to direct wasm file bundling.

- Testing was conducted on the function wrapped in a standalone runtime in VS Code

## Todo
- Test the server when wrapped in the web worker runtime with a web-based LSP client.
- Resolve the issue where the PartiQL Server could not be launched in the standalone runtime in VS Code following PR#328.
- Sync this branch with the current main branch of `aws/language-servers` after resolving the above issue, as this branch is currently several commits behind.

## License


server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
```diff
@@ -0,0 +1,195 @@
+import Parser from 'web-tree-sitter'
```
akoreman, line 1: I think you forgot to add `web-tree-sitter` as a dependency to the package.json of this server.

XuechunHHH: I believe `web-tree-sitter` package has already be centrally managed over [here](https://github.com/aws/language-servers/blob/main/package-lock.json#L16074).

XuechunHHH: Thanks for catching that! Added it.

server/aws-lsp-partiql/package.json
```diff
@@ -24,10 +26,10 @@
         "@aws/language-server-runtimes": "^0.2.4"
```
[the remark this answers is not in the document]
XuechunHHH: Updated it.

server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
```diff
@@ -0,0 +1,195 @@
...
+    bag_punc_start: SemanticTokenTypes.operator,
+    bag_punc_end: SemanticTokenTypes.operator,
+    bag_punc_separator: SemanticTokenTypes.operator,
+    ion: SemanticTokenTypes.string,
+}
+
+// Initialize and prepare the parser
+async function initParser() {
+    let wasmModuleUrl: string | null = null
+    if (typeof window !== 'undefined' && typeof window.URL.createObjectURL === 'function') {
```
akoreman, line 40: We're missing the case here where the server is being run inside a [Web worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers). For web workers it could look something like

```typescript
  else if (typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope) {
      // Webworker
      const wasmBlob = new Blob([tree_sitter_inline_1.default], { type: 'application/wasm' });
      wasmModuleUrl = URL.createObjectURL(wasmBlob);
  }
```

akoreman:
I tested the branch in a web-based LSP client and if we make the changes to support the webworker environment everything seems to work as expected 🚀

server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
```diff
@@ -0,0 +1,195 @@
...
+    const PartiQL = await Parser.Language.load(parserBase64)
+    parser.setLanguage(PartiQL)
+    return parser
+}
+
+export async function findNodes(
+    sourceCode: string | Parser.Input,
+    nodeType: SemanticTokenTypes | string
+): Promise<SemanticToken[]> {
+    const parser = await initParser()
```
akoreman, line 72: We now initialize the parser every time the client requests semantic tokens, that's probably a bit inefficient. Could we run `initParser` only once when the server/service gets created and then reuse it?

server/aws-lsp-partiql/src/server/language-service.ts
```diff
@@ -35,4 +60,19 @@ class PartiQLLanguageService {
         }
         return diagnostics
     }
+
+    public async doSemanticTokens(textDocument: TextDocument): Promise<SemanticTokens | null> {
+        const text = textDocument.getText()
+        const data: SemanticToken[] = []
+        for (const nodeType of semanticTokensLegend.tokenTypes) {
```
akoreman, line 67: Mainly just a question, why do we have these 2 different things to loop over (`semanticTokensLegend.tokenTypes` and `string2TokenTypes`). What are the differences between those two?

akoreman:
We only include the compiled version of the tree sitter rules (`tree-sitter-partiql.wasm`), can we also include the files required and some steps or a script to update/expand those tree sitter rules when required?

akoreman:
Could you also add some automated tests for the semantic tokens to `language-server.test.ts` and `language-service.test.ts` to verify that strings of source code get tokenized as expected?

server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
XuechunHHH: Thanks for pointing out. Added the Webworker Case in a latest commit. To avoid changing the code in `tsconfig.json`, use a type assertion instead here:
``` typescript
else if (typeof self !== 'undefined' && self.constructor.name === "DedicatedWorkerGlobalScope") {
    // Webworker
    const wasmBlob = new Blob([wasmBinaryArray], { type: 'application/wasm' })
    wasmModuleUrl = URL.createObjectURL(wasmBlob)
} 
```

server/aws-lsp-partiql/src/server/language-service.ts
XuechunHHH: Both `semanticTokensLegend.tokenTypes` and `string2TokenTypes` are node type defined in tree-sitter parser, where `semanticTokensLegend.tokenTypes` are the same type defined in LSP and `string2TokenTypes` are manually mapped to some type in LSP. 
For example, the `,` in `a, b` has type `operator` in tree-sitter and `,` in `{'a':1, 'b':1}` has type `tuple_punc_separator`. And the later one will be manually mapped to `SemanticTokenTypes.operator` defined in LSP.
I will combine the `semanticTokensLegend.tokenTypes` into `string2TokenTypes` if the code looks cleaner to you in that case. For now, I just keep them in two different lists to avoid overlapped definition and repeated code. Also, when adding a new `SemanticTokenTypes` support later, manual mapping will only be needed as a complementary.

XuechunHHH:
> We only include the compiled version of the tree sitter rules (`tree-sitter-partiql.wasm`), can we also include the files required and some steps or a script to update/expand those tree sitter rules when required?

That's an interesting proposal. We will definitely do this after open source the `tree-sitter-partiql` parser by adding a script to pull and compile the latest version of the parser everytime pre-compile the server. Fow now, we will just include the compiled version of parser.

server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
XuechunHHH: You are right. Current implementation is quite inefficient and created a global parser for reusage in a latest commit.

XuechunHHH:
> Could you also add some automated tests for the semantic tokens to `language-server.test.ts` and `language-service.test.ts` to verify that strings of source code get tokenized as expected?

Thank you for your suggestion. The tokenization behavior has already been tested within the development of `tree-sitter-partiql`. I will add some automated tests for the semantic tokens to `language-server.test.ts` and `language-service.test.ts` to verify the functionality of the `onSemanticTokens` request handler and the accurate identification of `SemanticTokenTypes`.

PR #377, opened 2024-07-11
Add Semantic Tokens Support for PartiQL Server

XuechunHHH:
## Problem
The current aws-lsp-partiql server does not support Semantic Tokens.
## Solution
This update introduces Semantic Tokens support for the aws-lsp-partiql server, enabling syntax highlighting for PartiQL files. The implementation leverages a fallible tree-sitter parser developed by the PartiQL Team.

The parser is compiled into a WebAssembly binary, which is then encoded as a base64 string within a TypeScript file for compatibility with the `web-tree-sitter` package. Additionally, the `tree-sitter.wasm` file from the `web-tree-sitter package` is similarly compiled into a base64 string and included in a TypeScript file. This setup enables differentiated instantiation in browser and Node.js environments, allowing the server to be used more efficiently as a web worker when assets are loaded from a CDN, compared to direct wasm file bundling.

- Tested the server wrapped in the webworker runtime a web-based LSP client.
- Added automated unit tests for  both the server and the service.

## License


server/aws-lsp-partiql/package.json
```diff
@@ -13,15 +13,19 @@
     "scripts": {
         "compile": "tsc --build",
         "test": "jest",
+        "precompile": "npm run update-treesitter-wasm",
```
akoreman, line 16: The `precompile` script is run every time `compile` is run, which can be quite often during development. Do we need to update the treesitter wasm every time wel compile? I think it only needs to be updated sometimes right?

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-parser-wasm.mts
```diff
@@ -0,0 +1,13 @@
+import { readFileSync, writeFileSync } from 'fs'
+
+// Read .wasm binary file, NOTICE and LICENSE file and add description of our modifications.
```
akoreman, line 3: Nitpick, but this comment isn't fully accurate, we're not reading the NOTICE and LICENSE files

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-parser-wasm.mts
```diff
@@ -0,0 +1,13 @@
...
+const base64EncodedBinary = readFileSync('./src/server/tree-sitter-parser/tree-sitter-partiql.wasm', 'base64')
+const modifications =
+    '`base64EncodedBinary` contains a base64 encoding of the tree-sitter-partiql parser compiled to WebAssembly.\n\tModifications around it are to import it from a JS/TS file and instantiate it using `WebAssembly.instantiate.`'
+
+// Copy the encoded binary, notice and license into a .ts file.
```
akoreman, line 8: Same comment as above

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-parser-wasm.mts
```diff
@@ -0,0 +1,13 @@
...
+const encodedPartiQLServer = `/*\n${modifications}\n*/\n\nconst base64String = atob("${base64EncodedBinary}");\nconst base64Buffer = new Uint8Array(base64String.length);\nfor (var i = 0; i < base64String.length; i++) base64Buffer[i] = base64String.charCodeAt(i);\nexport default base64Buffer;`
+writeFileSync('./src/server/tree-sitter-parser/tree-sitter-parser-inline.ts', encodedPartiQLServer)
+
+// eslint-disable-next-line no-undef
```
akoreman, line 12: Don't think we need this eslint disable here?

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-wasm.mts
```diff
@@ -0,0 +1,15 @@
+import { readFileSync, writeFileSync } from 'fs'
+import { fileURLToPath } from 'url'
+
+// Using import.meta.resolve to dynamically resolve the path to the wasm file
+const wasmModuleUrl = await import.meta.resolve('web-tree-sitter/tree-sitter.wasm')
+const filePath = fileURLToPath(wasmModuleUrl)
+
+// Read .wasm binary file, NOTICE and LICENSE file and add description of our modifications.
```
akoreman, line 8: Same comment as on the file above

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-wasm.mts
```diff
@@ -0,0 +1,15 @@
...
+const base64EncodedBinary = readFileSync(filePath, 'base64')
+const modifications =
+    '`base64EncodedBinary` contains a base64 encoding of the web-tree-sitter compiled to WebAssembly.\n\tModifications around it are to import it from a JS/TS file and instantiate it using `WebAssembly.instantiate.`'
+
+// Copy the encoded binary, notice and license into a .ts file.
```
akoreman, line 13: Same comment as on the file above

server/aws-lsp-partiql/package.json
XuechunHHH: Yea, this script is only needed to and has to run whenever the package version `web-tree-sitter` upgrades. To make sure the script runs when needed, I will force the `web-tree-sitter` version to be "0.22.6" for now and leave an update instruction message in README.md.

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-parser-wasm.mts
XuechunHHH: Thanks for pointing out. Update it.

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-parser-wasm.mts
XuechunHHH: Update.

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-wasm.mts
XuechunHHH: Update.

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-wasm.mts
XuechunHHH: Update.

server/aws-lsp-partiql/src/server/tree-sitter-parser/parserUtil.ts
```diff
@@ -0,0 +1,34 @@
...
+
+// Initialize and prepare the parser
+export async function initParser() {
+    let wasmModuleUrl: string | null = null
+    if (typeof window !== 'undefined' && typeof window.URL.createObjectURL === 'function') {
+        // Browser
+        const wasmBlob = new Blob([wasmBinaryArray], { type: 'application/wasm' })
+        wasmModuleUrl = URL.createObjectURL(wasmBlob)
+    } else if (typeof process !== 'undefined' && typeof process.versions.node !== 'undefined') {
+        // Node.js
```
akoreman, line 13: Doing some thinking, as far as I can tell, the method using the blobs we use for browser and web worker should also work in Node. Why wouldn't we just use the same method for all paths to reduce these somewhat complex branching paths here?

It might lead to a slightly slower initialisation on node, but I think that's negligible and happens only once per server lifecycle.

server/aws-lsp-partiql/build-scripts/compile-inline-tree-sitter-parser-wasm.mts
XuechunHHH: We use `require()` for `fs` and `path` here because they’re specific to Node.js, where `import` isn't as compatible. Though, the code compiles and runs correctly without this eslint disable here. I've disabled the ESLint rule `@typescript-eslint/no-var-requires` here to avoid errors reported.

server/aws-lsp-partiql/src/server/tree-sitter-parser/parserUtil.ts
XuechunHHH: I understand the appeal of a unified approach using `Blob` across environments. However, while Node.js supports `Blob` objects, the `URL.createObjectURL()` function, which is primarily used in browser environments to create URLs pointing to in-memory files, is not natively supported in Node.js. 

I have attempted to use this method in Node.js result in errors because it cannot properly interpret or use the blob URLs created. That's why I've implemented a different handling for the Node.js environment.

image

akoreman: Hmm weird, `URL.createObjectURL` does [exist](https://nodejs.org/api/url.html#urlcreateobjecturlblob) in Node >=16.7 but it indeed doesn't seem to work, not straight away at least.

server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
```diff
@@ -0,0 +1,210 @@
...
+
+async function ensureParserInitialized() {
+    if (!parserInitialized) {
+        globalParser = await initParser()
+        parserInitialized = true
+    }
+}
+
+// Initialize and prepare the parser
+async function initParser() {
```
akoreman, line 49: This code exists twice now right? here and in parserUtil.ts, or is there a difference? (I do think having this scaffolding code in a separate util file is a good idea)

server/aws-lsp-partiql/src/server/tree-sitter-parser/parserUtil.ts
akoreman: This should work across both Node, browser and webworker I think

```
// Initialize and prepare the parser
async function initParser() {
    await Parser.init({
        instantiateWasm(imports: any, succesCallback: any) {
            WebAssembly.instantiate(wasmBinaryArray, imports).then((arg) => succesCallback(arg.instance, arg.module));
        }
    })
    const parser = new Parser()
    const PartiQL = await Parser.Language.load(parserBase64)
    parser.setLanguage(PartiQL)
    return parser
}
```

I tested it works in Node, I'll also verify in webworker/browser.

akoreman: Also works in worker/browser

server/aws-lsp-partiql/tsconfig.json
```diff
@@ -3,7 +3,8 @@
     "compilerOptions": {
         "rootDir": "./src",
         "outDir": "./out",
-        "allowJs": true
+        "allowJs": true,
+        "esModuleInterop": true
```
akoreman, line 7: What do we need this for?

server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
XuechunHHH: Yeah, I placed the initialization in a separate utility because hover help and signature help will also use it later. It was an oversight on my part not to sync the code in `parser-tokens.ts` during the modification. Thanks for pointing it out!

server/aws-lsp-partiql/tsconfig.json
XuechunHHH: The module '"web-tree-sitter"' can only be default-imported using the 'esModuleInterop' flag. So added this setting line here.

akoreman (approved):

XuechunHHH:
FYI, I have included a link #363  to the closed draft PR below for historical context to help new viewers trace the development and discussion back to its origins.

server/aws-lsp-partiql/README.md
```diff
@@ -10,6 +10,8 @@ _Updating the binary requires the Docker CLI to be in your PATH_
 
 To update the binary used by the server, run `npm run update-parser-binary`. This will, inside a Docker container, pull the PartiQL playground repository from GitHub and run the `build` command to create a new binary. It is then encoded into base64 in a Typescript file and copied to the `src` folder. Running `npm run build` will then transpile the binary file to Javascript and place it in the `lib` folder.
 
+To update the binary used in package `web-tree-sitter`, run `npm run update-treesitter-wasm`. This update is forced to make everytime upgrading the version of `web-tree-sitter`. This will compile the `tree-sitter.wasm` file inside the package into a base64 string in a TypeScript file to the `src/tree-sitter-parser` folder as `tree-sitter-inline.ts`.
```
volodkevych, line 13: Have you considered excluding generated files from `prettier`, `eslint`?

server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
```diff
@@ -0,0 +1,174 @@
...
+            breakDownTokens.push(lastToken)
+        }
+    }
+    return breakDownTokens
+}
+
+export function encodeSemanticTokens(
+    tokens: SemanticToken[],
+    multilineTokenSupport: boolean
+): Promise<SemanticTokens | null> {
```
volodkevych, line 152: Why do you need it to return promise here?

volodkevych (approved):

server/aws-lsp-partiql/README.md
XuechunHHH: Yes, that's a great suggestion to keep our repository cleaner. I've added the generated files to the `.prettierignore` and `.eslintignore` configurations to prevent them from being unnecessarily reformatted or linted. Thank you for pointing this out!

server/aws-lsp-partiql/src/server/syntax-highlighting/parser-tokens.ts
XuechunHHH: Thanks for catching that! It was indeed a legacy error from earlier code modifications. I've removed the promise as suggested.

server/aws-lsp-partiql/src/server/tree-sitter-parser/parserUtil.ts
```diff
@@ -0,0 +1,16 @@
+import Parser from 'web-tree-sitter'
+import parserBase64 from '../tree-sitter-parser/tree-sitter-parser-inline'
+import wasmBinaryArray from '../tree-sitter-parser/tree-sitter-inline'
+
+// Initialize and prepare the parser
+export async function initParser() {
+    await Parser.init({
+        instantiateWasm(imports: any, succesCallback: any) {
```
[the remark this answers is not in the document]
XuechunHHH: Thanks for the heads up! I will add some comments here.

PR #396, opened 2024-07-23
Add Hover and SignatureHelp Support for PartiQL Server

XuechunHHH:
## Problem
The current aws-lsp-partiql server does not support Hover Help for Keywords and Signature Help for functions.
## Solution

This update introduces Keywords Hover and Functions Signature Help support for the `aws-lsp-partiql` server, enabling further language support for PartiQL files. 

After the change, the server now can provide description and doc link reference for keywords defined in PartiQL and signature help with examples for built-in functions defined in SQL:1992.

- Added automated unit tests for both the server and the service.
## License


server/aws-lsp-partiql/src/server/hover-info/hoverDictionary.ts
```diff
@@ -0,0 +1,96 @@
+export const hoverDictionary = {
+    keyword: {
+        as: [
+            '### AS',
+            'The AS keyword is optionally used when renaming a table or column. Aliases provide users with the ability to reduce the amount of code required for a query.',
+            '',
+            'Visit [PartiQL Documentation](https://partiql.org/dql/as.html)',
```
akoreman, line 7: PartiQL is used by a bunch of AWS consoles which all have their own documentation pages on query syntax, they might not be happy to have links to the PartiQL website inside their console pages (instead of to AWS docs). It might be better to either not have links, or add a parameter to the server creation to allow users to hide the link.

server/aws-lsp-partiql/src/server/hover-info/parser-type.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Hover, MarkupKind } from '@aws/language-server-runtimes/server-interface'
+import { hoverDictionary } from './hoverDictionary'
+
+// Global or service-level variables
+let parserInitialized = false
+let globalParser: Parser | null = null
+
+async function ensureParserInitialized() {
+    if (!parserInitialized) {
+        globalParser = await initParser()
```
akoreman, line 12: This would create a new parser instance separate from the one we use for the semantic tokens right? Do you see a way we could use the single instance in both places?

server/aws-lsp-partiql/src/server/language-server.test.ts
```diff
@@ -122,3 +123,72 @@ describe('PartiQL Language Service - Semantic Tokens', () => {
         jest.restoreAllMocks()
     })
 })
+
+describe('PartiQL Server - Hover Functionality', () => {
```
akoreman, line 127: Should we also add tests for the signature functionality for this file?

server/aws-lsp-partiql/src/server/hover-info/hoverDictionary.ts
```diff
@@ -0,0 +1,96 @@
+export const hoverDictionary = {
+    keyword: {
+        as: [
+            '### AS',
```
akoreman, line 4: For these short hover snippets, not sure we need a header. The user already knows which token they're hovering on so it might be a bit cluttering to also have the token name in the tooltip? 🤔

server/aws-lsp-partiql/src/server/language-server.test.ts
XuechunHHH: Definitely! Will add the test for signature request handler before turning it ready to review.

server/aws-lsp-partiql/src/server/hover-info/parser-type.ts
XuechunHHH: You are right. There is no need to use separate parsers. Will put the globalParser into `parserUtil.ts`.

server/aws-lsp-partiql/src/server/hover-info/hoverDictionary.ts
XuechunHHH: Removed the `url`.

server/aws-lsp-partiql/src/server/hover-info/hoverDictionary.ts
XuechunHHH: Change to a better looking as follows:

server/aws-lsp-partiql/src/server/hover-info/parser-type.ts
```diff
@@ -0,0 +1,35 @@
...
+    position: { line: number; character: number }
+): Promise<Hover | null> {
+    const node = await detectNodes(sourceCode, position)
+    const nodeType = node.type.toLowerCase() as keyof typeof hoverDictionary
+    const nodeText = node.text.toLowerCase() as keyof (typeof hoverDictionary)[keyof typeof hoverDictionary]
+    if (nodeType in hoverDictionary && nodeText in hoverDictionary[nodeType]) {
+        const hoverInfo = hoverDictionary[nodeType][nodeText]
+        return {
+            contents: {
+                kind: MarkupKind.Markdown,
```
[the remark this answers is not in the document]
XuechunHHH: Thanks for providing the context. Added `supportHoverMarkdown` toggle support.

akoreman (approved):

viktorsaws (approved):

PR #398, opened 2024-07-25
Add triggerEmptySuggestionsRequest command to CodeWhisperer server

andredcoliveira:
## Problem

CodeWhisperer API calls can fail depending on what permissions are associated with the credentials that are being used by the server/service.

It is useful for a client to have available a mechanism that behaves like a dry-run—in this case, an API call is performed using parameters that either are empty or produce no connection to the end-user.

For example, a destination using the server may wish to render different UIs based on whether the customer has permissions or not without that check implicating the customer in any relevant way—specifically, without having the customer data (e.g., file context) cross the AWS region boundary.

## Solution

Having the following considerations in mind:
- the server always has a copy of the customer's file contents due to file synchronization
- the destination has no current way to tell the server not to include these file contents in an API call
- this can be applicable to some language servers but not necessarily to the majority of them, and they're likely to have different nuances not simple to parameterise, so it's probably best to not make it a part of the protocol

The proposal in this pull request is to introduce a command that destinations can use to request the server to perform a request with no customer data. Destinations can then listen to telemetry events to detect permissions errors.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
@@ -556,6 +556,35 @@ export const CodewhispererServerFactory =
             lastUserModificationTime = new Date().getTime()
         })
 
+        const TriggerEmptySuggestionsRequestCommand = 'aws/codewhisperer/triggerEmptySuggestionsRequest'
+
+        lsp.onExecuteCommand(async (params, _token) => {
+            logging.log(`Trying to execute command: ${params.command}`)
+            if (params.command === TriggerEmptySuggestionsRequestCommand) {
+                return await codeWhispererService.generateSuggestions({
+                    fileContext: {
+                        leftFileContent: '',
+                        rightFileContent: '',
+                        filename: '',
+                        programmingLanguage: { languageName: 'typescript' },
+                    },
+                    maxResults: 1,
+                })
+            } else {
+                logging.log(`Unknown command: ${params.command}`)
+            }
+        })
+
+        lsp.addInitializer(() => {
+            return {
+                capabilities: {
+                    executeCommandProvider: {
+                        commands: [TriggerEmptySuggestionsRequestCommand],
+                    },
+                },
+            }
+        })
+
         logging.log('Amazon Q Inline Suggestion server has been initialised')
 
         return () => {
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
@@ -558,18 +558,37 @@ export const CodewhispererServerFactory =
 
         const TriggerEmptySuggestionsRequestCommand = 'aws/codewhisperer/triggerEmptySuggestionsRequest'
 
-        lsp.onExecuteCommand(async (params, _token) => {
+        lsp.onExecuteCommand((params, _token) => {
             logging.log(`Trying to execute command: ${params.command}`)
             if (params.command === TriggerEmptySuggestionsRequestCommand) {
-                return await codeWhispererService.generateSuggestions({
-                    fileContext: {
-                        leftFileContent: '',
-                        rightFileContent: '',
-                        filename: '',
-                        programmingLanguage: { languageName: 'typescript' },
+                const fileContext = {
+                    leftFileContent: '',
+                    rightFileContent: '',
+                    filename: 'x', // service definition requires a minimum length of 1
+                    programmingLanguage: { languageName: 'typescript' },
+                }
+                const maxResults = 1
+                // We don't use the session manager so that we don't interfere with ongoing sessions
+                const dummySession = new CodeWhispererSession({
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -556,6 +556,54 @@ export const CodewhispererServerFactory =
...
+                        maxResults,
+                    },
+                })
+                codeWhispererService
+                    .generateSuggestions({
+                        fileContext,
+                        maxResults,
+                    })
+                    .then(() => {
+                        emitServiceInvocationTelemetry(telemetry, dummySession)
```
viktorsaws, line 587: I don't think that using telemetry adding more implicit meaning to it is correct way to archive this. 
LSP's `workspace/executeCommand` supports sending responses from the server, you can use it directly. We return in securityScan, for example https://github.com/aws/language-servers/blob/31d001f8713adc1edcc328f4da9db4b914fd4d77/server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts#L205

andredcoliveira: I considered that and still chose this way for the following reasons:

- clients can detect permissions issues in the same way for this command as they currently do for normal inline completion requests
   - significantly reduces the effort on the client since it allows reusing logic at no particular effort cost to the server
- the server does not currently return responses of any kind for errors/failures: https://github.com/aws/language-servers/blob/31d001f8713adc1edcc328f4da9db4b914fd4d77/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts#L454-L463
   - I think exposing details of the response via the command would require something like one of the following:
      1. exposing more data than desired by returning the response transparently
      2. modifying the command to be permissions-specific and having specific parsing logic for the response that might only be applicable to IAM/SigV4 (i.e., detect permissions issues directly in the server)
         - this is what I started with, but I was asked by other maintainers (outside of this PR) to not implement IAM-specific logic
         - this would mean more logic to maintain in somewhat in line with what `emitServiceInvocationFailure` already does
- for all intents and purposes, this is indeed a "true" service invocation that reaches the backend, so I don't see why service invocation telemetry should not be emitted for it
   - It's just not triggerered on-demand by the client, and is not triggered via an inline completion request, which is why I created the session directly instead of using the session manager. This way, for all intents and purposes of this server's runtime, it's similar to, e.g., creating a separate server and/or service for the sole purpose of checking permissions.

What implicit meaning do you see being added to the telemetry, and why do you think it's not correct?

viktorsaws: > clients can detect permissions issues in the same way for this command as they currently do for normal inline completion requests
> ...
> the server does not currently return responses of any kind for errors/failures

I think this is main issue and the missing feature of server that is needed. We need to process failed requests and return meaningful error messages, instead of relying on parsing telemetry and logs from the server. This is a miss from initial implementation.

IMO, introducing separate command complicates integration and is unnecessary to solve this problem. Would it make sense to add error handling instead of this command for your use case?

andredcoliveira: To clarify, the goal of this command is not to solve the error handling issue. It's instead to allow detecting issues (e.g., permissions) before the end-user actually needs the feature and using a request that is guaranteed to not carry customer data across the AWS region border (this is the most important part).

Even if error handling is added, the server would automatically add customer data that was already shared via the LSP synchronization feature to the request's file context.

The only other proper way I see of doing this is for the client to open a virtual "hidden" file that is empty and has a dummy filename, but that's a bit contrived and doesn't make use of the LSP protocol as much, so I'm not a big fan of it—but I can be biased here since I'm working on the client-side, so I'm curious to hear your thoughts.

A third solution I see is to parameterise the inline completion request type to include a dry-run parameter of sorts, but that would require modifying the interface for all servers that integrate with https://github.com/aws/language-server-runtimes

andredcoliveira: And I guess a fourth solution would be to have something like a "detect permissions" request, but:
- permissions are request-specific, since you may have permissions for an API used in the server and not for another
- not all language servers have the notion of requiring "permissions"

viktorsaws: Separate dedicated Q API to verify permissions looks like the most appropriate way to implement this imo. Hacking this UX with custom commands makes server API confusing and more difficult to maintain

volodkevych: I'm missing context here: why are we programming for this specific case when user (accidentally?) does not have permissions? Shouldn't client know whether it has permissions or not?

volodkevych: @viktorsaws Could it be also integrated into update credentials request to return error if some server is not supported? As an optional check, for example.

PR #400, opened 2024-07-26
Ensure that the entire chat request is cancellable

makenneth:
## Description

The main goal of this PR is to make entire chat request cancellable. Currently, if user cancels before the response streaming starts, the request will still go through. For the fqn piece, we can use Workerpool's "cancel" feature. Otherwise, we just need to early terminate and return a `ResponseError` with `RequestCancelled` code.

I tried to make this reusable by creating a wrapper function that creates a promise for cancellation that will "race" against the core logic: the value that is resolved first will be returned. 

The other two changes:
- Combined `DocumentContext` and `TriggerContext` into a single `TriggerContextExtractor` as the separation offer no benefits
- Removed unneeded start trigger information

## Testing
I tested in the vscode test extension. It still needs to be verified

## License


makenneth:
Moving to https://github.com/aws/language-servers/pull/432 instead which uses a different branch origin

PR #406, opened 2024-07-29
Add completion proposal support for PartiQL Server 

XuechunHHH:
## Problem
The current `aws-lsp-partiql` server does not support completion proposal.

## Solution
This update introduces Completion Proposal support for the `aws-lsp-partiql` server, enabling further language support for PartiQL files.

After the change, the server use [ANTLR](https://github.com/partiql/partiql-lang-kotlin/blob/d660231f2e59ceae0a8306ce880491d8f710708b/partiql-parser/src/main/antlr/PartiQLParser.g4) grammar defined by PartiQL team to support completion proposal. The grammar in this server is slightly changed to support multi-quries.

https://github.com/user-attachments/assets/49a4856c-f597-4403-8a37-51d1a41668a5

- Added automated unit tests for both the server and the service.

## License


server/aws-lsp-partiql/src/server/completion-hint/parser-completion.ts
```diff
@@ -0,0 +1,153 @@
...
+    parser.removeErrorListeners()
+    return parser
+}
+
+function symbolAtCaretPosition(parseTree: TerminalNode, caretPosition: { line: number; character: number }) {
+    const start = parseTree.symbol.column
+    const stop = start + (parseTree.symbol.text?.length ?? 0)
+
+    return (
+        parseTree.symbol.line == caretPosition.line + 1 &&
```
akoreman, line 56: I think this `+ 1` here is to account for ANTLR lines being indexed from 1? It might make it a bit clearer to do the translation from LSP coords -> ANTLR coords a bit earlier, maybe as part of the `getPosition` method?

XuechunHHH: It is a smart way to clean up the code. Appreciate the suggestion!

akoreman (approved):

imykhai (approved):

PR #416, opened 2024-08-13
Fixing issue with random source framework selection

pranav-firake:
## Problem
When multiple projects/ solution is selected with multiple projects, source framework for project is randomly selected. This causes .net8 projects to be selected to transform and backend will reject this request because that project is already transformed

## Solution
Added logic to get minimum version of dot net and set it as source framework

## Significance of source framework here
only for cwspr API

- [ ✔] Pull request title describes the changes without ambiguity
- [ ✔] Added screenshots of visual/UI changes 
- [ ✔] Added Unit test coverage
- [ ✔] Removed all developer references (internal URLs, etc)
- [ ✔] Removed unused code, comments, references, namespaces
- [ ✔] Used constants/readonly where needed
- [ ✔] Used appropriate data structures
- [ ✔] Used consistent naming convention
- [ ✔] Used predefined Toolkit themes for UI elements
- [ ✔] Used appropriate method and property access levels (private vs public)
- [ ✔] Avoided global and singleton references where possible
- [ ✔] Enforced DRY principles (Do not Repeat Yourself)
- [ ✔] All TODOs have an internal tracking reference ID (e.g. SIM ID)
- [ ✔] Resolved merge conflicts with destination branch
- [ ] Reviewed by Huawen
- [x] Optional: Reviewed by Pranav, Sunwoo, Ke, Jiayu

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
@@ -39,7 +39,7 @@ function getKeyIndexOfVersion(key: any) {
     return targetFrameworkKeysArray.indexOf(key)
 }
 
-function findMinimumSourceVersion(projectMetadata: TransformProjectMetadata[], logging: Logging) {
+export function findMinimumSourceVersion(projectMetadata: TransformProjectMetadata[], logging: Logging) {
     var minimumVersionIndex = dummyVersionIndex
     projectMetadata.forEach(project => {
         if (project.ProjectTargetFramework != '' && targetFrameworkMap.has(project.ProjectTargetFramework)) {
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
@@ -31,25 +31,31 @@ export const targetFrameworkMap = new Map<string, string>([
     ['net6.0', 'NET_6_0'],
     ['net7.0', 'NET_7_0'],
     ['net8.0', 'NET_8_0'],
-  ]);
-const dummyVersionIndex = 999;
+])
+const dummyVersionIndex = 999
 
-const targetFrameworkKeysArray = Array.from(targetFrameworkMap.keys());
-function getKeyIndexOfVersion( key: any) {
-  return targetFrameworkKeysArray.indexOf(key);
+const targetFrameworkKeysArray = Array.from(targetFrameworkMap.keys())
+function getKeyIndexOfVersion(key: any) {
+    return targetFrameworkKeysArray.indexOf(key)
 }
 
-function findMinimumSourceVersion(projectMetadata : TransformProjectMetadata[], logging: Logging){
-    var minimumVersionIndex = dummyVersionIndex;
+function findMinimumSourceVersion(projectMetadata: TransformProjectMetadata[], logging: Logging) {
+    var minimumVersionIndex = dummyVersionIndex
     projectMetadata.forEach(project => {
-        if(project.ProjectTargetFramework != '' && targetFrameworkMap.has(project.ProjectTargetFramework)){
-            logging.log("Project version to compare " + project.ProjectTargetFramework);
-                minimumVersionIndex = getKeyIndexOfVersion(project.ProjectTargetFramework) < minimumVersionIndex ?
-                 getKeyIndexOfVersion(project.ProjectTargetFramework) : minimumVersionIndex;
-        }});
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
```diff
@@ -1,47 +1,63 @@
...
+        if(project.ProjectTargetFramework != ''){
+            if(targetFrameworkMap.has(project.ProjectTargetFramework)){
+                logging.log("Project version here " + project.ProjectTargetFramework);
+                selectedVersionIndex = getKeyIndexOfVersion(project.ProjectTargetFramework) < selectedVersionIndex ? getKeyIndexOfVersion(project.ProjectTargetFramework) : selectedVersionIndex;
+            }
+        }});
+    var selectedDotNetVersion = selectedVersionIndex != 999? targetFrameworkMap.get(targetFrameworkKeysArray[selectedVersionIndex]) : '';
+    logging.log("Selected version " + selectedDotNetVersion);
+    return selectedDotNetVersion;
+}
```
jonlouie, line 52: Would recommend restructuring this function based on the following to make it more readable and less error prone:
1. Rename the function to more accurately describe what it is doing (e.g. `convertTargetFrameworkForTransformRequest`)
2. Avoid writing literal values in multiple places (e.g. store the value 999 in a const like `DEFAULT_FRAMEWORK_INDEX`). We may not even need this change though (see below)
3. In the forEach loop, `break` once you've found a match.
4. Use simpler logical flow like:

```typescript
var convertedTargetFramework = '';
if(project.ProjectTargetFramework != ''){
    projectMetadata.forEach(project => {
        convertedTargetFramework = ... ;
    }
}
logging.log("For project {projectName}, converted target framework from \"{initialFramework}\" to \"{convertedTargetFramework}\"");

return convertedTargetFramework;
```

server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
```diff
@@ -1,47 +1,63 @@
...
 ): CodeWhispererTokenUserClient.StartTransformationRequest {
-    const targetProject = userInputRequest.ProjectMetadata.find(project => project.ProjectTargetFramework != '')
-    const sourceFramework =
-        targetProject == undefined
-            ? ''
-            : targetFrameworkMap.has(targetProject.ProjectTargetFramework)
-              ? targetFrameworkMap.get(targetProject.ProjectTargetFramework)
-              : ''
+    const sourceFramework = findMinimumSourceVersion(userInputRequest.ProjectMetadata, logging);
+    logging.log("Selected sourceFramework for startTransform Request" + sourceFramework);
```
jonlouie, line 60: Add a space after `"startTransform Request"`. Also, for string concatenation, I would recommend string interpolation for better customization. It's also a more visual approach to string building that would help catch this kind of thing.

server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
pranav-firake: hi @jonlouie , I did not get why we do not need the above check, from list, we need to find 1 minimum source framework. There can be multiple projects in metadata like .net48, net6, net8. What we need to tag is net48 for gumby request. 
please lmk if I am missing something

xueningl-az (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
cslong: This is a nit, but would it be easier to read the loop if combined the conditions instead of having them be nested?

cslong (approved):
left a small nitpick/question, but otherwise lgtm

server/aws-lsp-codewhisperer/src/language-server/netTransform/converter.ts
jonlouie: @pranav-firake Thanks for the offline explanation. Per our conversation, we've agreed on the following changes:
1. Store 999 in a local `const`
2. Rename variables from `selected...` to `minimum...`

We did not discuss this offline, but I would suggest updating the logging statement to something like `"Lowest target framework found: {minimumDotNetVersion}"`

pranav-firake: thanks Jon,

pranav-firake: yes Chris, agreed, refactored to use single condition check rather than nested

jonlouie (approved):

ege0zcan (approved):

PR #422, opened 2024-08-15
Sanitize file paths sent in chat requests

makenneth:
## Description
This is a low app sec concern that full paths can include sensitive information.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.ts
@@ -75,10 +75,14 @@ export class DocumentContextExtractor {
             cursorState: rangeWithinCodeBlock ? { range: rangeWithinCodeBlock } : undefined,
             text: document.getText(codeBlockRange),
             programmingLanguage: languageId ? { languageName: languageId } : undefined,
-            relativeFilePath: document.uri,
+            relativeFilePath: this.#getSantizedFilePath(document),
             documentSymbols,
             hasCodeSnippet: Boolean(rangeWithinCodeBlock),
             totalEditorCharacters: document.getText().length,
         }
+    }
+
+    #getSantizedFilePath(document: TextDocument) {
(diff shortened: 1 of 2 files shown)


ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.ts
```diff
@@ -75,10 +75,14 @@ export class DocumentContextExtractor {
...
-            relativeFilePath: document.uri,
+            relativeFilePath: this.#getSantizedFilePath(document),
             documentSymbols,
             hasCodeSnippet: Boolean(rangeWithinCodeBlock),
             totalEditorCharacters: document.getText().length,
         }
     }
+
+    #getSantizedFilePath(document: TextDocument) {
+        return document.uri.split('/').pop() ?? ''
```
viktorsaws, line 86: This does not look correct. Using .pop() here you always return only the last section of the document URI, which will be only filename (as tests above also confirm). Looking at logs for Q for VSCode extension, full path is `relativeFilePath` context field is expected.

Since in real project scenario, documents are located in deeply nested directories, in which case this function resolves to wrong path:

> 'file://User/folder/test.go'.split('/') 
>(5) ['file:', '', 'User', 'folder', 'test.go']
> 'file://User/folder/test.go'.split('/').pop()
> 'test.go'

Expected result of this function should preserve nested directories path:
> `/User/folder/test.go`

Could you confirm this logic? Using deeply nested paths in unit tests also would be good to emulate real inputs.

viktorsaws (changes requested):

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/documentContext.ts
makenneth: I think it's that the folders can contain sensitive information. But if Q does need the path for project information, we can close this out. I didn't think it had any functional use since we are extracting the context already.

PR #426, opened 2024-08-20
Changes to support customization/override/extension of methods of Jso…

suprajaven:
…nLanguageService

## Problem
In the current implementation, having any custom logic for the JsonLanguageService, i.e, overriding or extending the methods of the JsonLanguageService is not possible. This PR aims to create that feature. It extends the same logic to the YAML Service as well. 

## Changes
1. The JSONLanguageService was already exported and theoretically extensible. But it was getting implicitly created from the server function and no user was creating it prior to this change. 
2. Moved logic of resolution of URL with the getSchema method to the Service layer and created it as part of the ctor itself -> This meant that anyone creating a custom class without their custom URI Resolver could now have the the schema correctly resolved. This logic did not have any significance in the server layer and hence could be moved. 
3. Updated the createJsonLanguageServer method to now include a custom JsonLanguageService that could be passed. 
4. Added a new method for creating the server for anyone who wants to use custom services. 
5. Updated the test cases, extracted out common LspServer creation logic and added new testcases for the same. 
6. Refactored the index.ts - extracted the common methods to a common.ts and created two files, one for testing earlier behaviour and other with custom JsonLanguageService passed. 
7. All changes are replicated similarly for the YAML Server as well/ 

## Testing 

1. npm run test passes - two newer tests to test this functionality of overriding methods added, both pass as well. 

webpack 5.93.0 compiled with 1 warning in 932 ms
✔ Test JsonServer with CloudFormation schema (0.818791ms)
✔ Test JsonServer with CloudFormation schema and a custom implementation of the JsonLanguageService (0.035709ms)

  ✔ should return completion items for CloudFormation, JSON (653ms)
  ✔ should return hover item, JSON
  ✔ should return diagnostic items, JSON
  ✔ should return format items, JSON
  ✔ should return customized diagnostic items, JSON (666ms)
  ✔ should return customized hover item, JSON

  6 passing (2s)

✔ Test YamlServer with CloudFormation schema (0.807459ms)
✔ Test YamlServer with CloudFormation schema and a custom implementation of the YamlLanguageService (0.039375ms)

  ✔ should return hover item without header and footer, YAML (1282ms)
  ✔ should return diagnostic items, YAML (603ms)
  ✔ should return format items, YAML
  ✔ should return completion items for CloudFormation, YAML (1852ms)
  ✔ should return customized diagnostic items, YAML (676ms)
  ✔ should return customized hover item, YAML (1142ms)

## License


imykhai (approved):

server/aws-lsp-json/src/language-server/jsonServer.ts
```diff
@@ -133,22 +133,9 @@ export const JsonServerFactory =
...
-    uriResolver: (url: string) => Promise<string> = getSchema
-) =>
-    JsonServerFactory(
-        create({
-            defaultSchemaUri: defaultSchemaUri,
-            allowComments: allowComments,
-            uriResolver: uriResolver,
-        })
-    )
+    customServiceClass?: JsonLanguageService,
```
andredcoliveira, line 138: Since a) this approach breaks backwards incompatibility and b) a customServiceClass means that the other props would no longer be applicable—even though for example `defaultSchemUri` would still be mandatory—, I think a nice alternative here is to export a new function instead:

```ts
export function createCustomJsonLanguageServer(customService: JsonLanguageService) {
    return JsonServerFactory(new MutuallyExclusiveLanguageService([customService]));
}
```

suprajaven: Thanks for the callout - Updated to use the alternative approach with a new method for when a custom language service is needed.

server/aws-lsp-json/src/language-server/jsonServer.ts
```diff
@@ -133,17 +138,14 @@ export const JsonServerFactory =
...
-    return schema
+export function createCustomJsonLanguageServer(customService: JsonLanguageService) {
+    return JsonServerFactory(new MutuallyExclusiveLanguageService([customService]))
 }
 
 export const CreateJsonLanguageServer = (
     defaultSchemaUri: string,
-    allowComments: boolean = true,
-    uriResolver: (url: string) => Promise<string> = getSchema
+    allowComments?: boolean,
```
andredcoliveira, line 147: Is it a miss or intentional that we're changing the default value for `allowComments`? It used to be set to `true` in this line when it wasn't defined, but now it stays `undefined` in this function and when it reaches [jsonLanguageService.ts#L31](https://github.com/aws/language-servers/blob/dd76df1d5c6f7741c34494b516ba66224a78de34/server/aws-lsp-json/src/language-service/jsonLanguageService.ts#L31) it will become `false` instead.

Technically it's a breaking change too / not backwards compatible, so I recommend we avoid it

app/aws-lsp-json-runtimes/src/tests/jsonServerCFInteg.test.ts
```diff
@@ -12,41 +12,51 @@ import {
...
-        process = spawn('node', [runtimeFile, '--stdio'], {
-            shell: true,
-            stdio: 'pipe',
-        })
+    // Start the LSP server
+    // eslint-disable-next-line prefer-const
+    process = spawn('node', [runtimeFile, '--stdio'], {
+        shell: true,
+        stdio: 'pipe',
+    })
```
andredcoliveira, line 28: With your changes you can also change the `process` assignment to a `const` assignment and that would result in not having to disable the `eslint` rule

```suggestion
    // Start the LSP server
    const process = spawn('node', [runtimeFile, '--stdio'], {
        shell: true,
        stdio: 'pipe',
    })
```

app/aws-lsp-json-runtimes/src/tests/jsonServerCFInteg.test.ts
```diff
@@ -12,41 +12,51 @@ import {
...
+}
+
+describe('Test JsonServer with CloudFormation schema', () => {
+    let client: LspClient
+    let endpoint: JSONRPCEndpoint
+    let process: ChildProcessWithoutNullStreams
+    const runtimeFile = path.join(__dirname, '../../', 'build', 'aws-lsp-json-standalone.js')
+
+    before(async () => {
+        ;({ client, endpoint, process } = await createLSPServer(runtimeFile))
```
andredcoliveira, line 59: Is the semicolon (`;`) needed?

```suggestion
        ({ client, endpoint, process } = await createLSPServer(runtimeFile))
```

app/aws-lsp-json-runtimes/src/tests/jsonServerCFInteg.test.ts
```diff
@@ -141,4 +151,55 @@ describe('Test JsonServer with CloudFormation schema', () => {
...
-})
+}),
+    describe('Test JsonServer with CloudFormation schema and a custom implementation of the JsonLanguageService', () => {
+        let client: LspClient
+        let endpoint: JSONRPCEndpoint
+        let process: ChildProcessWithoutNullStreams
+        const runtimeFile = path.join(__dirname, '../../', 'build', 'aws-lsp-json-standalone-with-customization.js')
+
+        before(async () => {
+            ;({ client, endpoint, process } = await createLSPServer(runtimeFile))
```
andredcoliveira, line 162: Is the semicolon (`;`) needed?

```suggestion
            ({ client, endpoint, process } = await createLSPServer(runtimeFile))
```

andredcoliveira (commented):
Left a few comments.

Another question: do we plan on doing similar changes for the YAML server too?

server/aws-lsp-json/src/language-server/jsonServer.ts
suprajaven: Sure, thanks for the callout - I shall update it. 
And yes, there will be a followup to make this changes in Yaml too.

app/aws-lsp-json-runtimes/src/tests/jsonServerCFInteg.test.ts
suprajaven: No - this got added automatically added by one of the formatters - I will disable it and revert to the right way.

app/aws-lsp-json-runtimes/src/tests/jsonServerCFInteg.test.ts
suprajaven: Same as before - auto added from a vscode formatter, will update.

app/aws-lsp-json-runtimes/src/tests/jsonServerCFInteg.test.ts
andredcoliveira: If it's auto-formatted it's fine. It's just not frequent to see it and in this specific case it doesn't seem required. But no harm in keeping it.

andredcoliveira (approved):
Looks good to me

PR #429, opened 2024-08-22
fix(amazonq): fix LF doesn't trigger

zixlin7:
## Problem
the length of `\n` should be 1
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/auto-trigger/autoTrigger.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/auto-trigger/autoTrigger.ts
@@ -53,7 +53,7 @@ export const triggerType = (fileContext: FileContext): CodewhispererAutomatedTri
     }
 
     const lastLF = fileContext.leftFileContent.lastIndexOf('\n')
-    if (lastLF >= 0 && fileContext.leftFileContent.substring(lastLF + 2).trim() === '') {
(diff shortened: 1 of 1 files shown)


shruti0085 (approved):

leigaol (approved):

server/aws-lsp-codewhisperer/src/language-server/auto-trigger/autoTrigger.ts
```diff
@@ -53,7 +53,7 @@ export const triggerType = (fileContext: FileContext): CodewhispererAutomatedTri
     }
 
     const lastLF = fileContext.leftFileContent.lastIndexOf('\n')
-    if (lastLF >= 0 && fileContext.leftFileContent.substring(lastLF + 2).trim() === '') {
+    if (lastLF >= 0 && fileContext.leftFileContent.substring(lastLF + 1).trim() === '') {
```
volodkevych, line 56: Can we have test for this?

volodkevych (approved):

server/aws-lsp-codewhisperer/src/language-server/auto-trigger/autoTrigger.ts
andrewyuq: added a TODO for the test and tracked internally.

ege0zcan:
Would this check work for both windows line endings and unix line endings? Is it possible that this check doesn't cover all the potential cases?

PR #433, opened 2024-08-23
Ensure that the entire chat request is cancellable

makenneth:
## Description

The main goal of this PR is to make entire chat request cancellable. Currently, if user cancels before the response streaming starts, the request will still go through. For the fqn piece, we can use Workerpool's "cancel" feature. Otherwise, we just need to early terminate and return a `ResponseError` with `RequestCancelled` code.

I tried to make this reusable by creating a wrapper function that creates a promise for cancellation that will "race" against the core logic: the value that is resolved first will be returned. 

The other two changes:
- Combined `DocumentContext` and `TriggerContext` into a single `TriggerContextExtractor` as the separation offer no benefits
- Removed unneeded start trigger information

## Testing
- Regression tested in VS
- Also tested the end chat flow with temporary UI implementation (not included in this PR)

## License


client/vscode/src/chatActivation.ts
```diff
@@ -54,26 +55,40 @@ export function registerChat(languageClient: LanguageClient, extensionUri: Uri,
         languageClient.info(`vscode client: Received ${JSON.stringify(message)} from chat`)
 
         switch (message.command) {
+            case END_CHAT_REQUEST_METHOD:
```
ege0zcan, line 58: i would expect this case to be already taken care of by the 'default' case, I think we can remove this specific case

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
```diff
@@ -35,14 +35,9 @@ interface MessageTrigger extends TriggerContext {
     followUpActions?: Set<string>
 }
 
-interface StartTrigger {
-    triggerType?: TriggerType
-    hasUserSnippet?: boolean
```
ege0zcan, line 40: does this mean we never logged `hasUserSnippet` field anywhere so we won't be logging it going forward?

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -67,108 +67,123 @@ export class ChatController implements ChatHandlers {
...
+            try {
+                this.#log('Request for conversation id:', conversationIdentifier)
+                const requestInput = TriggerContextExtractor.getChatParamsFromTrigger(params, triggerContext)
+
+                metric.recordStart()
+                response = await session.generateAssistantResponse(requestInput)
+                this.#log('Response for conversationId:', conversationIdentifier, JSON.stringify(response.$metadata))
+            } catch (err) {
+                if (isObject(err) && 'name' in err && err.name === 'AbortError') {
+                    throw new CancellationError('Q api request aborted')
```
ege0zcan, line 96: here and down below, why do we throw cancellation error instead of returning an error object like we do in other error scenarios?

makenneth: I added an upstream error handling in `withLspCancellation`. I guess we should probably do one of the other

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
makenneth: IIRC currently we are not using `hasUserSnippet` from the start trigger but from message trigger. (But it could be a mistake since Q would have the snippet if the first request has it.)

ege0zcan:
I tested this change locally and it works, the changes in this PR LGTM.  

However currently the backend doesn't support cancellations and returns an improperly formed request error when you cancel a request and make a new request on the same chat afterwards. Therefore we should not merge this PR yet. 

I will update this PR with the related changes on the chat client as well for completeness

PR #437, opened 2024-08-26
Set customUserAgent for SDK calls in Q Servers

viktorsaws:
## Problem

AWS SDK API calls from Amazon Q Servers have default user agent.

## Solution

Use `customUserAgent` provided from the Runtime process during initialization and use it to configure AWS SDKs used in different Q Service classes.

The userAgent value will be computed using information passed in new object in `initializationOptions.aws` object or fallback to using standard `InitializeParams.clientInfo` field

## User Agent value example

With this change, custom user agent set by CW Language servers will contain next suffix:
`AWS-Language-Servers AWS-CodeWhisperer/x.y.z AWS-Toolkit-For-VSCode/x.y.z Visual-Studio-Code/x.y.z ClientId/$UUID`

## License


server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -61,6 +61,13 @@ export abstract class CodeWhispererServiceBase {
         AWS.config.update(awsConfig)
     }
 
+    /**
+     * Updates Service Client options after client was instantiated.
+     */
+    public updateClientConfig(options: ConfigurationOptions) {
```
rahmaniaam, line 67: Do we need this? AFAIK `updateAwsConfiguration` would apply globally so we can utilise that instead

viktorsaws: IDK and I didn't test that, that's why I created this specific implementation

viktorsaws: Thinking about this a bit more, I'd avoid setting global AWS configs from individual Language Server implementation, if it has side effects on other servers. This may result in unexpected result, where one server overrides settings in others servers

imykhai: Will different servers have different user agent info?

imykhai (approved):

server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
viktorsaws: Authors are free to set it according to their service expectations. Since SDK clients are part of server codebase, they can choose to any value and we can't/shouldn't enforce it

PR #441, opened 2024-08-27
Add retry to pollTransformation

mrkdeng:
## Problem

In the `transformHandler.ts`, polling to get transform status will return `FAILED` as job status if caught one error. But the error can be caused by network issue or the API call timeout.

## Solution
This PR adds the retry policy to `pollTransformation` from the server, it will retry 3 times after the first failure occurs. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
@@ -196,7 +196,7 @@ export class TransformHandler {
     async getTransformationPlan(request: GetTransformPlanRequest) {
         let getTransformationPlanAttempt = 0
         let getTransformationPlanMaxAttempts = 3
-        while (getTransformationPlanAttempt <= getTransformationPlanMaxAttempts) {
+        while (true) {
             try {
                 const getCodeTransformationPlanRequest = {
                     transformationJobId: request.TransformationJobId,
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -289,10 +290,17 @@ export class TransformHandler {
...
+                if (status === 'Retry' && getTransformRetry === getTransformMaxRetry) {
+                    const errorMessage = (e as Error).message ?? 'Error in GetTransformation API call'
+                    this.logging.log('CodeTransformation: GetTransformation error = ' + errorMessage)
+                    status = 'FAILED'
+                    break
+                } else if (status === 'Retry') {
+                    getTransformRetry += 1
+                } else {
+                    status = 'Retry' // first time failed, status will be overwritten if no failure in next poll
+                    getTransformRetry = 0
```
jonlouie, line 302: Imagine the experience as the end user. Would it be helpful to see the error and that the system is retrying? Would the current attempt and max attempt count also help?

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -289,10 +290,17 @@ export class TransformHandler {
...
                 }
             } catch (e: any) {
-                const errorMessage = (e as Error).message ?? 'Error in GetTransformation API call'
-                this.logging.log('CodeTransformation: GetTransformation error = ' + errorMessage)
-                status = 'FAILED'
-                break
+                if (status === 'Retry' && getTransformRetry === getTransformMaxRetry) {
+                    const errorMessage = (e as Error).message ?? 'Error in GetTransformation API call'
+                    this.logging.log('CodeTransformation: GetTransformation error = ' + errorMessage)
+                    status = 'FAILED'
```
jonlouie, line 296: Can we enumerate the polling status values? So far, I've seen:
* Retry
* FAILED
* Timed_out
* NOT_FOUND

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -289,10 +290,17 @@ export class TransformHandler {
...
+                    break
+                } else if (status === 'Retry') {
+                    getTransformRetry += 1
```
pranav-firake, line 299: can we add logs here that says something like 'Retrying for x th time?'

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -241,7 +241,8 @@ export class TransformHandler {
 
     async pollTransformation(request: GetTransformRequest, validExitStatus: string[], failureStates: string[]) {
         let timer = 0
-
+        let getTransformRetry = 0
```
pranav-firake, line 244: thanks for adding a logic here to have retries. also can we also explore if this will be helpful if we add similar retries in getTransformPlan and CancelTransform

jonlouie: If we add this logic to other functions, consider creating a wrapper function with retries that takes any function as a parameter along with the number of retries.

ege0zcan (approved):

PR #462, opened 2024-09-19
Create Q Customization Server

ege0zcan:
## Description

- Added Q configuration server that responds to get configuration requests with customization information
- Updated the sample extension with a command to get customizations from the server as well as setting customization through workspace configuration
- Refactored the  `updateConfiguration` logic in the inline completion server and tidied up the code

## License


server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -260,18 +261,6 @@ const mergeSuggestionsWithRightContext = (
     })
 }
 
-// Checks if any suggestion in list of suggestions matches with left context of the file
-const hasLeftContextMatch = (suggestions: Suggestion[], leftFileContent: string): boolean => {
```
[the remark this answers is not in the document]
ege0zcan: Yes, it was unused

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -526,26 +515,31 @@ export const CodewhispererServerFactory =
...
-                        logging.log('Configuration updated to share Amazon Q content with AWS')
-                    } else {
-                        codeWhispererService.shareCodeWhispererContentWithAWS = false
-                        logging.log('Configuration updated to not share Amazon Q content with AWS')
-                    }
-                })
-                .catch(reason => logging.log(`Error in GetConfiguration: ${reason}`))
+        const updateConfiguration = async () => {
+            try {
+                const qConfig = await lsp.workspace.getConfiguration(Q_CONFIGURATION_SECTION)
```
pras0131, line 520: I think here we just need to make sure we send the request with the correct parameters. According to LSP spec, https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#configurationParams an array of items has to be sent.

pras0131: I checked locally and it seems internally it somehow sends the request in the expected correct format. Can ignore this comment.

viktorsaws: FYI, under the hood `getConfiguration` is handled by this implementation, which converts arguments to LSP params 
https://github.com/microsoft/vscode-languageserver-node/blob/f58f4dff16ad2760028bb1cb95e882de30a1000f/server/src/common/configuration.ts#L4

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -526,26 +515,31 @@ export const CodewhispererServerFactory =
...
+                if (qConfig) {
+                    codeWhispererService.customizationArn = qConfig.customization ?? ''
```
viktorsaws, line 522: Is empty string is valid ARN value as fallback?

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -526,26 +515,31 @@ export const CodewhispererServerFactory =
...
+                    logging.log(`Inline completion configuration updated to use ${qConfig.customization}`)
+                }
+
+                const config = await lsp.workspace.getConfiguration('aws.codeWhisperer')
```
viktorsaws, line 526: nit: it should be possible to request multiple configuration sections with one `getConfiguration` call, instead of 2 https://github.com/microsoft/vscode-languageserver-node/blob/f58f4dff16ad2760028bb1cb95e882de30a1000f/server/src/common/configuration.ts#L18.

Our `Workspace.getConfiguration` proxy handler only supports `section` though, so it will need to be extended. Could be done as a follow-up.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -49,6 +49,7 @@ export abstract class CodeWhispererServiceBase {
     protected readonly codeWhispererRegion = 'us-east-1'
     protected readonly codeWhispererEndpoint = 'https://codewhisperer.us-east-1.amazonaws.com/'
     public shareCodeWhispererContentWithAWS = false
+    public customizationArn: string = ''
```
viktorsaws, line 52: nit: better to use `undefined` or `null` as negative fallback, instead of empty string. JS is weird ofc with treating empty string as false, but we can be explicit :)

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -0,0 +1,44 @@
...
+    CredentialsProvider,
+    GetConfigurationFromServerParams,
+    InitializeParams,
+    Server,
+} from '@aws/language-server-runtimes/server-interface'
+import { CodeWhispererServiceToken } from '../codeWhispererService'
+import { getUserAgent } from '../utils'
+
+// The configuration section that the server will register and listen to
+export const Q_CONFIGURATION_SECTION = 'aws.q'
```
viktorsaws, line 12: Could you add documentation for `QConfigurationServerToken` to README with an configuration key it supports and data type of response it provides? 
Also you can define a Typescript type for `onGetConfigurationFromServer` response, and it will serve as a documentation as well.

viktorsaws (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -330,6 +339,18 @@ export class ChatController implements ChatHandlers {
         return chatEventParser.getChatResult()
     }
 
+    updateConfiguration = async () => {
+        try {
+            const qCustomizationConfig = await this.#features.lsp.workspace.getConfiguration(Q_CONFIGURATION_SECTION)
```
viktorsaws, line 344: The Typescript type for expected configuration object defined in https://github.com/aws/language-servers/pull/462#discussion_r1777211491 can be used in calls to read documentation like here.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
pras0131: thanks for sharing the link Viktor.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
pras0131: I think instead of this, we can have a check like `if(qConfig && qConfig.customization)`

PR #463, opened 2024-09-20
Attach supplemental context to InlineCompletions request

viktorsaws:
## Problem
We don't attach cross file context Q InlineCompletion requests.

## Solution

* Migrate implementation of code to fetch supplemental context from open files in workspace from VSCode Q extension implementation https://github.com/aws/aws-toolkit-vscode/blob/4c3e64cc1987328ad59e2f879eaefd66dff2192a/packages/core/src/codewhisperer/util/supplementalContext/supplementalContextUtil.ts#L15
* Attaching cross file context for test files (UTG context) is out of scope.
 
Entry point for using supplementalContext changes start in [CodewhispererServer.ts](https://github.com/aws/language-servers/pull/463/files#diff-41618063d6869da8a8fcb2326f67e08e197e7a04c01cfab0b832996e320c653cR373-R380). Supplemental context is computed for requested TextDocument, cached in Inline completion Session and attached to `generateSuggestions` service call. Then, supplemental context fields also added to CodeWhisperer Telemetry events, emitted from the server.

All computation logic and helper functions are ported as is, with only adapting them to work with LSP documents instead of VSCode APIs. Most unit tests also migrated as is, with adding extra tests for coverage.

This PR requires https://github.com/aws/language-server-runtimes/pull/214/ to be released in Runtimes.
Diff between original code and new code introduced here can be seen in https://github.com/viktorsaws/language-servers/pull/5.

### TODO

- [x] Unit tests

## License


server/aws-lsp-codewhisperer/src/language-server/utilities/filesystemUtilities.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ *
+ * @returns file distance between fileA and fileB
+ * For example:
+ * The file distance between A/B/C.java and A/B/D.java is 0
+ * The file distance between A/B/C.java and A/D.java is 1
+ */
+export function getFileDistance(fileA: string, fileB: string): number {
+    // TODO: Test how it works with nested folders with spaces (%20 symbol)
```
ege0zcan, line 14: This todo is not coming from the toolkit codebase but something we are introducing, why is that?

server/aws-lsp-codewhisperer/src/language-server/utilities/supplementalContextUtil/supplementalContextUtil.ts
```diff
@@ -0,0 +1,73 @@
...
+export function fetchSupplementalContext(
+    document: TextDocument,
+    position: Position,
+    workspace: Workspace,
+    logging: Logging,
+    cancellationToken: CancellationToken
+): CodeWhispererSupplementalContext | undefined {
+    const timesBeforeFetching = performance.now()
+
+    const isUtg = isTestFile(document.uri, {
```
ege0zcan, line 26: I don't have full context, what does UTG actually stand for?

viktorsaws: I think it stands for "unit test generator". I don't know how significant this meaning is, tbh.

server/aws-lsp-codewhisperer/src/language-server/telemetry/types.ts
```diff
@@ -20,6 +20,10 @@ export interface CodeWhispererServiceInvocationEvent {
     codewhispererGettingStartedTask?: string
     reason?: string
     credentialStartUrl?: string
+    codewhispererSupplementalContextTimeout?: boolean
```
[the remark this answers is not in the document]
viktorsaws: I agree, but this is a Telemetry object field, as defined by CodeWhisperer service. We don't control the name and expected type here.

server/aws-lsp-codewhisperer/src/language-server/utilities/filesystemUtilities.ts
viktorsaws: this is something I noticed when comparing logs for VSCode Q and Language Server. In Language server, files URIs are encoded and spaces are converted to `%20` string. Since this is different from original environment, I wanted to double check that in this case this function works. I added few tests to cover this and will remove this comment,

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -573,6 +605,28 @@ export const CodewhispererServerFactory =
 
         logging.log('Amazon Q Inline Suggestion server has been initialised')
 
+        const logSupplementalContext = (supplementalContext: CodeWhispererSupplementalContext | undefined) => {
```
ege0zcan, line 608: Isn't this too verbose, i see in vscode toolkit codebase they use debug logging to log this information whereas it seems we log always

server/aws-lsp-codewhisperer/src/language-server/telemetry/userTriggerDecision.test.ts
```diff
@@ -760,15 +763,15 @@ describe('Telemetry', () => {
                     aUserTriggerDecision({
                         codewhispererSessionId: 'cwspr-session-id-0',
                         codewhispererSuggestionState: 'Discard',
-                        codewhispererTimeToFirstRecommendation: 2520,
+                        codewhispererTimeToFirstRecommendation: 1260,
```
ege0zcan, line 766: was this an intentional change?

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
viktorsaws: hm, yeah you're right. I added it as normal log since we don't have different log levels implemented.

server/aws-lsp-codewhisperer/src/language-server/telemetry/userTriggerDecision.test.ts
viktorsaws: Yes. This is caused by changing nested callback in CodeWhispererServer to `async` and I suspect it changes the chain of async callbacks resolution. The test emulates several inflight requests, and only this time in test environment is impacted. 
In my manual E2E testing with compiled server, times are at same rate as before this change. I didn't want to refactor this test as it too complex :(

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
viktorsaws: Removed excessive logging in last commit

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.test.ts
```diff
@@ -467,6 +474,39 @@ describe('CodeWhisperer Server', () => {
...
+                    fileContext: {
+                        filename: 'file:///TargetFile.java',
+                        programmingLanguage: { languageName: 'java' },
+                        leftFileContent: '',
+                        rightFileContent: '',
+                    },
+                    maxResults: 5,
+                    supplementalContexts: [
+                        { content: 'sample-content', filePath: '/SampleFile.java' },
+                        { content: 'sample-content', filePath: '/SampleFile.java' },
```
volodkevych, line 503: Why does it send the same file twice?

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.test.ts
```diff
@@ -467,6 +474,39 @@ describe('CodeWhisperer Server', () => {
             assert.deepEqual(result, EMPTY_RESULT)
         })
 
+        describe('Supplemental Context', () => {
+            it('should send supplemental context', async () => {
+                // Open 3 files supporting cross-file context
```
volodkevych, line 479: nit: 2 files?

server/aws-lsp-codewhisperer/package.json
```diff
@@ -21,12 +21,13 @@
         "compile": "tsc --build",
         "postcompile": "npm run copyServiceClient",
         "copyServiceClient": "copyfiles -u 1 --error ./src/client/sigv4/*.json out && copyfiles -u 1 --error ./src/client/token/*.json out",
+        "fix": "npm run fix:prettier",
```
volodkevych, line 24: nit: do we really need this duplication?

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -191,6 +202,9 @@ const emitAggregatedUserTriggerDecisionTelemetry = (
             : undefined,
         codewhispererTimeToFirstRecommendation: session.timeToFirstRecommendation,
         codewhispererPreviousSuggestionState: session.previousTriggerDecision,
+        codewhispererSupplementalContextTimeout: session.supplementalMetadata?.isProcessTimeout,
+        codewhispererSupplementalContextIsUtg: session.supplementalMetadata?.isUtg,
+        codewhispererSupplementalContextLength: session.supplementalMetadata?.contentsLength,
```
volodkevych, line 207: Is it expected we set 3 props in this metric, but 4 above? The tests are checking for 4 properties, at least.

PR #472, opened 2024-10-01
Add initial set up for aws-lsp-identity package, including sharedConfig

floralph:
Adding initial set up for the aws-lsp-identity server and getting wired into overall build process.  Adding sharedConfig files to support saving shared config files used by an upcoming PR for profile manipulation.  Additionally, DuckTyper will be used for upcoming profile service as well.

sharedConfig based on code in @smithy/shared-ini-file-loader, adds saveKnownFiles as counterpart to parseKnownFiles from that package to persist config & credentials files while maintaining as much of the user's existing file as possible.

Unit tests for this classes have been added and verify to pass along with all existing Flare tests.


server/aws-lsp-identity/src/utils/duckTyper.ts
```diff
@@ -0,0 +1,56 @@
...
+        this.rules[name] = DuckTyper.optionalPropertyRule
+        return this
+    }
+
+    disallowProperty(name: string): DuckTyper {
+        this.rules[name] = DuckTyper.disallowPropertyRule
+        return this
+    }
+
+    eval(value: object, onlyDefined: boolean = false): boolean {
```
volodkevych, line 24: What does `onlyDefined` mean? Why is it needed? Maybe worth putting comment with explanation.

server/aws-lsp-identity/src/sharedConfig/unmergeConfigFiles.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+}
+
+function isSecretSetting(name: string): boolean {
+    // Settings containing "secrets" as defined in the public docs are to be stored in creds
+    // if not already defined in config (this is for "retain" support, future options may
+    // allow forcing secrets to always be written to credentials, existing or not).
+    // https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html#file-format-creds
+    return ['aws_access_key_id', 'aws_secret_access_key', 'aws_session_token'].includes(name)
```
volodkevych, line 105: Are they going to be always lowercase?

server/aws-lsp-identity/src/sharedConfig/unmergeConfigFiles.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // Remove empty sections
+    !Object.keys(configSection).length && delete configFile[sectionName]
+    !Object.keys(credentialsSection).length && delete credentialsFile[sectionName]
+}
+
+export function removeDeletedEntries(args: { removeFrom: Record<string, unknown>; deletedFrom: object }): void {
+    const { removeFrom, deletedFrom } = args
+
+    if (!(removeFrom && deletedFrom)) {
```
volodkevych, line 89: nit: checking on `deletedFrom` should be enough

server/aws-lsp-identity/src/sharedConfig/unmergeConfigFiles.ts
floralph: Good catch, while all settings are lowercase in public documents, none state whether the files are case-sensitive or not.  Seems like this has caused some questions/problems in the past (see links below).  It appears that the AWS CLI is case-insensitive, but most AWS SDKs are case-sensitive.

Made changes to take the approach of being able to read any setting name regardless of case, but when writing, setting names are converted to lowercase to make coding in TypeScript easier (as property names are case-sensitive) and to promote better compatibility with AWS SDKs on file saves.

https://github.com/aws/aws-toolkit-vscode/issues/2394
https://github.com/aws/aws-sdk-ruby/issues/2458
https://github.com/aws/aws-sdk/issues/574
https://github.com/awslabs/aws-sdk-rust/issues/1020

server/aws-lsp-identity/src/sharedConfig/unmergeConfigFiles.ts
volodkevych: If `removeFrom` is falsy, we'll never get to `if` with `deletedFrom` in the code below.

viktorsaws (approved):

PR #475, opened 2024-10-03
Add listProfiles and updateProfile with ProfileService and store

floralph:
NOTE: Automated builds here are pending push of language-server-runtimes PR https://github.com/aws/language-server-runtimes/pull/218

This change builds on the prior PR (https://github.com/aws/language-servers/pull/472) that added persistence for shared config files.  It adds listProfiles and updateProfile API implementations and provides an initial IdentityServer implementation in which additional APIs will be added.

The ProfileService takes a ProfileStore implementation (supports load/save to appropriate store).  A SharedConfigProfileStore has been provided for standalone runtime support.  If needed in the future, a webworker runtime could have a separate ProfileStore implementation (e.g. BrowserStorageProfileStore).  Shared Config and Profile name spaces have been kept separate from general identity server types to make refactoring those types to a new location (such as in the runtime) if needed in the future.  

Some supporting functions/types were added to the ProfileService module as well.  

The majority of the code in this CR is in the unit tests.

Minor changes:
1. Adds a launch.json task to run ts-mocha against the current file in VSC
2. Add usage of chai in place of node assert in new tests
3. Change import of mock-fs to be compatible with TypeScript debugging/testing


server/aws-lsp-identity/src/language-server/identityServer.ts
```diff
@@ -0,0 +1,31 @@
...
+    listProfilesRequestType,
+    profileServer.listProfiles.bind(profileServer)
+)
+
+connection.onRequest<UpdateProfileParams, UpdateProfileResult, void, UpdateProfileError, void>(
+    updateProfileRequestType,
+    profileServer.updateProfile.bind(profileServer)
+)
+
+connection.listen()
```
viktorsaws, line 31: Why is this server is implemented as separate connection listener? This essentially makes it a separate standalone language server implementation, which can't be bundled with Runtimes Server implementation following Runtimes framework.

Standalone Runtime implementation exposes `identityManagement` feature and manages LSP connection, so this server does not have to work on raw LSP connection. https://github.com/aws/language-server-runtimes/blob/main/runtimes/runtimes/standalone.ts#L254-L264

Also this implementation does not allow to bundle this IdentityServer with other Servers in one bundle and will always require separate file.

server/aws-lsp-identity/src/language-server/profiles/profileService.ts
```diff
@@ -0,0 +1,245 @@
+import {
+    ListProfilesError,
+    ListProfilesParams,
+    ListProfilesResult,
+    Profile,
+    ProfileKind,
+    SsoSession,
+} from '@aws/language-server-runtimes/protocol/identity-management'
```
viktorsaws, line 8: nit: for Servers implemented in `language-servers` repo recommendation is to import types from `@aws/language-server-runtimes/server-interface` directory instead of `protocol`. `server-interface` is setup to re-export most of `protocol` where it make sense, so all there types must be available.

Since Servers implementations are decoupled from LSP Clients using `/server-interface`, this provides better separation of concerns between what Client and Server should implement. `/protocol` is intended to be used by  Client extensions implementors.

server/aws-lsp-identity/src/language-server/profiles/profileService.ts
```diff
@@ -0,0 +1,245 @@
...
+
+        // Validate params
+        this.throwOnInvalidProfile(!params.profile, 'Profile required.')
+        const profile = params.profile!
+        this.throwOnInvalidProfile(
+            profile.kind !== ProfileKind.SsoTokenProfile,
+            'Profile must be non-legacy sso-session type.'
+        )
+        this.throwOnInvalidProfile(!profile.name, 'Profile name required.')
+        this.throwOnInvalidProfile(!profile.settings.sso_session, 'Sso-session name required on profile.')
```
viktorsaws, line 109: Would this throw if `profile.settings` is undefined?

viktorsaws: Also same below at with `!ssoSession.settings`

server/aws-lsp-identity/src/language-server/identityServer.ts
floralph: Yep, I realized this last night as I started implementing and testing SsoTokenChanged.  I will pull over the updates from that PR for IdentityServer and related files to this one.

server/aws-lsp-identity/src/language-server/profiles/profileService.ts
floralph: I ran into this problem last night too.  identity-management wasn't exported with server-interface, which is why they were imported from protocol.  Please check this PR https://github.com/aws/language-server-runtimes/pull/219 to add identity-management to server-interface then I will update these imports here.

server/aws-lsp-identity/src/language-server/profiles/profileService.ts
floralph: No as the DuckTyper upstream would've failed and the flow would've never made it this far.  However, for defense in-depth and the possibility that this code is used in a different flow in the future, adding a check and unit tests for it.

server/aws-lsp-identity/src/language-server/identityServer.ts
viktorsaws: got it :)

server/aws-lsp-identity/src/language-server/profiles/profileService.ts
```diff
@@ -0,0 +1,243 @@
...
+    UpdateProfileError,
+    UpdateProfileErrorData,
+    updateProfileOptionsDefaults,
+    UpdateProfileParams,
+    UpdateProfileResult,
+    AwsErrorCodes,
+    AwsResponseError,
+    AwsResponseErrorData,
+} from '@aws/language-server-runtimes/server-interface/identity-management'
+import { CancellationToken } from 'vscode-languageserver'
```
viktorsaws, line 17: nit: `server-interface` re-exports `CancellationToken`, so `vscode-languageserver` dependency can be omitted

viktorsaws (approved):

server/aws-lsp-identity/src/language-server/profiles/profileService.ts
floralph: I'll update this in the upcoming PR to avoid further delay on this one.

ivikash (approved):

PR #476, opened 2024-10-03
[Team Review] Add logging when polling get transformation status failed

JiayuW2:
## Problem
The "FAILED" status during polling transformation is typically due to build validation failures. Currently, no guidance is logged when the job status fails. 

## Solution
Add logs to recommend that customers clear the binary cache and retry.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
@@ -356,6 +356,12 @@ export class TransformHandler {
             }
         }
         this.logging.log('poll : returning response from server : ' + JSON.stringify(response))
+        if (failureStates.includes(status)) {
+            this.logging.log(
+                `Transformation job for job ${request.TransformationJobId} is ${status} due to "${response.transformationJob.reason}". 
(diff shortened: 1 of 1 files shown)


huawenm (approved):

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -356,6 +356,12 @@ export class TransformHandler {
             }
         }
         this.logging.log('poll : returning response from server : ' + JSON.stringify(response))
+        if (failureStates.includes(status)) {
+            this.logging.log(
+                `Transformation job for job ${request.TransformationJobId} is ${status} due to "${response.transformationJob.reason}". 
+                Please close Visual Studio, delete the bin and obj folders for the projects, and try running the transformation again.`
```
jonlouie, line 362: Update for next time: 

1. Suggestion:  `Please close Visual Studio, delete the directories where build artifacts are generated (e.g. `bin` and `obj`), and try running the transformation again.`
2. We should only suggest the above when the failure reason is "Build validation failed..."
3. Consider making a helper function called `LogSuggestionForFailureResponse(request, response, status)`. There are probably other places where we log some suggestion after a failure message and this will keep the `if-else` logic contained in another function.

PR #488, opened 2024-10-09
Add automated workflow to publish chat-client to npm registry

viktorsaws:
## Problem
`@aws/chat-client` package is not publish to npm, and proper semver versioning history was not maintained.

## Solution

Added automated workflow to publish new version of `chat-client` to npm. Current version `0.0.4` is published manually https://www.npmjs.com/package/@aws/chat-client

## License

### DIFF
--- a/chat-client/.npmignore
+++ b/chat-client/.npmignore
@@ -1,4 +1,7 @@
+build/
 node_modules/
+src/
 package-lock.json
 .*
 tsconfig.*
+webpack.*.config.js
--- a/.github/workflows/publish-chat-client.yaml
+++ b/.github/workflows/publish-chat-client.yaml
@@ -1,0 +1,13 @@
+name: Publish Codewhisperer Language Server to npmjs
+
+on:
+    push:
+        tags:
+            - 'chat-client/v**'
+
+jobs:
+    publish:
+        uses: ./.github/workflows/publish-to-npm.yaml
+        with:
+            workspace: 'chat-client'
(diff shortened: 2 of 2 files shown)


volodkevych (approved):

chat-client/.npmignore
```diff
@@ -1,4 +1,7 @@
+build/
 node_modules/
+src/
 package-lock.json
 .*
-tsconfig.*
\ No newline at end of file
+tsconfig.*
+webpack.*.config.js
```
ege0zcan, line 7: is the `*` intentional? this means we are publishing `webpack.*.config.js`

viktorsaws: it's intentional to prevent publishing all possible webpack.config files. Same rule works with `codewhisperer` package https://www.npmjs.com/package/@aws/lsp-codewhisperer?activeTab=code fine

ege0zcan: it prevents possible configs but not the actual config which is `webpack.config.js` (this is what I tried to say in my first comment but mistakenly put the star in the middle, updated my initial comment as well)

So we should either put: `webpack.*` or `*.config.js` here.  The reason it works in the other subpackage is because we have `webpack.lint.config.js` over there, in chat client we have  `webpack.config.js`

viktorsaws: yeah, you are right, fixed it

ege0zcan (approved):

PR #489, opened 2024-10-10
Add SsoTokenChanged, profile bug fixes, and testability of aws-lsp-id enity in vscode client

floralph:
This PR pending merge and publication of new @aws/language-server-runtimes package due to protocol updates.  See https://github.com/aws/language-server-runtimes/pull/222 for details.

## SsoTokenChanged
1. Added simple AsyncEvent implementation of event pattern to support easy usage in the identity server as well as chaining to JSON-RPC server to send.  Existing Node and JavaScript events were too heavyweight and unsuitable for this application.

## Add IdentityServer app
1. Added app/aws-lsp-identity-runtimes modeled after other apps in the app folder to support launching standalone runtime of identity server.

## Profile-related bug fixes/enhancements
1. Added outFiles to support debugging of ts-mocha launch.json configuration.
2. Removed notes about using identity-management for imports instead of service-interface after guidance in prior PR.
3. Removed unnecessary statements for collating profiles and ssoSessions with the shared config files versions in ProfileService.
4. Added ensureSsoAccountAccessScope (added to UpdateProfileParams in separate language-server-runtimes PR as well) option to UpdateProfile to allow opt-out of automatically adding sso:account:access to sso_registration_scopes.  This functionality was moved from sharedConfigProfileStorage to ProfileService.
5. Added additional comments and unit tests
6. Changed sharedConfigProfileStorage.save from removing settings that were missing to not change them in sharedConfig.  Removal now supported by explicit undefined property which is in line with how data is sent and received through LSP messages (i.e. if the client just doesn't touch a setting, the right thing happens in all cases whereas before the client couldn't really remove a setting, only force it to undefined which would be invalid in an INI file).
7. By extension of the above, if the settings property is undefined, the whole section is removed from shared config as a section with no settings makes no sense.  This keeps the behavior the same at both section and setting level and adds support for section deletes in the future.

## Testability in vscode client
1. Added launch.json entry to support debugging of aws-lsp-identity server using sample vscode client.  
8. Added "Identity: Test Command" to be able to test various functionality during the development process.  This may become more formalized later, but is sufficient to enable dev loop for now.
9. Added identityActivation modeled after other activations in the same folder.  This is where a developer can try out new APIs quickly.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -172,20 +172,6 @@
             "args": ["--forbid-only=false", "--forbid-pending=false", "./**/out/**/${fileBasenameNoExtension}.js"],
             "outFiles": ["${workspaceFolder}/**/out/**/*.(m|c|)js", "!**/node_modules/**"],
             "console": "integratedTerminal"
-        },
-        {
-            "name": "ts-mocha (current tab)",
-            "type": "node",
-            "request": "launch",
-            "program": "${workspaceFolder}/node_modules/.bin/ts-mocha",
-            "args": ["--forbid-only=false", "--forbid-pending=false", "${file}"],
-            "outFiles": [
-                "${workspaceFolder}/**/src/**/*.ts",
-                "${workspaceFolder}/**/out/**/*.ts",
-                "${workspaceFolder}/**/out/**/*.(m|c|)js",
-                "!**/node_modules/**"
-            ],
-            "console": "integratedTerminal"
         }
     ],
     "compounds": [
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -135,6 +135,25 @@
             "preLaunchTask": "npm: compile"
         },
         {
+            "name": "Identity Server + VSCode Extension",
+            "type": "extensionHost",
+            "request": "launch",
+            "runtimeExecutable": "${execPath}",
+            "args": ["--extensionDevelopmentPath=${workspaceFolder}/client/vscode", "--disable-extensions"],
+            "env": {
+                "LSP_SERVER": "${workspaceFolder}/app/aws-lsp-identity-runtimes/out/standalone.js",
+                "ENABLE_IDENTITY": "true",
+                "ENABLE_TOKEN_PROVIDER": "true"
+            },
+            "outFiles": [
+                "${workspaceFolder}/client/vscode/out/**/*.js",
+                "${workspaceFolder}/app/aws-lsp-identity-runtimes/out/**/*.js",
+                "${workspaceFolder}/server/aws-lsp-identity/out/**/*.js"
+            ],
+            "autoAttachChildProcesses": true,
+            "preLaunchTask": "watch"
+        },
+        {
             "name": "Unit Tests",
             "type": "node",
             "request": "launch",
@@ -160,10 +179,13 @@
             "request": "launch",
             "program": "${workspaceFolder}/node_modules/.bin/ts-mocha",
             "args": ["--forbid-only=false", "--forbid-pending=false", "${file}"],
-            "outFiles": ["${workspaceFolder}/**/out/**/*.(m|c|)js", "!**/node_modules/**"],
-            "console": "integratedTerminal",
-            "pauseForSourceMap": true,
-            "stopOnEntry": true
+            "outFiles": [
+                "${workspaceFolder}/**/src/**/*.ts",
+                "${workspaceFolder}/**/out/**/*.ts",
+                "${workspaceFolder}/**/out/**/*.(m|c|)js",
+                "!**/node_modules/**"
+            ],
+            "console": "integratedTerminal"
         }
     ],
     "compounds": [
--- a/client/vscode/src/identityActivation.ts
+++ b/client/vscode/src/identityActivation.ts
@@ -1,0 +1,69 @@
+import { commands, window } from 'vscode'
+import {
+    ListProfilesError,
+    ListProfilesParams,
+    ListProfilesResult,
+    ProfileKind,
+    SsoTokenChangedParams,
+    UpdateProfileError,
+    UpdateProfileParams,
+    UpdateProfileResult,
+} from '@aws/language-server-runtimes/protocol'
+
+import { LanguageClient, ProtocolNotificationType, ProtocolRequestType } from 'vscode-languageclient/node'
+
+const ssoTokenChangedRequestType = new ProtocolNotificationType<SsoTokenChangedParams, void>(
+    'aws/identity/ssoTokenChanged'
+)
+
+const listProfilesRequestType = new ProtocolRequestType<
+    ListProfilesParams,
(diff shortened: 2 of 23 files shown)


.vscode/launch.json
```diff
@@ -160,10 +179,13 @@
             "request": "launch",
             "program": "${workspaceFolder}/node_modules/.bin/ts-mocha",
             "args": ["--forbid-only=false", "--forbid-pending=false", "${file}"],
-            "outFiles": ["${workspaceFolder}/**/out/**/*.(m|c|)js", "!**/node_modules/**"],
-            "console": "integratedTerminal",
-            "pauseForSourceMap": true,
-            "stopOnEntry": true
+            "outFiles": [
+                "${workspaceFolder}/**/src/**/*.ts",
+                "${workspaceFolder}/**/out/**/*.ts",
```
volodkevych, line 184: Why would you need `ts files here, both `src` and `out`?

volodkevych (approved):

client/vscode/src/identityActivation.ts
```diff
@@ -0,0 +1,69 @@
...
+    UpdateProfileResult,
+} from '@aws/language-server-runtimes/protocol'
+
+import { LanguageClient, ProtocolNotificationType, ProtocolRequestType } from 'vscode-languageclient/node'
+
+const ssoTokenChangedRequestType = new ProtocolNotificationType<SsoTokenChangedParams, void>(
+    'aws/identity/ssoTokenChanged'
+)
+
+const listProfilesRequestType = new ProtocolRequestType<
```
viktorsaws, line 19: nit: I think it's not necessary to re-define request types here, since they are exported from runtimes/protocol https://github.com/aws/language-server-runtimes/blob/2231a9bee86eb5c91a909595a17e582dac915d48/runtimes/protocol/identity-management.ts#L106

server/aws-lsp-identity/src/language-server/identityServer.ts
```diff
@@ -0,0 +1,55 @@
...
+} from '@aws/language-server-runtimes/server-interface'
+import { SharedConfigProfileStore } from './profiles/sharedConfigProfileStore'
+import { IdentityService } from './identityService'
+import { CancellationToken } from 'vscode-languageserver'
+
+export const IdentityServerFactory =
+    (identityService: IdentityService, profileService: ProfileService): Server =>
+    (features: {
+        credentialsProvider: CredentialsProvider
+        chat: Chat
```
viktorsaws, line 23: nit: you can skip defining variables for features that are not used, e.g. Chat

server/aws-lsp-identity/src/language-server/identityService.ts
```diff
@@ -0,0 +1,6 @@
+import { SsoTokenChangedParams } from '@aws/language-server-runtimes/protocol'
```
viktorsaws, line 1: nit: `SsoTokenChangedParams` can be imported from 'server-interface'

server/aws-lsp-identity/src/language-server/profiles/sharedConfigProfileStore.ts
```diff
@@ -81,72 +78,83 @@ export class SharedConfigProfileStore implements ProfileStore {
...
 
-                if (!value) {
-                    throw new Error(`Setting [${name}] must have a value.`)
+                // If setting passed with null or undefined then remove setting
+                // If setting passed with any other value then update setting
+                // If setting not passed then preserve setting in file as-is
+                if (value === null || value === undefined) {
+                    Object.hasOwn(parsedSection, name) && delete parsedSection[name]
+                } else {
+                    parsedSection[name] = value.toString()
```
viktorsaws, line 152: When `value` is empty string, it will go to this codepath, is it expected?

server/aws-lsp-identity/src/language-server/profiles/sharedConifgProfileStore.test.ts
viktorsaws, line 1: nit: typo in filename "shared**Conifg**ProfileStore.test.ts"

viktorsaws (approved):

.vscode/launch.json
floralph: The debugger is using these files.  The out/**/*.js.map files alone didn't seem to work for breakpoints with ts-mocha.  After adding those lines the breakpoints started working across both tests and regular .ts files.  I didn't really have time to debug the debugger, so I just left them because being able to set breakpoints is what I needed.

server/aws-lsp-identity/src/language-server/identityService.ts
floralph: Updated, will appear in the next PR.

server/aws-lsp-identity/src/language-server/profiles/sharedConifgProfileStore.test.ts
floralph: Updated, will appear in the next PR.

client/vscode/src/identityActivation.ts
floralph: Actually, it is.  

When attempting to use the types defined in language-server-runtimes an "Unknown parameter structure by name" error.  It's discussed in the links below.  It would seem that there are two different versions of vscode-jsonrpc being imported.  I suspect the one version from @aws/language-server-runtimes and the other from vscode-languageclient/node.  

Looking at the other activation code in that folder, some recreate the types (e.g. credentialsActivation.ts, futureTypes.ts), some add ".method" to the end of the type to call a different overload of sendRequest that doesn't raise the same error (e.g. customizationActivation.ts), and one, chatActivation.ts, uses the type directly imported from language-server-runtimes with the import of LanguageClient as well.  

I wasn't sure how to test chatActivation.ts to see if actually works or results in the same error.  When I removed the ".method" in customziationActivation.ts, the same error occurred, so it's not specific to the identityActivation.ts file.

For now, I'll just add ".method" to the end of the types and update in the next PR.

https://github.com/microsoft/vscode-docker/issues/2601
https://github.com/TypeFox/monaco-languageclient/issues/254
https://github.com/TypeFox/monaco-languageclient/issues/269

server/aws-lsp-identity/src/language-server/profiles/sharedConfigProfileStore.ts
floralph: Good catch.  I wanted to allow falsy values such as 0 and false here, but an empty/whitespace-only string is the same as no value.  Updated here and above on the settings property check to handle these additional no value cases and updated unit tests for these cases.

ivikash (approved):

client/vscode/src/identityActivation.ts
viktorsaws: Got it, thanks for explanation. This is something we could improve later to simplify this integration with VSCode

viktorsaws (approved):

PR #493, opened 2024-10-10
Extend Chat and Completion Telemetry with Customization

ege0zcan:
## Description
Update the chat and completion telemetry to contain customization information. 

For chat we need to be careful to preserve the correct customization associated with a message, this PR introduces logic into the chat telemetry controller 

Tested using the sample extension with IDC login. Verified that customization information is correctly preserved for past messages that were obtained by using customizations and verified that setting a new customization after receiving a response doesn't cause wrong telemetry for current or past chat messages. 

## License


server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
```diff
@@ -85,8 +88,13 @@ export class ChatTelemetryController {
         return tabId && this.#tabTelemetryInfoByTabId[tabId]?.conversationId
     }
 
+    public getCustomizationId(tabId?: string, messageId?: string) {
```
viktorsaws, line 91: nit: I think params should not be optional, since function does nothing otherwise.

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
```diff
@@ -54,13 +55,15 @@ export class ChatTelemetryController {
     #activeTabId?: string
     #tabTelemetryInfoByTabId: { [tabId: string]: ConversationTriggerInfo }
     #currentTriggerByTabId: { [tabId: string]: TriggerType } = {}
+    #customizationInfoByTabAndMessageId: { [tabId: string]: { [messageId: string]: string } }
```
viktorsaws, line 58: Would be using only `messageId` to store this value be enough? If `messageId` is unique uuid across all tabs, then grouping per tab is redundant

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
Jurredr: spotted this too, +1

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
ege0zcan: Very good question, I also considered that but storing by message id means we have no way of knowing when a message is no longer active/relevant. 

When we store messages inside their respective tabs, then we can safely clean up all the messages that are no longer active by deleting the tab entry. So this is to ensure the `customizationInfoByTabAndMessageId` doesn't keep growing infinitely

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -550,10 +553,11 @@ export const CodewhispererServerFactory =
             try {
                 const qConfig = await lsp.workspace.getConfiguration(Q_CONFIGURATION_SECTION)
                 if (qConfig) {
-                    codeWhispererService.customizationArn = undefinedIfEmpty(qConfig.customization)
-                    logging.log(
-                        `Inline completion configuration updated to use ${codeWhispererService.customizationArn}`
-                    )
+                    const customizationValue = undefinedIfEmpty(qConfig.customization)
+                    codeWhispererService.customizationArn = customizationValue
+                    sessionManager.getCurrentSession()?.setCustomizationArn(customizationValue)
```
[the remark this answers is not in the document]
ege0zcan: Given sessions are very short lived I was not able to trigger such a scenario in my testing. After some testing I will actually remove this line so that we get set customization arn on the sessions at creation time only

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
```diff
@@ -168,11 +179,19 @@ export class ChatTelemetryController {
...
+                    codewhispererCustomizationArn: metric.codewhispererCustomizationArn,
                     // not possible: cwsprChatResponseType: metric.cwsprChatResponseType,
                 },
             },
             tabId
         )
+        // Store the customization value associated with the message
+        if (metric.cwsprChatMessageId && metric.codewhispererCustomizationArn) {
+            this.#customizationInfoByTabAndMessageId[tabId] = {
+                ...this.#customizationInfoByTabAndMessageId[tabId],
```
viktorsaws, line 191: nit: destructing assignment is not needed here. If treating object as set, you can just assign new value for `metric.cwsprChatMessageId` in `this.#customizationInfoByTabAndMessageId[tabId]`  directly.

viktorsaws (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
ege0zcan: Are you suggesting to do  `this.#customizationInfoByTabAndMessageId[tabId][metric.cwsprChatMessageId] = metric.codewhispererCustomizationArn` because that doesn't work when `this.#customizationInfoByTabAndMessageId[tabId]` doesn't exist yet

viktorsaws: ah, yeah, either create new array or push to existing one

PR #494, opened 2024-10-11
updating messages in logs for transform

pranav-firake:
## Problem
Updating log messages so that they are more consistent

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
@@ -75,9 +75,9 @@ export class TransformHandler {
 
             const uploadId = await this.preTransformationUploadCode(payloadFilePath)
             const request = getCWStartTransformRequest(userInputrequest, uploadId, this.logging)
-            this.logging.log('send request to start transform api: ' + JSON.stringify(request))
+            this.logging.log('Sending request to start transform api: ' + JSON.stringify(request))
             const response = await this.client.codeModernizerStartCodeTransformation(request)
-            this.logging.log('response start transform api: ' + JSON.stringify(response))
+            this.logging.log('Response from start transform api: ' + JSON.stringify(response))
             return getCWStartTransformResponse(
                 response,
                 uploadId,
@@ -101,7 +101,7 @@ export class TransformHandler {
     async preTransformationUploadCode(payloadFilePath: string): Promise<string> {
         try {
             const uploadId = await this.uploadPayloadAsync(payloadFilePath)
-            this.logging.log('artifact successfully uploaded. upload tracking id: ' + uploadId)
+            this.logging.log('Code artifacts were successfully uploaded. Tracking id for upload is : ' + uploadId)
             return uploadId
         } catch (error) {
             const errorMessage = (error as Error).message ?? 'Failed to upload zip file'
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -308,28 +307,31 @@ export class TransformHandler {
...
                     this.logging.log(
-                        'status is included in validExitSTatus for poll ' +
+                        'Status matches to the desired final status of polling request' +
                             validExitStatus.includes(response.transformationJob.status)
                     )
                 }
 
                 if (validExitStatus.includes(status)) {
-                    this.logging.log('returning status as : ' + status)
+                    this.logging.log('Exiting polling loop with job status as : ' + status)
```
jonlouie, line 327: Reads as a debug log. Can we tighten this? Anytime we exit this polling loop, there should be a log indicating that the work here is complete, whether the job failed, was canceled, or completed successfully. We shouldn't need to explicitly state it unless it has meaning to the customer.

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -338,7 +340,9 @@ export class TransformHandler {
                 getTransformAttempt = 0 // a successful polling will reset attempt
             } catch (e: any) {
                 const errorMessage = (e as Error).message ?? 'Error in GetTransformation API call'
-                this.logging.log('poll : error polling transformation job from the server: ' + errorMessage)
+                this.logging.log(
+                    'polling details : Error in polling transformation job from the server: ' + errorMessage
+                )
```
jonlouie, line 345: We should be more explicit in our error messages and try to use as much standardized/common language as possible.

Example:
"getTransformation request failed: ${errorMessage}"

If this catch block spans more code than a single request, we should still update the language to keep it within the context of the business workflow rather than the logic itself, e.g. "An error occurred while waiting for transformation to complete: ..."

server/aws-lsp-codewhisperer/src/language-server/netTransformServer.ts
```diff
@@ -92,14 +96,17 @@ export const QNetTransformServerToken =
                     }
                     case PollTransformForPlanCommand: {
                         const request = params as GetTransformRequest
-                        logging.log('Calling pollTransformForPlan request with job Id: ' + request.TransformationJobId)
+                        logging.log(
+                            'Calling pollTransformForPlan request with Transformation Job ID: ' +
+                                request.TransformationJobId
+                        )
```
jonlouie, line 102: Taking an opinionated stance: we should only mention names of API endpoints in these logs. Naming an internal function will: 
1. Not have meaning for the customer
2. Cause confusion for ourselves (from just the logs, how will we know if we're talking about an actual endpoint or some function? Did we mean to reference an actual endpoint and it was misspelled?)

Suggestion: 
"Awaiting  transformationPlan for Transformation Job ID: ${request.TransformationJobId}."

jonlouie (changes requested):

jonlouie:
Spoke offline with @pranav-firake and we will take a deeper look at the log messages themselves at another time.

huawenm (approved):

PR #500, opened 2024-10-15
Add cache for client registration and sso token

floralph:
Adds cache for client registration and sso token files stored in $USER/.aws/sso/cache.  Leverages existing smithy-type and AWS SDK for JavaScript code where possible.

ssoCache.ts contains the cache interface along with a few supporting classes (SSOToken is reused from smithy-types, SsoClientRegistration added here in similar manner).  

FileSystemSsoCache is implementation for standalone runtime.  If an when webworker runtime needs an SsoCache, the interface can be implemented for it at that time (e.g. BrowserStorageSsoCache).

The cache is very basic, it just gets and sets the sso tokens and client registrations.  As it is a cache, it tends to swallow many errors so as to just allow the object to be recreated and not break the user experience.  For sso tokens, they will expire each hour, but it's better that they work at all than fail.  

Logging throughout the identity server will be added in a future PR (IDE-14689).  This will notify users of cache issues without interrupting their work.

### DIFF
--- a/server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
+++ b/server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
@@ -1,0 +1,107 @@
+import { getSSOTokenFilepath, getSSOTokenFromFile, SSOToken } from '@smithy/shared-ini-file-loader'
+import { SsoCache, SsoClientRegistration, ssoClientRegistrationDuckTyper, ssoTokenDuckTyper } from './ssoCache'
+import { readFile, writeFile } from 'fs/promises'
+import { AwsError, tryAsync } from '../../awsError'
+import { AwsErrorCodes } from '@aws/language-server-runtimes/protocol'
+
+// As this is a cache, we tend to swallow a lot of the errors as it is ok to just recreate the items each
+// time, though recreating the SSO token without a refresh token will be every hour.  This is a better
+// situation that failing to get an SSO token at all.
+
+export class FileSystemSsoCache implements SsoCache {
+    async getSsoClientRegistration(
+        clientName: string,
+        ssoRegion: string,
+        ssoStartUrl: string
+    ): Promise<SsoClientRegistration | undefined> {
+        const id = toSsoClientRegistrationCacheId(clientName, ssoRegion, ssoStartUrl)
+        return await tryGetAsync(async () => {
+            const registration = await getSsoClientRegistrationFromFile(id)
+            return ssoClientRegistrationDuckTyper.eval(registration) ? registration : undefined
+        })
+    }
+
+    async setSsoClientRegistration(
+        clientName: string,
+        ssoRegion: string,
+        ssoStartUrl: string,
+        clientRegistration: SsoClientRegistration
+    ): Promise<void> {
+        if (!clientName || !ssoRegion || !ssoStartUrl || !ssoClientRegistrationDuckTyper.eval(clientRegistration)) {
+            return
+        }
+
+        const id = toSsoClientRegistrationCacheId(clientName, ssoRegion, ssoStartUrl)
+        await tryAsync(
+            () => writeSsoObjectToFile(id, clientRegistration),
+            error => AwsError.wrap(error, AwsErrorCodes.E_CANNOT_WRITE_SSO_CACHE)
+        )
+    }
+
+    async getSsoToken(ssoSessionName: string): Promise<SSOToken | undefined> {
+        return await tryGetAsync(async () => {
+            const ssoToken = await getSSOTokenFromFile(ssoSessionName)
+            return ssoTokenDuckTyper.eval(ssoToken) ? ssoToken : undefined
+        })
+    }
+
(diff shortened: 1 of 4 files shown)


server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
```diff
@@ -0,0 +1,107 @@
...
+
+// Based on:
+// https://github.com/aws/aws-sdk-js-v3/blob/6e61f0e78ff7a9e3b1f2cd651bde5fc656d85ba9/packages/token-providers/src/writeSSOTokenToFile.ts
+async function writeSsoObjectToFile(id: string, ssoObject: SSOToken | SsoClientRegistration): Promise<void> {
+    try {
+        const filepath = getSSOTokenFilepath(id)
+        const json = JSON.stringify(ssoObject, null, 2)
+        return await writeFile(filepath, json)
+    } catch {
+        // Writing to cache is best effort, based on:
```
volodkevych, line 98: Do we at least want to log something?

server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
```diff
@@ -0,0 +1,107 @@
...
+        startUrl: ssoStartUrl,
+        tool: clientName,
+    })
+}
+
+// Based on:
+// https://github.com/smithy-lang/smithy-typescript/blob/main/packages/shared-ini-file-loader/src/getSSOTokenFromFile.ts
+async function getSsoClientRegistrationFromFile(id: string): Promise<SsoClientRegistration> {
+    const filepath = getSSOTokenFilepath(id)
+    const json = await readFile(filepath, 'utf8')
```
volodkevych, line 86: No need to catch possible errors here?

server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
```diff
@@ -0,0 +1,107 @@
...
+import { SsoCache, SsoClientRegistration, ssoClientRegistrationDuckTyper, ssoTokenDuckTyper } from './ssoCache'
+import { readFile, writeFile } from 'fs/promises'
+import { AwsError, tryAsync } from '../../awsError'
+import { AwsErrorCodes } from '@aws/language-server-runtimes/protocol'
+
+// As this is a cache, we tend to swallow a lot of the errors as it is ok to just recreate the items each
+// time, though recreating the SSO token without a refresh token will be every hour.  This is a better
+// situation that failing to get an SSO token at all.
+
+export class FileSystemSsoCache implements SsoCache {
```
viktorsaws, line 11: When cache will be used in server, is there a possibility of conflict of any kind when 2 or more IDE instances are running and will read/write data to cache? E.g. can 2 instances constantly rewrite and refresh each other tokens?
Or `clientName` and `ssoSessionName` will be unique per IDE running instance, so this is not an issue?

server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
floralph: Yes, that is addressed at the bottom of the PR comments above "Logging throughout the identity server will be added in a future PR (IDE-14689). This will notify users of cache issues without interrupting their work." 

There is a task(s) to add logging (w/o PII) throughout the server, but to support the deadlines in the related project, I have to get GetSsoToken out first.  

There will additionally be an update to the cache to remove token files for the InvalidateSsoToken, but that is coming as planned/scheduled near future work as well.

server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
floralph: No, this is just a module private function to help keep things similar to smithy-typescript.  The called getSsoClientRegistration handles any errors thrown.  This function would not be exposed beyond this module.

server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
floralph: Yes, this was discussed last week.  This is an open issue to resolve, which may be at the cache level or maybe at a higher level.  It is not addressable at this time though.  It is a corner case that can be addressed before the first destination usage.

server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
volodkevych: IMO, adding logging right away should not be a big effort. Troubleshooting server without logs can take more time than adding logging in process of implementation at the end.

server/aws-lsp-identity/src/sso/cache/fileSystemSsoCache.ts
viktorsaws: Got it, thanks for explanation. No concerns since we're tracking this issue

viktorsaws (approved):

ivikash (approved):

PR #505, opened 2024-10-21
Amazon Q LS: Add makeUserContext utility to create UserContext payload for sendTelemetry event

viktorsaws:
## Problem
To start using Codewhisperer `sendTelemetryEvent` API we need to attach UserContext object to each event using https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/client/token/bearer-token-service.json#L1568-L1578 interface.

## Solution

* Created new `makeUserContextObject` utility function to use data passed in `InitializationParams` to compute UserContext object https://github.com/aws/language-server-runtimes/blob/6d40b171e3564e37ffe7c9d3bc8ea71db1e3aab6/runtimes/protocol/lsp.ts#L49-L81. New function is in `server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts`
* Choir: Move `getUserAgent` function to `utilities/telemetryUtils.ts`

When computing `ideCategory` in `makeUserContextObject`, I chose to use `InitializeParams.initializationOptions.aws.clientInfo.extension` Runtimes protocol extension only https://github.com/aws/language-servers/pull/505/files#diff-493b137e67edd04749e7b8e457784304758d37ca3222843a15c5581be7b34e3eR62. The reason is that we expect this field for first party integrators, that we control and can treat them as most predictable values. Other fields, like `InitializeParams.clientInfo` are controlled by destination IDE, and can change at any point in time. In the future we can extend the logic to use other fields from InitializeParams.

Alternative solution is to introduce new `IdeCategory` field on Runtime protocol level, and recommend predefined enum of acceptable values (similar to how LSP defines `LanguageId`). This require protocol extension and adds redundancy to `InitializeParams.initializationOptions.aws.clientInfo.extension` field and all integrating destinations to pass extra field at handshake. This solution can be cleaner though and does not couple to custom extension names. I am open to change this approach, but we need to align with all integrating Toolkits on passing this value.

## License


server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
```diff
@@ -0,0 +1,103 @@
+import { InitializeParams, ServerInfo } from '@aws/language-server-runtimes/server-interface'
+import { IdeCategory, UserContext } from '../../client/token/codewhispererbearertokenclient'
+
+const USER_AGENT_PREFIX = 'AWS-Language-Servers'
+
+export const getUserAgent = (initializeParams: InitializeParams, serverInfo?: ServerInfo): string => {
```
viktorsaws, line 6: `getUserAgent` is unchanged and only moved to new location.

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (initializeParams.initializationOptions?.aws.clientInfo?.extension?.name) {
+        ideCategory = mapClientNameToIdeCategory(initializeParams.initializationOptions.aws.clientInfo.extension.name)
+    }
+
+    return ideCategory || 'UNKNOWN'
+}
+
+// Map result from https://nodejs.org/api/process.html#process_process_platform to expected Operating system
+const getOperatingSystem = () => {
+    switch (process.platform) {
```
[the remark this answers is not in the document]
viktorsaws: When compiled with webpack, `process` is replaced with browser-compatible version. I didn't check what value is in that case, but we can extend this code when that's needed.

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
```diff
@@ -0,0 +1,103 @@
...
+        const { clientInfo } = initializeParams || {}
+        if (clientInfo) {
+            items.push(`${clientInfo.name ? format(clientInfo.name) : 'UNKNOWN'}/${clientInfo.version || 'UNKNOWN'}`)
+        }
+    }
+
+    return items.join(' ')
+}
+
+const IDE_CATEGORY_MAP: { [key: string]: IdeCategory } = {
```
viktorsaws, line 49: I'm looking for feedback on using custom extension name to get `IdeCategory`. 
Alternatives are:
1. Use `InitializeParams.clientInfo` set by IDE - since this value is outside of our control, we will need to maintain ever growing list of different variants to keep mapping up to date
2. Introduce new `IdeCategory` field in our custom `InitializeParams.initializationOptions.aws.clientInfo` and document enum of supported values.

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+            return 'UNKNOWN'
+    }
+}
+
+export const makeUserContextObject = (initializeParams: InitializeParams, product: string): UserContext | undefined => {
+    const userContext: UserContext = {
+        ideCategory: getIdeCategory(initializeParams),
+        operatingSystem: getOperatingSystem(),
+        product: product,
+        clientId: initializeParams?.initializationOptions?.aws?.clientInfo?.clientId,
```
volodkevych, line 93: nit: `aws` cannot be null based on interface

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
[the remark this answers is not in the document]
volodkevych: I think we do not need to report OS for web, we need to ensure it does not fail.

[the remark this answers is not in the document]
viktorsaws: when platform is empty, we will not add whole UserContext object to telemetry. But I'll check for cross-platform alternative here

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+        ideVersion:
+            initializeParams?.initializationOptions?.aws?.clientInfo?.version || initializeParams.clientInfo?.version,
+    }
+
+    if (userContext.ideCategory === 'UNKNOWN' || userContext.operatingSystem === 'UNKNOWN') {
+        return
```
volodkevych, line 104: `SendTelemetryEvent` won't work for web due to auth support. We'll need to make sure we report CW business telemetry for standalone only.

volodkevych: Then, we need to log error, if `ideCategory` or `operatingSystem` is undefined.

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
```diff
@@ -0,0 +1,108 @@
...
+        ideCategory = mapClientNameToIdeCategory(initializeParams.initializationOptions.aws.clientInfo.extension.name)
+    }
+
+    return ideCategory || 'UNKNOWN'
+}
+
+// Map result from https://nodejs.org/api/process.html#process_process_platform to expected Operating system
+const getOperatingSystem = () => {
+    // Node.js environment
+    if (typeof process !== 'undefined') {
```
volodkevych, line 74: Would injecting this from runtime as part of `Runtime` be an option?

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
volodkevych: Both options not perfect:
1. We'll need to maintain the list, does not scale
2. IDEs need to inject CW-specific representation of IDE name, which should ideally be abstracted away
Do you think having both would make sense?

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Compute UserContext object for sendTelemetryEvent API call.
+// Do not return context when unknown IDE or Operating system is found.
+// This behaviour may change in the future, when API will accept not enumerated values in API definition.
+export const makeUserContextObject = (initializeParams: InitializeParams, product: string): UserContext | undefined => {
+    const userContext: UserContext = {
+        ideCategory: getIdeCategory(initializeParams),
+        operatingSystem: getOperatingSystem(),
+        product: product,
```
pras0131, line 97: Which field should be populated by client/destination for the product field? and what can be the possible values of product?

pras0131: or should it be like from codeWhispererServer, we pass INLINE but for QServer, we pass CHAT?

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
volodkevych: Why are we mapping from `extension.name`, not `name`?

viktorsaws: `name` is meant to set name of IDE (or client destination), which is larger, more incoherent list, than `extension.name`. And typically it will be set from some hardcoded value set by IDE itself, which means that we (AWS) don't control this list.

`extension.name` is fully controlled by Toolkits and signifies toolkit extension name - smaller list, which we control.

viktorsaws: > I'd probably stick with 1, as I would not expect changes to happen often. And the values are anyway hardcoded on CW side.

Which option do you mean: the one I implemented to use `aws.clientInfo.extension.name` or `1. Use InitializeParams.clientInfo set by IDE`?

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
viktorsaws: I leave this logic to be implemented in TelemetryService that we are adding in https://github.com/aws/language-servers/pull/524. This PR just adds utility helper to compute this object

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
viktorsaws: `product` will be set by every Server, which use TelemetryService, and it means specific Amazon Q Feature emitting telemetry (e.g. INLINE, CHAT).

Rest of fields are computed based on data set by Client following protocol.

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
viktorsaws: that's an idea, we can make it part of Runtime feature we pass https://github.com/aws/language-server-runtimes/blob/main/runtimes/server-interface/runtime.ts

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
volodkevych: Sorry, I meant the one you implemented. Ideally, CW should modify their model to make IDE category optional.

viktorsaws: > Ideally, CW should modify their model to make IDE category optional.

yeah, I discussed it internally with them, this might be done in future

pras0131 (approved):

viktorsaws:
Builds are blocked due to backwards incompatible changes in identityManagement protocol, until https://github.com/aws/language-servers/pull/531 is merged

volodkevych (approved):

PR #507, opened 2024-10-22
Ignore streaming client from dependabot PRs

rahmaniaam:
## Problem
The streaming client is not to be manually updated, this means any dependabot PRs pointing to that directory should be ignored. Currently we still get those PRs

## Solution
There currently is no option to exclude the entire subdirectory from the scan, but there is a workaround to autoclose the PRs that only targets the streaming client https://github.com/dependabot/dependabot-core/issues/4364#issuecomment-2006542987. 

## License

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,48 @@
 version: 2
 updates:
     - package-ecosystem: 'npm'
-      directory: '/'
-      target-branch: 'main' # Avoid updates to "staging".
+      directory: '/server/aws-lsp-buildspec'
+      target-branch: 'main'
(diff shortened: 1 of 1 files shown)


ege0zcan:
Had a discussion offline, this solution can potentially be flaky because usually dependabot PRs don't only target the streaming-client so this change wouldn't solve the problem. We are thinking to work around this limitation by ensuring dependabot doesn't group dependency upgrades, meaning 1 dependabot PR targets a single subpackage and doesn't group multiple subpackages

volodkevych (approved):

ege0zcan (approved):

PR #524, opened 2024-10-22
Telemetry service with common components encapsulated

pras0131:
## Problem
Implement a common telemetry client/service which acts as a base to emit other events (for inline and chat as of now)
Reference ticket: https://issues.amazon.com/issues/AWS-Cloud9-28670

## Solution
The PR aims to create a common service named `telemetryService` which will be responsible for holding the up-to date state of common pieces like UserContext, OptOutPreference, EnableTelemetryEventsToDestination. The `TelemetryService` has methods to update those states. 
TelemetryService has been initialized in both the CodeWhispererServer and QServer classes & will be passed/injected to the respective classes, based on when event emission is required. 
The PR also contains the updates to the flags like optOutTelemetry and enableTelemetryEventsToDestination whenever there is a update in the configuration and these flags will be pulled from aws.q configuration. 
It contains sample integration with UserTriggerDecisionEvent as well, although some fields of the event which still needs to be populated will be done separately.

## License


server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,83 @@
...
+    private userContext!: UserContext
+    private optOutPreference!: OptOutPreference
+    private enableTelemetryEventsToDestination!: boolean
+    private toolkitTelemetry!: Telemetry
+
+    constructor(credentialsProvider: CredentialsProvider, additionalAwsConfig: AWS.ConfigurationOptions = {}) {
+        super(credentialsProvider, additionalAwsConfig)
+    }
+
+    public setToolkitTelemetry(toolkitTelemetry: Telemetry): void {
```
viktorsaws, line 22: I'm not sure for the purpose of this method from the name, it is needed at all, you can pass `telemetry` object in constructor. `UserContext` can be passed in constructor as well.

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,83 @@
...
+            generatedLine: 0,
+            numberOfRecommendations: 0
+             */
+        }
+        this.sendTelemetryEvent({
+            telemetryEvent: {
+                userTriggerDecisionEvent: event,
+            },
+            userContext: this.userContext ? this.userContext : undefined,
+            optOutPreference: this.optOutPreference ? this.optOutPreference : undefined,
```
viktorsaws, line 80: For `optOutPreference` value, are we expecting Toolkit to send us a string value or boolean? I'd suggest to send boolean, and then we convert it to actual value SendTelemetryEvent expects

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: I can pass telemetry in the constructor but userContext for let's say QServer is not defined within the same file but ChatController instead.

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public updateEnableTelemetryEventsToDestination(enableTelemetryEventsToDestination: boolean): void {
+        this.enableTelemetryEventsToDestination = enableTelemetryEventsToDestination
+    }
+
+    private getSuggestionState(session: CodeWhispererSession): SuggestionState {
+        let suggestionState: SuggestionState
+        if (session.getAggregatedUserTriggerDecision() === undefined) {
+            suggestionState = ''
```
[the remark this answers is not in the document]
pras0131: will make the changes

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,83 @@
...
+            /* TODO: populate correct values for generatedLine and numberOfRecommendations
+            generatedLine: 0,
+            numberOfRecommendations: 0
+             */
+        }
+        this.sendTelemetryEvent({
+            telemetryEvent: {
+                userTriggerDecisionEvent: event,
+            },
+            userContext: this.userContext ? this.userContext : undefined,
```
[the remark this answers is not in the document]
pras0131: you mean if they are undefined, then not even include the userContext key?

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: Toolkit sends a boolean only, we convert from boolean to string.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -289,12 +294,14 @@ export const CodewhispererServerFactory =
 
         const sessionManager = SessionManager.getInstance()
         const codeWhispererService = service(credentialsProvider)
+        const telemetryService = new TelemetryService(credentialsProvider)
+        telemetryService.setToolkitTelemetry(telemetry)
 
         lsp.addInitializer((params: InitializeParams) => {
             codeWhispererService.updateClientConfig({
```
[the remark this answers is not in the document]
pras0131: just wondering why would we need userAgent in telemetryService?

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,83 @@
...
+    private optOutPreference!: OptOutPreference
+    private enableTelemetryEventsToDestination!: boolean
+    private toolkitTelemetry!: Telemetry
+
+    constructor(credentialsProvider: CredentialsProvider, additionalAwsConfig: AWS.ConfigurationOptions = {}) {
+        super(credentialsProvider, additionalAwsConfig)
+    }
+
+    public setToolkitTelemetry(toolkitTelemetry: Telemetry): void {
+        this.toolkitTelemetry = toolkitTelemetry
```
volodkevych, line 23: I do not think we should name it as a `**toolkit**Telemetry`.

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,83 @@
...
+            recommendationLatencyMilliseconds: 0,
+            timestamp: new Date(Date.now()),
+            triggerToResponseLatencyMilliseconds: session.timeToFirstRecommendation,
+            suggestionReferenceCount: session.suggestions.length,
+            /* TODO: populate correct values for generatedLine and numberOfRecommendations
+            generatedLine: 0,
+            numberOfRecommendations: 0
+             */
+        }
+        this.sendTelemetryEvent({
```
volodkevych, line 75: It will not work for `CodeWhispererServerIAM` due to unsupported auth.

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: clientTelemetry?

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: do we need to make it work for **CodeWhispererServerIAM** as well though? Is there a destination/client consuming [link](https://github.com/aws/language-servers/blob/main/app/aws-lsp-codewhisperer-runtimes/src/iam-standalone.ts) ?

pras0131: For CodeWhispererServerIAM, we use CodeWhispererSigv4Client client and the class/interface for the client methods only has two methods: generateRecommendations of two variants, so it should be okay

volodkevych: `sendTelemetryEvent` in CW client did not support IAM when I last looked at it.

pras0131: yes when I create another PR for remaining changes of UserTriggerDecision event, i'll add the change to not emit for IAM.

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: discussed this with Viktor to pass just as telemetry and in the constructor

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -559,6 +583,10 @@ export const CodewhispererServerFactory =
                     logging.log(
                         `Inline completion configuration updated to use ${codeWhispererService.customizationArn}`
                     )
+                    const enableTelemetryEventsToDestination = qConfig['enableTelemetryEventsToDestination'] == true
```
viktorsaws, line 586: nit: better ot use strict comparison `===` where we expect booleans

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
```diff
@@ -2,20 +2,22 @@ import { CredentialsProvider, InitializeParams, Server } from '@aws/language-ser
...
         const chatSessionManagementService: ChatSessionManagementService = service(credentialsProvider)
+        const telemetryService = new TelemetryService(credentialsProvider, {}, telemetry)
 
-        const chatController = new ChatController(chatSessionManagementService, features)
+        const chatController = new ChatController(chatSessionManagementService, features, telemetryService)
 
         lsp.addInitializer((params: InitializeParams) => {
             chatSessionManagementService.setCustomUserAgent(getUserAgent(params, runtime.serverInfo))
-
+            // TODO: set the user context here for TelemetryService
```
viktorsaws, line 20: We need to set `customUserAgent` for `TelemetryService` here as well

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export class TelemetryService extends CodeWhispererServiceToken {
+    private userContext!: UserContext
+    private optOutPreference!: OptOutPreference
+    private enableTelemetryEventsToDestination!: boolean
+    private telemetry: Telemetry
+
+    constructor(
+        credentialsProvider: CredentialsProvider,
+        additionalAwsConfig: AWS.ConfigurationOptions = {},
```
viktorsaws, line 20: Better to move optional params to the end of the list

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,97 @@
...
+        this.optOutPreference = optOutPreference
+    }
+
+    public updateEnableTelemetryEventsToDestination(enableTelemetryEventsToDestination: boolean): void {
+        this.enableTelemetryEventsToDestination = enableTelemetryEventsToDestination
+    }
+
+    private getSuggestionState(session: CodeWhispererSession): SuggestionState {
+        let suggestionState: SuggestionState
+        if (session.getAggregatedUserTriggerDecision() === undefined) {
```
viktorsaws, line 41: (super nit pick and personal choice) why not using `switch-case` here? Makes code much easier to read

viktorsaws:
I'd suggest to merge this PR only when whole logic for handling optOut flags and credential modes are implemented in https://github.com/aws/language-servers/pull/542. 
I'd avoid introducing half-finished implementation to main, since it might be accidentally released to prod

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: you are right, switch-case looks a bit cleaner, will update

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,122 @@
...
+        if (this.optOutPreference !== undefined) {
+            request.optOutPreference = this.optOutPreference
+        }
+        this.sendTelemetryEvent(request)
+    }
+
+    public emitUserTriggerDecision(session: CodeWhispererSession, timeSinceLastUserModification?: number) {
+        if (this.isLoginInvalidForTelemetry()) {
+            return
+        }
```
viktorsaws, line 86: I think this condition should be moved to `invokeSendTelemetryEvent`, as it is needed for other events

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,122 @@
...
+
+        const event: UserTriggerDecisionEvent = {
+            sessionId: session.codewhispererSessionId || '',
+            requestId: session.responseContext?.requestId || '',
+            customizationArn: session.customizationArn === '' ? undefined : session.customizationArn,
+            programmingLanguage: {
+                languageName: getRuntimeLanguage(session.language),
+            },
+            completionType:
+                session.suggestions.length > 0 ? getCompletionType(session.suggestions[0]).toUpperCase() : '',
```
viktorsaws, line 108: Looks like completionType can't be empty string https://github.com/aws/language-servers/blob/4aee4d35769d1bfc79fbf3b3b8e5b7a3330a327a/server/aws-lsp-codewhisperer/src/client/token/bearer-token-service.json#L1090-L1093

VSCode defaults to `LINE`, guess we should do the same

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -0,0 +1,122 @@
...
+            suggestionState: this.getSuggestionState(session),
+            recommendationLatencyMilliseconds: session.firstCompletionDisplayLatency
+                ? session.firstCompletionDisplayLatency
+                : 0,
+            timestamp: new Date(Date.now()),
+            triggerToResponseLatencyMilliseconds: session.timeToFirstRecommendation,
+            suggestionReferenceCount: referenceCount,
+            generatedLine: generatedLines,
+            numberOfRecommendations: session.suggestions.length,
+            perceivedLatencyMilliseconds: timeSinceLastUserModification,
```
viktorsaws, line 118: Looking in VSCode `perceivedLatencyMilliseconds`, is computed based on triggerType https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/util/codeWhispererSession.ts#L86-L92

We need to add case for manual trigger and use `timeToFirstRecommendation` here

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -169,17 +171,20 @@ const emitUserTriggerDecisionTelemetry = (
...
 }
 
 const emitAggregatedUserTriggerDecisionTelemetry = (
     telemetry: Telemetry,
+    telemetryService: TelemetryService,
     session: CodeWhispererSession,
     timeSinceLastUserModification?: number
 ) => {
+    telemetryService.emitUserTriggerDecision(session, timeSinceLastUserModification)
+    // TODO: the below emitted event to the Toolkit DWH will be moved to telemetryService as well.
```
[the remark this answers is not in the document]
pras0131: we have it already

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: I can do this in a follow up PR, as I added test cases basis if invokeSendTelemetryEvent should be caled or not

viktorsaws (approved):

PR #535, opened 2024-10-23
Handle insertToCursorPosition event in the chat server

ege0zcan:
## Description

Currently the chat server is not capable of handling insertToCursor position event. With the recent addition of `applyWorkspaceEdit` functionality to the runtimes (https://github.com/aws/language-server-runtimes/pull/233) the server becomes capable of handling the event. This PR enables the chat server to handle `insertToCursorPosition` event

This PR accommodates for existing usages of the chat server and won't cause breaking changes for existing destinations

In addition: 
- Updated the sample extension to test the server handling the insert event 
- Minor: Updated `cSharpDependencyGraph` test file to prevent the tests from breaking whenever there are changes to the runtimes package
- Minor: Also updated the tasks.json file: removed unused task and removed the label from compile task because it was breaking the launch.json tasks 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
@@ -2,7 +2,15 @@ import {
     GenerateAssistantResponseCommandInput,
     GenerateAssistantResponseCommandOutput,
 } from '@amzn/codewhisperer-streaming'
-import { ErrorCodes, FeedbackParams, chatRequestType } from '@aws/language-server-runtimes/protocol'
+import {
+    ApplyWorkspaceEditParams,
+    ErrorCodes,
+    FeedbackParams,
+    InsertToCursorPositionParams,
+    TextDocumentEdit,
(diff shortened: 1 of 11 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -169,7 +177,63 @@ export class ChatController implements ChatHandlers {
...
+        if (indent && indent.trim().length !== 0) {
+            indent = ' '.repeat(indent.length - indent.trimStart().length)
+        }
+
+        let textWithIndent = ''
+        params.code.split('\n').forEach((line, index) => {
+            if (index === 0) {
+                textWithIndent += line
+            } else {
+                textWithIndent += '\n' + indent + line
```
rahmaniaam, line 213: I remember there is a difference to newlines in visual studios/windows, cmiiw but it should only affect when reading from the file rather and not when writing to it right?

rahmaniaam (approved):

PR #540, opened 2024-10-25
Implement auto-refresher

floralph:
NOTE:  This PR builds on https://github.com/aws/language-servers/pull/539 and pends that one being merged.  The unit tests are still outstanding and I will work on those in the morning.  Posting this PR to allow for an early review.  The code was verified to function as expected using the vscode sample client.

The SsoTokenAutoRefresher class relies on the RefreshingSsoCache to handle all the refresh steps.  The auto-refresher just attempts to fetch the token via a RefreshingSsoCache and schedules the next time to attempt another fetch to force a refresh.  Auto-refresh is best effort, so errors are generally swallowed.  Auto-refresh is automatic when fetching a token via GetSsoToken and is only terminated if the token is no longer cached (never existed, expired, maybe someone else invalidated it) or is invalidated via InvalidateSsoToken.

Unit tests have been declared, but must be written.  

SsoTokenChanged.Refreshed event is raised from RefreshingSsoCache when a successful refresh attempt is made.

### DIFF
--- a/server/aws-lsp-identity/src/language-server/ssoTokenAutoRefresher.ts
+++ b/server/aws-lsp-identity/src/language-server/ssoTokenAutoRefresher.ts
@@ -1,26 +1,65 @@
-import { SsoTokenChangedParams } from '@aws/language-server-runtimes/protocol'
-import { RefreshingSsoCache } from '../sso/cache/refreshingSsoCache'
+import { SsoSession, SsoTokenChangedParams } from '@aws/language-server-runtimes/protocol'
+import { RefreshingSsoCache, refreshWindowMillis, retryWindowMillis } from '../sso/cache/refreshingSsoCache'
+import { throwOnInvalidClientName, throwOnInvalidSsoSession, throwOnInvalidSsoSessionName } from '../sso/utils'
+
+const bufferedRefreshWindowMillis = refreshWindowMillis * 0.95
+const bufferedRetryWindowMillis = retryWindowMillis * 1.05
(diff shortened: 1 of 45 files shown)


imykhai (approved):

server/aws-lsp-identity/src/language-server/ssoTokenAutoRefresher.ts
```diff
@@ -1,26 +1,65 @@
...
+        // Token doesn't exist, was invalidated by another process, or has expired
+        if (!ssoToken) {
+            return
+        }
+
+        const nowMillis = Date.now()
+        const expiresAtMillis = Date.parse(ssoToken.expiresAt)
+        let delayMillis: number
+
+        if (nowMillis < expiresAtMillis - refreshWindowMillis) {
```
ivikash, line 40: 1. Should we create a common utility for jitters?
2. `Math.random()` will ensure the jitter is between 0 to ~`1m66sec`. Is that enough?

ivikash (approved):

server/aws-lsp-identity/src/language-server/ssoTokenAutoRefresher.ts
floralph: Jitter can be a bit more variable whether it occur +/- or just one one side and the timing is dependent on the application.  For now, I'm going to leave it in-line, but if I find it starts getting reused more often, I will refactor it into a more reusable form.

PR #542, opened 2024-10-25
User trigger decision event

pras0131:
## Problem
Remaining changes for the UserTriggerDecision event.

## Solution
The PR has changes to make sure we don't emit the metric when the login is via IAM. 
There is a condition to make sure metric is emitted in case of identityCenter even in case optOut is false (to make it consistent with VSCode implementation)
It has a utility method to identify the type of login. Test cases have been added for the respective emitUTDE function.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
@@ -7,13 +7,14 @@ import {
     UserContext,
     OptOutPreference,
 } from '../client/token/codewhispererbearertokenclient'
-import { getCompletionType } from './utils'
+import { getCompletionType, getLoginTypeFromProvider, LoginType } from './utils'
 
 export class TelemetryService extends CodeWhispererServiceToken {
     private userContext!: UserContext
     private optOutPreference!: OptOutPreference
     private enableTelemetryEventsToDestination!: boolean
     private telemetry: Telemetry
+    private loginType: LoginType
 
     constructor(
         credentialsProvider: CredentialsProvider,
@@ -22,6 +23,7 @@ export class TelemetryService extends CodeWhispererServiceToken {
     ) {
         super(credentialsProvider, additionalAwsConfig)
         this.telemetry = telemetry
+        this.loginType = getLoginTypeFromProvider(credentialsProvider)
     }
 
     public updateUserContext(userContext: UserContext): void {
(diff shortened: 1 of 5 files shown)


server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -52,7 +54,24 @@ export class TelemetryService extends CodeWhispererServiceToken {
         return suggestionState
     }
 
+    private isLoginInvalidForTelemetry(): boolean {
+        return this.loginType === 'iam' || (this.loginType !== 'identityCenter' && this.optOutPreference !== 'OPTOUT')
+    }
+
+    private invokeSendTelemetryEvent(event: any) {
```
viktorsaws, line 61: nit: can we specify type instead of `any` for `event`?

server/aws-lsp-codewhisperer/src/language-server/utils.ts
```diff
@@ -131,3 +133,15 @@ export const getUserAgent = (initializeParams: InitializeParams, serverInfo?: Se
 
     return items.join(' ')
 }
+
+export function getLoginTypeFromProvider(credentialsProvider: CredentialsProvider): LoginType {
+    if (credentialsProvider.hasCredentials('iam')) {
```
viktorsaws, line 138: Unfortunately, we cant rely on presence of `iam` type to decide which login type - `CredentialProvider allows both types to exist simultaneously, which might result in bug when client does not clean not used type correctly.

Instead, I think we should rely on the specific CodewhispererService implementation class used in server.

server/aws-lsp-codewhisperer/src/language-server/utils.ts
```diff
@@ -131,3 +133,15 @@ export const getUserAgent = (initializeParams: InitializeParams, serverInfo?: Se
...
+        return 'iam'
+    }
+    const connectionMetadata = credentialsProvider.getConnectionMetadata()
+    return connectionMetadata && connectionMetadata.sso && connectionMetadata.sso.startUrl
```
viktorsaws, line 142: nit: this conditions chain can be replaced with optional chaining `connectionMetadata?.sso?.startUrl?.includes(builderIdStartUrl)`

server/aws-lsp-codewhisperer/src/language-server/utils.ts
```diff
@@ -131,3 +133,15 @@ export const getUserAgent = (initializeParams: InitializeParams, serverInfo?: Se
...
+        ? connectionMetadata.sso.startUrl.includes(builderIdStartUrl)
+            ? 'builderId'
+            : 'identityCenter'
+        : 'builderId'
```
viktorsaws, line 146: Why this falls back to `builderId`, is this fallback scenario in VSCode as well?

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: I added any intentionally because this is like a common method I intend to use for all the events. Once, all are implemented - I'll change it to Event1|Event2|.. something like this.

server/aws-lsp-codewhisperer/src/language-server/utils.ts
pras0131: yes..if I see this code; they have a comment that it must be IDC by the process of elimination..in vscode they try to check using some connection parameters but here we have builderId URL defined but idc start urls can be different - that's why this condition.
`    public isEnterpriseSsoInUse(): boolean {
        const conn = this.conn
        // we have an sso that isn't builder id, must be IdC by process of elimination
        const isUsingEnterpriseSso = conn?.type === 'sso' && !isBuilderIdConnection(conn)
        return conn !== undefined && isUsingEnterpriseSso
    }`

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -52,7 +54,24 @@ export class TelemetryService extends CodeWhispererServiceToken {
...
+            optOutPreference: this.optOutPreference,
+        })
+    }
+
     public emitUserTriggerDecision(session: CodeWhispererSession) {
+        if (this.isLoginInvalidForTelemetry()) {
+            return
+        }
         const completionSessionResult = session.completionSessionResult ?? {}
         const acceptedItemId = Object.keys(completionSessionResult).find(k => completionSessionResult[k].accepted)
```
viktorsaws, line 76: you can simplify finding accepted item: ID is stored in `Session.acceptedSuggestionId` value

PR #550, opened 2024-10-28
Amazon Q: Emit CodeCoverageEvent to CodeWhisperer.SendTelemetryEvent API

viktorsaws:
## Problem

We need to emit CodeCoverageEvent telemetry to Amazon Q SendTelemetry API.

## Solution

* Added new `emitCodeCoverageEvent` method to `TeleetryService`
* Extended `CodePercentageTracker` to track customization used and call `emitCodeCoverageEvent` at scheduled time
* Extended unit tests to cover new methods
* Bug fix: in existing telemetry sent over LSP channel, we used `codewhispererAcceptedTokens` to send length of all accepted recommendations. Per reference implementation in VSCode, we should use `codewhispererSuggestedTokens` for this. Switched to using `codewhispererSuggestedTokens` for this event

Reference implementation https://github.com/aws/aws-toolkit-vscode/blob/6c420c7f8c2f03fc1c6236669d701b2581cba6e0/packages/core/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts#L132-L146.

Difference with reference implementation is that we don't send `unmodifiedAcceptedCharacterCount` field. This field is optional, deprecated and will be removed from API.

### TODO

* [x] Unit tests
* [x] Clarify requirements for `unmodifiedAcceptedCharacterCount`

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/telemetryService.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/telemetryService.test.ts
@@ -334,4 +334,39 @@ describe('TelemetryService', () => {
             expect(calledArg.telemetryEvent.chatInteractWithMessageEvent.acceptedLineCount).to.be.undefined
         })
     })
+
+    it('should emit CodeCoverageEvent event', () => {
+        const timestamp = new Date(Date.now())
+        const expectedEvent = {
+            codeCoverageEvent: {
+                customizationArn: 'test-arn',
+                programmingLanguage: { languageName: 'typescript' },
+                acceptedCharacterCount: 123,
+                totalCharacterCount: 456,
+                // timestamp,
+            },
+        }
+        mockCredentialsProvider.setConnectionMetadata({
+            sso: {
+                startUrl: 'idc-start-url',
+            },
+        })
+        telemetryService = new TelemetryService(mockCredentialsProvider, 'bearer', {} as Telemetry, {})
+        const invokeSendTelemetryEventSpy: sinon.SinonSpy = sinon.spy(
+            telemetryService,
+            'invokeSendTelemetryEvent' as any
+        )
+        telemetryService.updateOptOutPreference('OPTIN')
+
+        telemetryService.emitCodeCoverageEvent({
+            languageId: 'typescript',
+            customizationArn: 'test-arn',
+            acceptedCharacterCount: 123,
+            totalCharacterCount: 456,
+        })
+
+        sinon.assert.calledOnce(invokeSendTelemetryEventSpy)
+        sinon.assert.calledWith(invokeSendTelemetryEventSpy, sinon.match(expectedEvent))
+        invokeSendTelemetryEventSpy.restore()
+    })
 })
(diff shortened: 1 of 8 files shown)


server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -119,6 +120,30 @@ export class TelemetryService extends CodeWhispererServiceToken {
...
             perceivedLatencyMilliseconds: perceivedLatencyMilliseconds,
         }
-        this.invokeSendTelemetryEvent(event)
+
+        this.invokeSendTelemetryEvent({
+            userTriggerDecisionEvent: event,
+        })
+    }
+
+    public emitCodeCoverageEvent(params: {
```
pras0131, line 129: This function will be used to emit CodeCoverage events to toolkits DWH as well. Can we change the function arguments so it has the superset fields to be used for both STE and Toolkits DWH?

server/aws-lsp-codewhisperer/src/language-server/telemetryService.test.ts
```diff
@@ -207,4 +209,39 @@ describe('TelemetryService', () => {
...
+
+        telemetryService.emitCodeCoverageEvent({
+            languageId: 'typescript',
+            customizationArn: 'test-arn',
+            acceptedCharacterCount: 123,
+            totalCharacterCount: 456,
+        })
+
+        sinon.assert.calledOnce(invokeSendTelemetryEventSpy)
+        sinon.assert.calledWith(invokeSendTelemetryEventSpy, sinon.match(expectedEvent))
```
pras0131, line 244: Earlier I was also using that, somehow at times it does not match the exact copy and creates an issue. 

Two things can be changed here - using sinon.stub instead of sinon.spy and second - so the called function with arguments has a timestamp (which cannot be exactly matched if we put timestamp in expectedEvent), we can use something like the one i used in another PR [this](https://github.com/aws/language-servers/pull/555/files#diff-d004eac4f42ad129b7b75ec944d1065dbec5f55d7d996d18d66aede173e58034R207)

server/aws-lsp-codewhisperer/src/language-server/telemetryService.test.ts
```diff
@@ -207,4 +209,39 @@ describe('TelemetryService', () => {
...
+                // timestamp,
+            },
+        }
+        mockCredentialsProvider.setConnectionMetadata({
+            sso: {
+                startUrl: 'idc-start-url',
+            },
+        })
+        telemetryService = new TelemetryService(mockCredentialsProvider, 'bearer', {} as Telemetry, {})
+        const invokeSendTelemetryEventSpy: sinon.SinonSpy = sinon.spy(
```
pras0131, line 230: Let's mock sendTelemetryEvent instead?

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
viktorsaws: Let's do this as part of follow-up refactoring task. This params based on CodeCoverage event

pras0131 (approved):

PR #554, opened 2024-10-29
Update workspace fs interface to use expected defaults

natdurlik:
## Problem
Currently, in the standalone Workspace file system, there are some unreasonable default options passed to the Node.js fs functions, which can be confusing to the developers and produce unexpected results.
## Solution
Updated interface in the language-server-runtimes repo: https://github.com/aws/language-server-runtimes/pull/247. 
Applied the necessary changes in the language servers to use the modified workspace interface.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/dependencyGraph/dependencyGraph.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/dependencyGraph/dependencyGraph.ts
@@ -186,7 +186,7 @@ export abstract class DependencyGraph {
      */
     protected async removeDir(dir: string) {
         if (await this.workspace.fs.exists(dir)) {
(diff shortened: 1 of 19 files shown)


volodkevych (approved):

server/aws-lsp-codewhisperer/src/language-server/dependencyGraph/dependencyGraph.ts
```diff
@@ -186,7 +186,7 @@ export abstract class DependencyGraph {
      */
     protected async removeDir(dir: string) {
         if (await this.workspace.fs.exists(dir)) {
-            await this.workspace.fs.remove(dir)
+            await this.workspace.fs.rm(dir, { force: true })
```
ege0zcan, line 189: This should also set `recursive: true`

PR #557, opened 2024-10-30
Make AWS Q endpoint url configurable

imykhai:
## Notes
Making AWS Q endpoint url configurable. 
When `AWS_Q_ENDPOINT_URL` is set, sdk clients will use it instead of the default endpoint. 
The SDK clients will also honour `AWS_Q_REGION` but if the endpoint url already contains a region then it will be ignored based on my observations which is the desired behaviour. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/client/streamingClient/codewhispererStreamingClient.ts
+++ b/server/aws-lsp-codewhisperer/src/client/streamingClient/codewhispererStreamingClient.ts
@@ -1,10 +1,7 @@
 import { CodeWhispererStreaming, CodeWhispererStreamingClientConfig } from '@amzn/codewhisperer-streaming'
 import { ConfiguredRetryStrategy } from '@aws-sdk/util-retry'
 import { readFileSync } from 'fs'
-
-const codeWhispererRegion = 'us-east-1'
-
-const codeWhispererEndpoint = process?.env.AWS_Q_ENDPOINT_URL ?? 'https://codewhisperer.us-east-1.amazonaws.com/'
+import { AWS_Q_REGION, AWS_Q_ENDPOINT_URL } from '../../constants'
 
 export class StreamingClient {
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/client/streamingClient/codewhispererStreamingClient.ts
```diff
@@ -4,7 +4,7 @@ import { readFileSync } from 'fs'
 
 const codeWhispererRegion = 'us-east-1'
 
-const codeWhispererEndpoint = 'https://codewhisperer.us-east-1.amazonaws.com/'
+const codeWhispererEndpoint = process?.env.AWS_Q_ENDPOINT_URL ?? 'https://codewhisperer.us-east-1.amazonaws.com/'
```
pras0131, line 7: we can configure this URL in some util file, so we don't need to change in multiple files.

volodkevych, line 7: How will we define those in browser?

imykhai: process will be undefined in browser and this will fall back to the default url.

imykhai: done

imykhai: if we wanted to, we could use webstorage/cookies or [self](https://developer.mozilla.org/en-US/docs/Web/API/Window/self) to override the default endpoint URL

volodkevych: We could inject env vars through `runtime` feature, as it's currently done for `platform`.

volodkevych (approved):

pras0131 (approved):

PR #559, opened 2024-10-31
Bug fixes and preparing for observability additions

floralph:
identityActivation.ts only contains changes for validating during development.

Removed potential PII from error messages created by identity server.

ensureSsoAccountAccessScope added to ensure that sso:account:access scope is added to scopes if it doesn't exist before calls to registerClient and createToken.

Observability interface added that holds references to logging and telemetry features currently.  An instance of observability is plumbed through all aws-lsp-identity classes/functions that can emit logging/telemetry.  Actual log/emitMetric calls will be added in an upcoming PR.

Added a ServerBase and supporting types in language-server/utils.ts for making IdentityServer and future servers easier to write.

IdentityServer updated to be class extending ServerBase, IdentityServerFactory code moved into initialize() method almost verbatim.  Now retrieves clientName from InitializeParams.  Will remove clientName from protocol/identity-management in a future update.

Added support to cancel authorization webserver waiting if destination uses CancellationToken.

Minor style updates to success/failure web page (index.html) and now show client name instead of hard-coded VSC Toolkit as tool.

### DIFF
--- a/server/aws-lsp-identity/src/language-server/utils.ts
+++ b/server/aws-lsp-identity/src/language-server/utils.ts
@@ -1,0 +1,67 @@
+import {
+    Chat,
+    CredentialsProvider,
+    IdentityManagement,
+    Logging,
+    Lsp,
+    Notification,
+    Runtime,
+    Workspace,
+    PartialInitializeResult,
+    Telemetry,
+} from '@aws/language-server-runtimes/server-interface'
+import { InitializeParams, HandlerResult, InitializeError } from 'vscode-languageserver'
+
+export function ensureSsoAccountAccessScope(scopes?: string[]): string[] {
+    const ssoAccountAccessScope = 'sso:account:access'
+
+    if (!scopes) {
+        scopes = [ssoAccountAccessScope]
+    } else if (!scopes.includes(ssoAccountAccessScope)) {
+        scopes.push(ssoAccountAccessScope)
+    }
+
+    return scopes
+}
+
+export interface ServerFeatures {
+    chat: Chat
+    credentialsProvider: CredentialsProvider
+    identityManagement: IdentityManagement
+    logging: Logging
+    lsp: Lsp
+    notification: Notification
+    runtime: Runtime
+    telemetry: Telemetry
+    workspace: Workspace
+}
+
+export interface Observability {
+    logging: Logging
+    telemetry: Telemetry
+}
+
+export abstract class ServerBase implements Disposable {
+    protected readonly observability: Observability
+    protected readonly disposables: (Disposable | (() => void))[] = []
+
+    constructor(protected readonly features: ServerFeatures) {
+        this.features.lsp.addInitializer(params => this.initialize(params))
+        this.observability = { logging: this.features.logging, telemetry: this.features.telemetry }
+    }
+
+    [Symbol.dispose]() {
+        let disposable
+        while ((disposable = this.disposables.pop())) {
+            if (disposable instanceof Function) {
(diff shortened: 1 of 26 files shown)


server/aws-lsp-identity/src/language-server/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (!scopes) {
+        scopes = [ssoAccountAccessScope]
+    } else if (!scopes.includes(ssoAccountAccessScope)) {
+        scopes.push(ssoAccountAccessScope)
+    }
+
+    return scopes
+}
+
+export interface ServerFeatures {
```
volodkevych, line 27: This should be coming from the server interface from runtimes package for single source of truth.

server/aws-lsp-identity/src/language-server/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    telemetry: Telemetry
+}
+
+export abstract class ServerBase implements Disposable {
+    protected readonly observability: Observability
+    protected readonly disposables: (Disposable | (() => void))[] = []
+
+    constructor(protected readonly features: ServerFeatures) {
+        this.features.lsp.addInitializer(params => this.initialize(params))
+        this.observability = { logging: this.features.logging, telemetry: this.features.telemetry }
```
volodkevych, line 50: nit: Having `features` and `observability` as separate fields on the level of server does not look super clean to me - observability is just a subset of features, both objects give access to the same stuff.

server/aws-lsp-identity/src/language-server/utils.ts
floralph: I agree and I'm happy to move it there, but I was trying to avoid changing the runtimes repo again until Eclipse/Q gives me the green light to remove clientName from the identity-management APIs.

Or are you saying this already exists in the runtimes package?  This https://github.com/aws/language-server-runtimes/blob/main/runtimes/server-interface/server.ts is the closest I found, but not usable in that form.

server/aws-lsp-identity/src/language-server/utils.ts
floralph: I agree its redundant, but the Observability interface was more for having to thread both logging and telemetry through the constructors of almost every class/function in the aws-lsp-identity.  I was trying to reduce the addition of multiple parameters everywhere and not have to thread new observability features through all of the code again (i.e. just add them to the Observability interface and they become available everywhere).  

I could remove the ServerBase.observability property, but I'd still have to create a instance during the initialize call to pass around to all of the dependents.  For reuse, it seemed easer to just have a property containing an instance ready for use at any time.  Every subclass will need to use it in its initialize method unless it doesn't emit observability data for some reason.

server/aws-lsp-identity/src/language-server/utils.ts
volodkevych: Another temporary option would be something like this:
`export type ServerFeatures = Parameters<Server>[0]`, until we define interface properly in runtimes package. I would like to avoid different set of features in different places of code.

volodkevych: Otherwise, please make sure to move this to runtime in the following PR.

volodkevych (approved):

server/aws-lsp-identity/src/language-server/utils.ts
floralph: Added a TODO comment here to migrate this interface to runtimes

PR #562, opened 2024-10-31
Emit chat add message event

pras0131:
## Problem
Reference: https://issues.amazon.com/issues/AWS-Cloud9-28678
Emit add message event
## Solution
The PR emits an chatAddMessage event, which is emitted once full response has been retrieved for a query in a chat. 
**hasProjectLevelContext** field has been set to false as it depends on `@workspace` feature, which is not provided through LSP server, as of now. 
If language is not present and event.programmingLanguage is passed as undefined, it results in an error because code still tries to access this field. That's why it has been set separately to not even set it if language is not present. 

Tested through test cases and trigerring an actual event on local, which passed successfully.

## License


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
```diff
@@ -95,6 +103,18 @@ describe('ChatController', () => {
...
+            getCredentials: sinon.stub().returns({ token: 'token' }),
+            getConnectionMetadata: sinon.stub().returns({
+                sso: {
+                    startUrl: undefined,
+                },
+            }),
+        }
+
+        telemetryService = new TelemetryService(mockCredentialsProvider, 'bearer', {} as Telemetry, {})
+        invokeSendTelemetryEventStub = sinon.stub(telemetryService, 'sendTelemetryEvent' as any)
```
viktorsaws, line 117: Did you plann to add unit tests for new TelemetryService invocation here? Let's add it as part of the change introducing new call

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -199,4 +203,50 @@ export class TelemetryService extends CodeWhispererServiceToken {
...
+        activeEditorTotalCharacters?: number
+        timeToFirstChunkMilliseconds?: number
+        timeBetweenChunks?: number[]
+        fullResponselatency?: number
+        requestLength?: number
+        responseLength?: number
+        numberOfCodeBlocks?: number
+        hasProjectLevelContext?: number
+    }) {
+        if (!params.conversationId || !params.messageId) {
```
viktorsaws, line 223: should we make these params required?

viktorsaws (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
pras0131: not yet but can add those later and create a task,whereever telemetry service is being invoked.

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: source where we get the conversationId and messageId from are optional. All the null/non-null checks i thought to better add here instead. 

One thing I can do here, when I move the respective event to destination in the telemetryService file, then I can see if we need to pass required from there then. Will take a note of this.

PR #574, opened 2024-11-05
Emit chat and inline events to destination

pras0131:
## Problem
Reference: https://issues.amazon.com/issues/AWS-Cloud9-28780, https://issues.amazon.com/issues/AWS-Cloud9-28679
~~Report chat and inline completion events to destination based on the flag sent by the destination **enableTelemetryEventsToDestination**~~

UPD: in this PR we do not enable the configuration `enableTelemetryEventsToDestination` flag, and we always emit events with LSP and to SendTelemetryEvent API.

We will introduce the flag to disable telemetry, when all events are migrated to SendTelemetryEvent.

## Solution
All the respective chat events are moved to TelemetryService, so it being a single point to emit events to STE or destination. 
The PR also removes the call where they were being called before, so they are only called based on the condition now. 
For **emitChatAddMessage** method in telemetry service, the arguments have been divided into params and additionalParams..additionalParams are only needed for event emit to destination. So in future when emitting events to destination has to be removed, the struct can be removed all together.
There is a change in TelemetryService constructor to set the default value of **enableTelemetryEventsToDestination**  flag to false. 
Test cases in the telemetry service have been updated to make sure if the flag is on, the respective event to destination is also emitted. 

## License


server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -185,6 +191,16 @@ export class TelemetryService extends CodeWhispererServiceToken {
         if (options?.conversationId === undefined) {
             return
         }
+        if (this.enableTelemetryEventsToDestination) {
```
viktorsaws, line 194: Let's add a unit test for this condition

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -209,6 +225,18 @@ export class TelemetryService extends CodeWhispererServiceToken {
         modificationPercentage: number
         customizationArn?: string
     }) {
+        if (this.enableTelemetryEventsToDestination) {
```
viktorsaws, line 228: I suggest to set `enableTelemetryEventsToDestination` to `false` as default value explicitly where it is first defined first https://github.com/aws/language-servers/pull/574/files#diff-9651c5ce85ab6404bc20ff59b61b93254f540d5e677f455b75e3d17ce5c07f08R36

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: There is already a unit test update for this when I am setting telemetryService.updateTelemetryEventsToDestination to true.

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
pras0131: I'll set as default false in the constructor itself.

viktorsaws:
As we are introducing flag, which disables telemetry sent back to destination over LSP, we need to make sure all telemetry events respect the flag value. 
Currently, there are events that are still sent to destination: e.g. in `CodeWhispererServer` we have `codewhisperer_serviceInvocation`, `codewhisperer_perceivedLatency` and `codewhisperer_userDecision`. In `ChatController` we have `amazonq_sendFeedback`. 
And there's whole bunch of them in `netTransform` server.

We need to review and make sure they also are disabled on flag.

server/aws-lsp-codewhisperer/src/language-server/telemetryService.ts
```diff
@@ -132,6 +140,43 @@ export class TelemetryService extends CodeWhispererServiceToken {
     }
 
     public emitUserTriggerDecision(session: CodeWhispererSession, timeSinceLastUserModification?: number) {
+        if (this.enableTelemetryEventsToDestination) {
```
viktorsaws, line 143: We still need explicit unit test for cases when `enableTelemetryEventsToDestination` is false. Currently we don't have test that would verify that this codepath is not called.

pras0131: added the same.

viktorsaws (approved):

PR #577, opened 2024-11-06
Add skeleton of notification server.

floralph:
This commit adds the skeleton of the notification server, an app for it, and stubbed in activation in vscode sample client.  Some types reused from identity have started to be migrated to aws-lsp-core.  Comments have been added and some code in the server/service has been added for initialization.  The structure of the notification fetcher pipeline has been defined as types, but not much further yet.


server/aws-lsp-notification/src/language-server/notificationServer.ts
```diff
@@ -0,0 +1,76 @@
...
+import { NotificationService, ShowNotification } from './notificationService'
+import { NotificationFollowupParams, NotificationParams } from '@aws/language-server-runtimes/protocol'
+import { FilesystemMetadataStore } from '../notifications/metadata/filesystemMetadataStore'
+import { S3Fetcher } from '../notifications/toolkits/s3Fetcher'
+import { CritieriaFilteringFetcher } from '../notifications/toolkits/criteriaFilteringFetcher'
+import { MetadataFilteringFetcher } from '../notifications/metadata/metadataFilteringFetcher'
+import { Fetcher } from '../notifications/fetcher'
+
+export class NotificationServer extends ServerBase {
+    // Silently drop notifications from constructor to initliaze when ability to send notifications is available
```
volodkevych, line 18: Why do you need to do this?

floralph: Just TS minutiae.  It's not used before initialize is called, but it can't be set until initialized is called.  The other option is to make it optional and do non-null assertions everywhere.

floralph: In general, the code here will likely change/evolve quite a bit as notifications is implemented.

volodkevych (approved):

PR #578, opened 2024-11-06
improve clean script to prevent issues with caching

francescoopiccoli:
## Problem
Improve clean up script to prevent caching mechanisms to override changes made in the code and lead to misleading results.

## Solution
Add `git clean -fdx `to cleanup script for a thorough cleanup

## License

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
         "check:formatting": "prettier --check .",
         "format": "prettier --write .",
         "format-staged": "npx pretty-quick --staged",
-        "clean": "ts-node ./script/clean.ts && npm run clean --workspaces --if-present",
+        "clean": "ts-node ./script/clean.ts && npm run clean --workspaces --if-present && sh ./script/git-clean.sh",
         "precompile": "npm run precompile --workspaces --if-present",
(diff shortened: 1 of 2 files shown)


script/git-clean.sh
imykhai, line 1: nice script!

package.json
```diff
@@ -17,7 +17,7 @@
         "check:formatting": "prettier --check .",
         "format": "prettier --write .",
         "format-staged": "npx pretty-quick --staged",
-        "clean": "ts-node ./script/clean.ts && npm run clean --workspaces --if-present",
+        "clean": "ts-node ./script/clean.ts && npm run clean --workspaces --if-present && sh ./script/git-clean.sh",
```
imykhai, line 20: if git will remove anything not under source control, do we need to run `ts-node ./script/clean.ts && npm run clean --workspaces --if-present` first?

imykhai (approved):

package.json
francescoopiccoli: I think keeping it this way, makes it easier to see the files that `git clean` would delete, since it prompts a much shorter list of files, potentally avoiding unintended deletes (newly created files currently unstaged), hence making it safer for the user. With a long list of mostly all modules files the user might assume its all modules imo.

PR #579, opened 2024-11-06
Amazon Q: fix proxy server configuration

viktorsaws:
## Problem
HTTP proxy setup does not work for CodeWhisperer capabilities except Q Chat.

## Solution
* Fixed bug in `httpOptions` AWS Configuration passed to `CodewhispererService` in `proxy-servers` versions of capabilities: proxy agent object was not passed in correct field expected (https://github.com/aws/language-servers/pull/579/files#diff-6f77316cd6785df25fcc13badc7a24b9fa77b4e594455d07011c4ccc817d0be2R25-R27)
* Switched to `hpagent` package for all proxies. `proxy-http-agent` that was used originally was also misconfigured - arguments passed were not in correct shape.

Tested with local proxy setup using Charles and Sample VSCode extension on my Macbook. Verified that with setting `HTTPS_PROXY` and `AWS_CA_BUNDLE` in https://github.com/aws/language-servers/blob/main/.vscode/launch.json#L90-L91 result in calls for `InlineCompletion` and `QConfigurationServer` to go through proxy.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts
@@ -8,6 +8,7 @@ import { QChatServer } from './qChatServer'
 import { QConfigurationServerToken } from './configuration/qConfigurationServer'
 import { readFileSync } from 'fs'
 import { ConfigurationOptions } from 'aws-sdk'
+import { HttpsProxyAgent } from 'hpagent'
 
 const makeProxyConfig = () => {
     let additionalAwsConfig: ConfigurationOptions = {}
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts
```diff
@@ -7,76 +7,48 @@ import { QNetTransformServerToken } from './netTransformServer'
...
-        const proxyAgent = getProxyHttpAgent({
+        const certs = process.env.AWS_CA_BUNDLE ? [readFileSync(process.env.AWS_CA_BUNDLE)] : undefined
+        const agent = new HttpsProxyAgent({
             proxy: proxyUrl,
             ca: certs,
         })
+
         additionalAwsConfig = {
-            httpOptions: proxyAgent,
+            httpOptions: {
```
pras0131, line 25: do we also need to set httpsOptions, as I see QChatServer setting both agents for http and https

imykhai (approved):

PR #586, opened 2024-11-07
Remove conditional requires and convert to top level import

ege0zcan:
## Description

Let's not mix require and import statements in the same file. The benefits of conditional require can be discussed as a followup 

## License


dogusata (approved):

server/aws-lsp-codewhisperer/src/client/streamingClient/codewhispererStreamingClient.ts
```diff
@@ -2,6 +2,8 @@ import { CodeWhispererStreaming, CodeWhispererStreamingClientConfig } from '@amz
 import { ConfiguredRetryStrategy } from '@aws-sdk/util-retry'
 import { readFileSync } from 'fs'
 import { AWS_Q_REGION, AWS_Q_ENDPOINT_URL } from '../../constants'
+import { NodeHttpHandler } from '@smithy/node-http-handler'
```
volodkevych, line 5: Why do we continue using it for chat only?

PR #595, opened 2024-11-08
Amazon Q: port logic for CodePercentage modifications calculation

viktorsaws:
## Problem
* CodePercentage is emitted with 0 values when there were failed calls to service API. Expected behaviour - it should not emit telemetry
* When returned suggestion is right-merged with document content, it sometimes resulted in incorrect calculation of accepted token count being more than total document changes.
* CodeCoverage event calculation is not in sync with VSCode implementation. Up to date implementation is in https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts and it contains next changes:
  * New line characters inputs are treated as 1 character
  * Whitespace-only changes are ignored
  * Inserts over 50 characters are ignored

## Solution
Port logic from https://github.com/aws/aws-toolkit-vscode/blob/master/packages/core/src/codewhisperer/tracker/codewhispererCodeCoverageTracker.ts.

* Increment `CodePercentage.invocationCount` only after service call succeeds. Before we always incremented it, which resulted in telemetry being sent after errors with 0-ed values.
* Changed logic to ignore content that was right-merged by language server, by caching actually returned `insertText` for sugestions  in InlineCompletion `Session` object
* Update logic for `CodePercentage.countTotalTokens` to handle special cases mentioned above.
  * To prevent double-counting when accepted recommendation is less than 50 characters, added `fromCodeWhisperer` flag and special conditional case.

TODO
* [x] Check how changes for deletion are handled in VSCode - not handled
* [x] Unit tests

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
@@ -30,14 +30,8 @@ import {
     CodeWhispererPerceivedLatencyEvent,
     CodeWhispererServiceInvocationEvent,
     CodeWhispererUserDecisionEvent,
-    CodeWhispererUserTriggerDecisionEvent,
 } from './telemetry/types'
-import {
-    getCompletionType,
-    getEndPositionForAcceptedSuggestion,
-    getUnmodifiedAcceptedTokens,
-    isAwsError,
-} from './utils'
+import { getCompletionType, getEndPositionForAcceptedSuggestion, isAwsError } from './utils'
 import { getUserAgent, makeUserContextObject } from './utilities/telemetryUtils'
 import { Q_CONFIGURATION_SECTION } from './configuration/qConfigurationServer'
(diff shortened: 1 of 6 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -569,11 +568,11 @@ export const CodewhispererServerFactory =
                 k => params.completionSessionResult[k].accepted
             )
             const acceptedSuggestion = session.suggestions.find(s => s.itemId === acceptedItemId)
-
-            if (acceptedSuggestion !== undefined) {
+            if (acceptedSuggestion !== undefined && acceptedSuggestion.insertText) {
```
pras0131, line 571: nit: we can simplify this to `if (acceptedSuggestion?.insertText) ` and remove the inner if clause

pras0131 (approved):

PR #599, opened 2024-11-11
Load proxy configurations in code whisperer base class

pras0131:
## Problem
Reference: https://issues.amazon.com/issues/AWS-Cloud9-28845
The PR aims to initialize proxy settings in Code Whisperer base class rather than fetching the value passed from constructor. 

## Solution

- The PR loads the proxy configurations from the code whisperer base class and sets the proxy configurations once the client is initialized. 
- The PR also contains the requisite changes to update the client specific AWS configuration, rather than loading global configurations. 
- Other servers which were using CodeWhispererServer/IAM - TokenProxy now no need to pass this proxy configuration in the constructor.
- The above changes have been tested on local environment. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts
@@ -7,50 +7,23 @@ import { QNetTransformServerToken } from './netTransformServer'
 import { QChatServer } from './qChatServer'
 import { QConfigurationServerToken } from './configuration/qConfigurationServer'
 import { readFileSync } from 'fs'
-import { ConfigurationOptions } from 'aws-sdk'
 import { HttpsProxyAgent } from 'hpagent'
 import { NodeHttpHandler } from '@smithy/node-http-handler'
 
-const makeProxyConfig = () => {
-    let additionalAwsConfig: ConfigurationOptions = {}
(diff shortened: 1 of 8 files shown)


server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts
```diff
@@ -7,50 +7,23 @@ import { QNetTransformServerToken } from './netTransformServer'
...
-            httpOptions: {
-                agent: agent,
-            },
-        }
-    }
-
-    return additionalAwsConfig
-}
-
 export const CodeWhispererServerTokenProxy = CodewhispererServerFactory(credentialsProvider => {
```
viktorsaws, line 13: With this change we are able to remove this `*Proxy` factories - they don't do anything special now, and we have exports of duplicate "regular" factories already, for example here https://github.com/aws/language-servers/blob/479871ed7bce0343526e9c7b985784a4c744372b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts#L664-L669.

Only if we remove these, we'll need to make sure we consume correct exports in our distribution pipeline.

We can first migrate distribution process, and then remove these as follow-up.

viktorsaws (approved):

PR #615, opened 2024-11-15
chore: move codewhisperer-streaming package to core workspace

viktorsaws:
## Problem
Codewhisperer streaming client and `@aws/lsp-codewhisperer` package is setup as a nested monorepo, which causes issues with lifting and conflicting dependencies in main monorepo dependencies tree.

## Solution
Move streaming client to a dedicated workspace in shared `core` workspace of language-servers monorepo.

Tested that `@aws/lsp-codewhisperer` takes dependency, can be compiled, packed and consumed as npm package outside of monorepo.

## License


package.json
```diff
@@ -18,9 +18,11 @@
         "format": "prettier --write .",
         "format-staged": "npx pretty-quick --staged",
         "clean": "ts-node ./script/clean.ts && npm run clean --workspaces --if-present && sh ./script/git-clean.sh",
-        "precompile": "npm run precompile --workspaces --if-present",
-        "compile": "tsc --build --verbose && npm run compile:servers && npm run compile --workspaces --if-present",
+        "precompile": "npm run precompile --workspaces --if-present && npm run compile:core",
+        "compile": "tsc --build --verbose && npm run compile:core && npm run compile:servers && npm run compile:rest",
+        "compile:core": "npm run compile --workspace=core --if-present && npm run build --workspace=core --if-present",
         "compile:servers": "npm run compile --workspace=server --if-present",
+        "compile:rest": "npm run compile --workspace app --workspace client --workspace chat-client --if-present",
```
volodkevych, line 25: nit: can we rename it to `compile:client`?

volodkevych (approved):

package.json
viktorsaws: it's not only client, but also `app` directory which is hosting compiled language servers configs, so `client` can be a bit confusing. I don't like rest either, but didn't come up with better name.

volodkevych: you could do `complie:apps` and `compile:client` separately

PR #627, opened 2024-11-23
Adding logs for invalid projects

pranav-firake:
Adding logs when invalid project is detected in transformation

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/validation.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/validation.ts
@@ -15,10 +15,16 @@ export function validateProject(userInputrequest: StartTransformRequest, logging
     var selectedProject = userInputrequest.ProjectMetadata.find(
         project => project.ProjectPath == userInputrequest.SelectedProjectPath
     )
-    logging.log(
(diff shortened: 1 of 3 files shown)


JiayuW2 (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/validation.ts
```diff
@@ -10,11 +11,20 @@ export function isSolution(userInputrequest: StartTransformRequest): boolean {
...
 
-export function validateProject(userInputrequest: StartTransformRequest): boolean {
+export function validateProject(userInputrequest: StartTransformRequest, logging: Logging): boolean {
     var selectedProject = userInputrequest.ProjectMetadata.find(
         project => project.ProjectPath == userInputrequest.SelectedProjectPath
     )
-    if (selectedProject) return supportedProjects.includes(selectedProject?.ProjectType)
+
+    if (selectedProject) {
+        var isValid = supportedProjects.includes(selectedProject?.ProjectType)
```
ege0zcan, line 20: nit: this should be a const instead of var and then the function could be rewritten like:

```
if (!selectedProject)
   return false
const isValid = supportedProjects.includes(selectedProject.ProjectType)
logging.log(...)
return isValid
```

PR #634, opened 2024-11-29
Make logging stubs more resilient to runtime changes

ege0zcan:
## Problem
Whenever there is a change in the runtimes related to the Logging interface, we need to update our tests. For cases where logging functionality is not the main subject under test, tests shouldn't be impacted by runtimes changes

## Solution
Update the tests to use type assertion so that any changes to runtimes doesn't break the tests

## License


rahmaniaam (approved):

justinmk3:
A much more resilient approach, which also would result in higher-coverage tests for near-zero cost, as well as being able to more meaningfully test real codepaths, would be to assert the actual logs, and drop the use of "stubs" or "mocks" entirely. Instead of stubs/mocks, just let tests *query the logs*.

PR #636, opened 2024-11-29
Improve and add debugging instructions

rahmaniaam:
## Problem
Debugging in the VSCode client is currently broken

## Solution
Pass the args for debugging correctly and update the README instructions on how to debug in the VSCode client as well as independently. Tested by launching the client in local

## License

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,8 +103,10 @@ node ./app/aws-lsp-codewhisperer-binary/aws-lsp-codewhisperer-token-binary.js --
 This repo contains a minimal vscode client that can be used to easily run and
 debug changes to this language server.
 
-1. In the `Run & Debug` menu, run `"Launch as VSCode Extension + Debugging"`
-2. Set breakpoints in `src` where needed.
+1. Run `npm run package` in the root folder
+2. In the `Run & Debug` menu, run `"Launch as VSCode Extension + Debugging"`. Make sure this is launching the server you wish to use, look at the [launch configuration](.vscode/launch.json#L202). Keep the `Attach to AWS Documents Language Server` to use the debugger.
+3. Set breakpoints in the source where needed.
+4. (Optional) in the [activation file](client/vscode/src/activation.ts#L81) you can swap `--inspect` with `--inspect-brk` if you need to debug immediately when the node process starts. Read more about it on this [site](https://www.builder.io/blog/debug-nodejs#:~:text=%2D%2Dinspect%20versus%20%2D%2Dinspect%2Dbrk)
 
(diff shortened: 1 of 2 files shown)


client/vscode/src/activation.ts
```diff
@@ -78,7 +78,9 @@ export async function activateDocumentsLanguageServer(extensionContext: Extensio
         debugOptions.execArgv.push('--pre-init-encryption')
         // Used by the aws-language-server-runtime based servers
         debugOptions.execArgv.push('--set-credentials-encryption-key')
-        const child = cp.spawn('node', [serverModule, ...debugOptions.execArgv])
+
+        // --inspect or --inspect-brk has to be passed right after the node command, else debuggers won't be able to be attached to the process
```
volodkevych, line 82: Is it because `inspect` was treated as a script argument instead of `node` argument? Then comment is a bit misleading.

rahmaniaam: you're right, I'll add that bit as well

rahmaniaam: Good question, I'm not entirely aware on what they do. I'll check first but these most probably will need to be moved as well

CONTRIBUTING.md
```diff
@@ -140,6 +142,14 @@ and be able to debug it all.
 
 > **NOTE**: The lsp client is activated by one of the `activationEvents` defined [here](https://github.com/aws/language-servers/blob/06fd81d1e936648ef43243865039f89c7ac142a7/client/vscode/package.json#L18-L22), the lsp client then starts the LSP server.
 
+### With Other Clients
+Using other clients can also be done with the bundle created from this package.
+
+1. Produce a local server bundle `npm run package`. The `app/` folder contains the configuration of server(s) and it's runtime(s).
+2. Take note of the bundle you wish to use. For this example we will use `app/aws-lsp-codewhisperer-runtimes/out/token-standalone.js`.
+3. Run the bundle using these args `node --inspect=6012 {rootPath}/app/aws-lsp-codewhisperer-runtimes/out/token-standalone.js --nolazy --preserve-symlinks --stdio --pre-init-encryption --set-credentials-encryption-key` or adjust as needed. Refer to the [activation file](client/vscode/src/activation.ts). *NOTE: make sure that --inspect or --inspect-brk args are passed right after the `node` command*
```
awschristou, line 150: Consider adding a sentence explaining why the inspect arg needs to be where it is. Its obvious to me right now because that was the thing I most recently got wrong, but six months from now I'll likely forget, and the next person may not realize the significance.

awschristou (approved):

volodkevych (approved):

PR #638, opened 2024-12-03
Remove telemetry generation from identity server, will not be used.

floralph:
Removes telemetry definition generation added in https://github.com/aws/language-servers/pull/625 This creates coupling to toolkit-specific telemetry and brings in much more than is needed.  As aws-lsp-identity is a Flare server and defined in Flare, the limited telemetry emitted by this server will be defined directly in this repo as well.

NOTE: The identity server in Flare is available to all clients ("destinations") that consume Flare, not just AWS Toolkits. Metrics generated in the identity server are emitted to clients (not posted to backends) via the [LSP telemetry/event notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#telemetry_event) and it is up to the client to handle that metric as it sees fit (whether this is to ignore, transform, and/or post it to some backend system).


ivikash (approved):

justinmk3:
> the limited telemetry emitted by this server will be defined directly in this repo as well.

If it's limited, then only the limited metrics from toolkits telemetry would be used. 

> This creates coupling to toolkit-specific telemetry and brings in much more than is needed.

Metrics in https://github.com/aws/aws-toolkit-common are "common". Any toolkit-specific metrics can just be not used by clients that don't want them. But for metrics that are useful, isn't it better to align on common shapes?

floralph:
> > the limited telemetry emitted by this server will be defined directly in this repo as well.
> 
> If it's limited, then only the limited metrics from toolkits telemetry would be used.
> 
> > This creates coupling to toolkit-specific telemetry and brings in much more than is needed.
> 
> Metrics in https://github.com/aws/aws-toolkit-common are "common". Any toolkit-specific metrics can just be not used by clients that don't want them. But for metrics that are useful, isn't it better to align on common shapes?

The identity server in Flare is available to all clients ("destinations") that consume Flare, not just AWS Toolkits.  Metrics generated in the identity server are emitted to clients (not posted to backends) via the [LSP telemetry/event notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#telemetry_event) and it is up to the client to handle that metric as it sees fit (whether this is to ignore, transform, and/or post it to some backend system).

There is a possibility TKC may post error/usage related metrics specific to the identity server's operation itself for support purposes.  That implementation can be designed as the need arises.

volodkevych (approved):

volodkevych (approved):

PR #640, opened 2024-12-04
Inject fs feature for reading proxy configs

rahmaniaam:
## Problem
node `fs` is used to read proxy configurations. This results in webpack bundling to fail since it's not available in browser

## Solution
Use the `workspace.fs` feature injected from runtimes instead

- Inject workspace to server creation
- Create mock `workspace` for tests

Note: to be tested in local

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
@@ -122,25 +122,7 @@ describe('ChatController', () => {
             }),
         }
 
-        const mockWorkspace: Workspace = {
-            getWorkspaceFolder: sinon.stub(),
-            getTextDocument: sinon.stub(),
-            getAllTextDocuments: sinon.stub(),
-            fs: {
-                copyFile: sinon.stub(),
-                exists: sinon.stub(),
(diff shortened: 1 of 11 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
```diff
@@ -121,11 +122,31 @@ describe('ChatController', () => {
...
+                getFileSize: sinon.stub(),
+                getServerDataDirPath: sinon.stub(),
+                getTempDirPath: sinon.stub(),
+                readFile: sinon.stub(),
+                readdir: sinon.stub(),
+                isFile: sinon.stub(),
+                rm: sinon.stub(),
+                writeFile: sinon.stub(),
+                appendFile: sinon.stub(),
+                mkdir: sinon.stub(),
```
volodkevych, line 141: Do we have to explicitly specify all the methods of `workspace`? Extending `workspace` will break the build in replica sync again. Maybe we can do something like [here](https://github.com/aws/language-servers/pull/635/files#diff-bb635788256abd858778bc03153a2d8b258179aa8c5a208a85a956b7b83bf9d5R26)

volodkevych (approved):

PR #649, opened 2024-12-06
Add ANTLR4 Language Server

whazor:
ANTLR4 language server and service that can take a grammar, for example one from the [community provided grammars](https://github.com/antlr/grammars-v4), and the server provides diagnostics and auto-complete thanks to [antlr4ng](https://github.com/mike-lischke/antlr4ng) and [antlr4-c3](https://github.com/mike-lischke/antlr4-c3).

In the ANTLR4 runtime, we test the PostgreSQL grammar as a more comprehensive test.

## License


app/aws-lsp-antlr4-runtimes/README.md
```diff
@@ -0,0 +1,4 @@
+Package bundles ANTLR4 LSP Server with standalone DEXP runtime and run integration tests against it.
```
akoreman, line 1: Should probably rename this to 'Flare' runtime

server/aws-lsp-antlr4/src/language-service/service.ts
```diff
@@ -0,0 +1,150 @@
...
+        parserConstructor: (tokenStream: CommonTokenStream) => Parser,
+        parserRoot: string,
+        ignoredTokens?: number[]
+    ) {
+        this.lexerConstructor = lexerConstructor
+        this.parserConstructor = parserConstructor
+        this.parserRoot = parserRoot
+        this.ignoredTokens = new Set(ignoredTokens ?? [])
+    }
+    isSupported(document: TextDocument): boolean {
```
akoreman, line 87: Do we use this anywhere?

whazor: this is implemented via the interface and used internally

akoreman (approved):

imykhai (approved):

app/aws-lsp-antlr4-runtimes/src/antlr-generated/PostgreSQLParserBase.ts
volodkevych, line 1: Why do we need these generated files in the repo?
Will they ever need to be updated? How? It may be good to add this info to readme.

akoreman: These generated files are used to run tests against the language server, without files like these the language server won't be able to work. They are generated by running the `generate-grammar` npm command. 

There should be little need to update these generated files, only when we develop the server further and add features unsupported by these parser/lexer files would they need to be updated.

volodkevych: Do you have to commit them? Cannot `generate-grammar` npm command be part of the build?

volodkevych (approved):

PR #664, opened 2024-12-10
feat(chat-client)!: add legal disclaimer with acknowledgement button

Jurredr:
## Problem
Currently, we show long disclaimer on the bottom of chat: [source](https://github.com/aws/language-servers/blob/main/chat-client/src/client/chat.ts#L51). This needs to move to promptInputStickyCard with "Acknowledge" button in every active or newly added tab, until user clicks that "Acknowledge" button. When user hits that, it should disappear from all the tabs and shouldn't be visible ever again.

## Solution
- Shortened the legal disclaimer in the footer (always visible)
- Added `disclaimerAcknowledged` event to `OutboundChatApi` (`DISCLAIMER_ACKNOWLEDGED` has been added to the contract in `language-server-runtimes`, will be available through NPM once a new release is made there)
- Extended the ChatClientConfig by a boolean flag `disclaimerAcknowledged`

## License

### DIFF
--- a/chat-client/src/client/chat.ts
+++ b/chat-client/src/client/chat.ts
@@ -48,7 +48,7 @@ import { TabFactory } from './tabs/tabFactory'
 const DEFAULT_TAB_DATA = {
     tabTitle: 'Chat',
     promptInputInfo:
-        'Amazon Q Developer uses generative AI. You may need to verify responses. See the [AWS Responsible AI Policy](https://aws.amazon.com/machine-learning/responsible-ai/policy/). Amazon Q Developer processes data across all US Regions. See [here](https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/cross-region-inference.html) for more info. Amazon Q may retain chats to provide and maintain the service.',
+        'Amazon Q Developer uses generative AI. You may need to verify responses. See the [AWS Responsible AI Policy](https://aws.amazon.com/machine-learning/responsible-ai/policy/).',
     promptInputPlaceholder: 'Ask a question or enter "/" for quick actions',
 }
 
--- a/chat-client/src/client/chat.ts
+++ b/chat-client/src/client/chat.ts
@@ -90,7 +90,7 @@ export const createChat = (
                 const chatConfig: ChatClientConfig = params?.quickActions?.quickActionsCommandGroups
                     ? {
                           quickActionCommands: params.quickActions.quickActionsCommandGroups,
-                          disclaimerAcknowledged: config?.disclaimerAcknowledged,
+                          disclaimerAcknowledged: config?.disclaimerAcknowledged ?? false,
                       }
                     : {}
 
--- a/chat-client/src/client/chat.ts
+++ b/chat-client/src/client/chat.ts
@@ -15,6 +15,7 @@ import {
     SEND_TO_PROMPT,
     SendToPromptMessage,
     UiMessage,
+    DISCLAIMER_ACKNOWLEDGED,
 } from '@aws/chat-client-ui-types'
 import {
     CHAT_REQUEST_METHOD,
@@ -52,7 +53,7 @@ const DEFAULT_TAB_DATA = {
     promptInputPlaceholder: 'Ask a question or enter "/" for quick actions',
 }
 
-type ChatClientConfig = Pick<MynahUIDataModel, 'quickActionCommands'>
+type ChatClientConfig = Pick<MynahUIDataModel, 'quickActionCommands'> & { disclaimerAcknowledged?: boolean }
 
 export const createChat = (
     clientApi: { postMessage: (msg: UiMessage | ServerMessage) => void },
@@ -87,7 +88,10 @@ export const createChat = (
             case CHAT_OPTIONS: {
                 const params = (message as ChatOptionsMessage).params
                 const chatConfig: ChatClientConfig = params?.quickActions?.quickActionsCommandGroups
-                    ? { quickActionCommands: params.quickActions.quickActionsCommandGroups }
+                    ? {
+                          quickActionCommands: params.quickActions.quickActionsCommandGroups,
+                          disclaimerAcknowledged: config?.disclaimerAcknowledged,
+                      }
                     : {}
 
                 tabFactory.updateDefaultTabData(chatConfig)
@@ -154,6 +158,9 @@ export const createChat = (
 
             window.addEventListener('message', handleMessage)
         },
+        disclaimerAcknowledged: () => {
+            sendMessageToClient({ command: DISCLAIMER_ACKNOWLEDGED })
+        },
     }
 
     const messager = new Messager(chatApi)
@@ -162,7 +169,7 @@ export const createChat = (
         ...(config?.quickActionCommands ? { quickActionCommands: config.quickActionCommands } : {}),
     })
 
-    const [mynahUi, api] = createMynahUi(messager, tabFactory)
+    const [mynahUi, api] = createMynahUi(messager, tabFactory, config?.disclaimerAcknowledged ?? false)
 
     mynahApi = api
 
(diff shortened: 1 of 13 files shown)


chat-client/src/client/chat.ts
```diff
@@ -15,6 +15,7 @@ import {
     SEND_TO_PROMPT,
     SendToPromptMessage,
     UiMessage,
+    DISCLAIMER_ACKNOWLEDGED,
```
viktorsaws, line 18: Looks like we need to publish latest `@aws/chat-client-ui-types` to npm to expose missing types

Jurredr: Yeah, I'm aware. See the second bullet point in disclaimer.

viktorsaws:
Why did we decided to make this a part of Chat Client here and not a feature in MynahUI itself? Looks like common reusable pattern that might occur in other destinations using MynahUI directly.

chat-client/src/client/mynahUi.ts
```diff
@@ -144,6 +150,7 @@ export const createMynahUi = (messager: Messager, tabFactory: TabFactory): [Myna
             messager.onTabAdd(tabId)
             const defaultTabConfig: Partial<MynahUIDataModel> = {
```
dogusata, line 151: I suspect that this is not gonna work as expected. Below, we're setting the tab defaults, and while doing that depending on the initial value of the disclaimerAcknowledged argument we're adding the disclaimer part. 
This is set to all tabs which will open in future too. 

In here, depending on the situation we're adding or not adding the disclaimer card. But NOT adding means that we're also not touching the defaults. And defaults still contain the disclaimer card.

What I would prefer would be not playing with the defaults and not setting the disclaimerCard to defaults at all. 

Do it in the initial `tabs`, do it in the `onTabAdd`, but not in the `defaults`. We don't need to write it to defaults since we're checking the statement when a new tab is open and adding it depending on the statement to that tab.

chat-client/src/client/mynahUi.ts
```diff
@@ -237,14 +244,30 @@ export const createMynahUi = (messager: Messager, tabFactory: TabFactory): [Myna
...
         tabs: {
             [initialTabId]: {
                 isSelected: true,
-                store: tabFactory.createTab(true),
+                store: tabFactory.createTab(true, disclaimerCardActive),
             },
         },
         defaults: {
-            store: tabFactory.createTab(true),
+            store: tabFactory.createTab(true, disclaimerCardActive),
```
dogusata, line 270: We should not give it on defaults. It should only work when a new tab is open by checking the acknowledgement state and add it depending on the situation. 

When it is acknowledged, it will not come back. So what we need to do is to check only `should we add it, or not.`. But if we add it to defaults, it turns out to `should we REMOVE it or not`. We should not care about the removal part, only adding depending on the statement should be fine.

We don't need it in defaults.

dogusata (changes requested):

client/vscode/src/chatActivation.ts
```diff
@@ -198,7 +198,7 @@ function generateJS(webView: Webview, extensionUri: Uri): string {
     <script type="text/javascript" src="${entrypoint.toString()}" defer onload="init()"></script>
     <script type="text/javascript">
         const init = () => {
-            amazonQChat.createChat(acquireVsCodeApi());
+            amazonQChat.createChat(acquireVsCodeApi(), {disclaimerAcknowledged: false});
```
ege0zcan, line 201: If I understand this correctly, we now have an expectation that destinations will store this acknowledgement in their local storage. In this case let's be sure to mark the changes in this PR as breaking change and bump versions accordingly and also make sure we talk with toolkit teams. If we were to roll this out now, users would be shown the acknowledge message on each IDE restart

ege0zcan:
nit: Is the commit history correct on this PR? I see commit with commit message: `Skip publishing Slack notification for Dependabot PRs`

chat-client/src/client/mynahUi.ts
Jurredr: Resolved, thanks.

chat-client/src/client/mynahUi.ts
```diff
@@ -237,14 +244,30 @@ export const createMynahUi = (messager: Messager, tabFactory: TabFactory): [Myna
...
         },
+        onInBodyButtonClicked: (tabId, messageId, action, eventId) => {
+            if (action.id === disclaimerAcknowledgeButtonId) {
+                // Hide the legal disclaimer card
+                disclaimerCardActive = false
+
+                // Update the disclaimer getting acknowledged
+                messager.onDisclaimerAcknowledged()
+
+                // TODO: create telemetry
```
volodkevych, line 255: Do we need this TODO?

chat-client/src/client/mynahUi.ts
```diff
@@ -237,14 +244,30 @@ export const createMynahUi = (messager: Messager, tabFactory: TabFactory): [Myna
...
+
+                // Remove all disclaimer cards from all tabs
+                Object.keys(mynahUi.getAllTabs()).forEach(storeTabKey => {
+                    mynahUi.updateStore(storeTabKey, { promptInputStickyCard: null })
```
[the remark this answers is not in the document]
volodkevych: I'm confused, because `promptInputStickyCard ` does not look disclaimer-specific. Should not we look for it somehow by `disclaimerAcknowledgeButtonId`?

client/vscode/src/chatActivation.ts
volodkevych: True, we need to bump the version.

Jurredr: This is a breaking change indeed, as there's a new config field with expectation that destinations will store the acknowledgement state somewhere themselves. The version should be bumped, and I'm in contact with toolkit teams.

chat-client/src/client/mynahUi.ts
Jurredr: No, already removed.

Jurredr:
> nit: Is the commit history correct on this PR? I see commit with commit message: `Skip publishing Slack notification for Dependabot PRs`

Thanks for pointing that out, not sure how that one ended up in here, that commit has already been merged into main from a different branch.

chat-client/src/client/mynahUi.ts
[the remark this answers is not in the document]
Jurredr: Correct, a tab can only have 1 `promptInputStickyCard` in FlareChatUI. As far as I know, there is no way to check the content of the `promptInputStickyCard` for each tab to ensure it's actually closing the legal acknowledgement card. It's done in the same way in the VSCode toolkits PR here (https://github.com/aws/aws-toolkit-vscode/pull/6151).

viktorsaws:
when merging this PR, prefer to use `Squash & Merge` option to keep history in main clean. I'd suggest to do this going forward for PRs with many small commits, that at the end contribute only 1 feature

Jurredr:
Will do, I also updated the commit history so the messages follow the new conventional commits format.

client/vscode/src/chatActivation.ts
volodkevych: Not from API, but from behavior point of view. If we push the same major version, clients will get disclaimer popping up again and again until the corresponding event is handled on client side. Which is unacceptable.

chat-client/src/client/mynahUi.ts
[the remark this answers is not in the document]
volodkevych: Something for followup: what if customer never acknowledges the disclaimer, would `promptInputStickyCard` become unusable for any other functionality? What if some other logic uses `promptInputStickyCard` to set some different card, will mynah-ui catch it somehow or will it just get overriden? Do we have corresponding tests? @dogusata

volodkevych (approved):

Jurredr:
Noticed it was failing, will fix tomorrow.

chat-client/src/client/mynahUi.ts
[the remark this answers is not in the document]
dogusata: @volodkevych Yes, we can have only one card on that place. Since the above is full of cards. We're not giving any opportunity to use that place also as a conversation block.

That place is specifically designed only for disclaimer or terms conditions update. However UX team needs to decide it if they want to show a second thing there. When they want, we can do a backwards compatible implementation on MynahUI side. Not here.

dogusata (approved):

PR #675, opened 2024-12-16
feat: handle virtual spaces when inserting code to cursor position

ege0zcan:
## Desription
Adds the ability to handle virtual spaces to chat server

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
@@ -566,6 +566,90 @@ describe('ChatController', () => {
             assert.deepStrictEqual(insertedText, `const x = 1;\n${' '.repeat(12)}const y = 2;`)
         })
 
+        it('handles code starting with multiple blank lines correctly', async () => {
+            // Create a document with some existing indentation
+            const documentContent = `${' '.repeat(4)}const existingCode = true;`
+            const document: TextDocument = TextDocument.create('test.ts', 'typescript', 1, documentContent)
+            testFeatures.workspace.getTextDocument.resolves(document)
+
+            // Position cursor at an indented position
+            const cursorPosition = Position.create(0, 4)
+            const params: InsertToCursorPositionParams = {
+                textDocument: { uri: 'test.ts' },
+                cursorPosition,
+                // Code starts with 3 blank lines, followed by actual code
+                code: '\n\n\nfunction test() {\n    console.log("test");\n}',
+                tabId: mockTabId,
+                messageId: 'XXX',
+            }
+            await chatController.onCodeInsertToCursorPosition(params)
+
+            const documentChanges = testFeatures.lsp.workspace.applyWorkspaceEdit.firstCall.args[0].edit.documentChanges
+            assert(documentChanges)
+            const insertedText = (documentChanges[0] as TextDocumentEdit).edits[0].newText
+            // The blank lines should have no indentation
+            // Only the actual code lines should be indented
+            // First three lines should be empty with no indentation
+            // Following lines should have the indentation
+            assert.deepStrictEqual(
+                insertedText,
+                `\n\n\n${' '.repeat(4)}function test() {\n${' '.repeat(8)}console.log("test");\n${' '.repeat(4)}}`
+            )
+        })
+
+        it('handles insertion of code with multiple leading blank lines into empty document at position 0', async () => {
+            const documentContent = ''
+            const document: TextDocument = TextDocument.create('test.py', 'python', 1, documentContent)
+            testFeatures.workspace.getTextDocument.resolves(document)
+
+            const cursorPosition = Position.create(0, 0)
+            const params: InsertToCursorPositionParams = {
+                textDocument: { uri: 'test.ts' },
+                cursorPosition,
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -200,24 +200,42 @@ export class ChatController implements ChatHandlers {
...
+            indent = ' '.repeat(cursorPosition.character)
+            hasVirtualSpace = true
+            cursorPosition.character = 0
         }
 
         let textWithIndent = ''
         params.code.split('\n').forEach((line, index) => {
             if (index === 0) {
+                if (hasVirtualSpace) {
+                    textWithIndent += indent
```
awschristou, line 229: Similar to Line 233, if the first line is blank, it shouldn't receive indentations. (Can this also be added as a test case please?)

ege0zcan: I was under the impression that in the case of a virtual space we actually need this extra indent. 

For a regular insertion (no virtual spaces) we don't add any indentation on the first line since the expectation is to insert the code block at the current cursor position and adding an indentation to the insertion would break the expectation

However, in the case of virtual spaces, the cursor might appear to the user at a certain position due to virtual spaces (e.g., 4 virtual spaces on a line, with the server receiving the cursor position as character 4). In this situation, the server must convert the virtual indentation into a real indentation when inserting the content into the file.

This is what the line in question achieves. To ensure the indentation is applied correctly, the server adjusts the cursor position to character 0 on line 222, so that the real spaces are inserted at the beginning of the line.

awschristou: Here is an arbitrary example for illustrative purposes. My comment above refers to line 81 in the sample image, but I have provided a complete representation of the desired outcome. It's a corner case where the first line is blank - generally this should be uncommon.
image

ege0zcan: Good catch and a nice demonstration, this makes sense to me. I will update the code and continue on adding the test cases

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
```diff
@@ -469,4 +473,336 @@ describe('ChatController', () => {
...
+            // This test will insert an extra for loop inside the existing for loop in the fibonacci function above
+            const forLoop = `for (let i = 2; i <= n; i++) {
+    const next = prev + current;
+    prev = current;
+    current = next;
+}
+`
+            // Given the for loop is inside a function and we will be inserting a new for loop inside, the for loop to be inserted will have 8 prefix spaces
+            const twiceIndentedForLoop = `for (let i = 2; i <= n; i++) {
+${' '.repeat(8)}    const next = prev + current;
```
volodkevych, line 664: looks like it will have 12 not 8 prefix spaces - is it expected?

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -200,24 +200,42 @@ export class ChatController implements ChatHandlers {
...
         const indentRange = {
             start: { line: cursorPosition.line, character: 0 },
             end: cursorPosition,
         }
         const documentContent = await this.#features.workspace.getTextDocument(params.textDocument.uri)
         let indent = documentContent?.getText(indentRange)
+
+        let hasVirtualSpace = false
         if (indent && indent.trim().length !== 0) {
             indent = ' '.repeat(indent.length - indent.trimStart().length)
```
volodkevych, line 214: Not in scope of your change, but I think it will not calculate indent correctly in some corner cases, e.g. if cursor is somewhere within the real indent not after.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -200,24 +200,42 @@ export class ChatController implements ChatHandlers {
...
+            // In this case, the indentation is determined by the cursorPosition
+            this.#log('Indent is nullish and the cursor position is greater than zero while inserting code')
+            indent = ' '.repeat(cursorPosition.character)
+            hasVirtualSpace = true
+            cursorPosition.character = 0
         }
 
         let textWithIndent = ''
         params.code.split('\n').forEach((line, index) => {
             if (index === 0) {
```
volodkevych, line 227: nit: this code will trim the first newline in case the code starts with `/n/n/n`

volodkevych: In general, this code becomes difficult to read with nested conditions. I would think about how to flatten it out (like, move `if(!textWithIndent) {textWithIndent+='\n' + indent}` to top) or split to functions for different pre-conditions (like, with virtual spaces and without).

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
volodkevych: nit: I would also define different variables for cursorIntent and calculatedIntent - it would be more readable and we would not need `else` below.

volodkevych (approved):

PR #697, opened 2025-01-03
feat: pass supplemental contexts only for token client

francescoopiccoli:
## Problem
From version v0.0.14 to v0.0.15 of lsp-codewhisper, we introduced [supplementalContexts](https://github.com/aws/language-servers/pull/463) universally, regardless of whether it's a iam or token client. However, supplementalContext is only accepted by backend when using token authentication. When using iam authentication, the parameter is not accepted, thus we need to introduce authentication specific logic to deal with supplementalContext.
 
bearer token based client: [https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/client/token/bearer-token-service.json#L2707 ](https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/client/token/bearer-token-service.json#L2707)
iam based client: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/client/sigv4/service.json

## Solution
Introducing client specific logic to pass  supplementalContext data

I tested the solution via the CodeWhisperer Server Token debug config, and in the spawned vscode client I tried using inline completion. 

## License


server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -373,15 +373,18 @@ export const CodewhispererServerFactory =
...
                 const requestContext: GenerateSuggestionsRequest = {
                     fileContext,
                     maxResults,
-                    supplementalContexts: supplementalContext?.supplementalContextItems
-                        ? supplementalContext.supplementalContextItems.map(v => ({
-                              content: v.content,
-                              filePath: v.filePath,
-                          }))
-                        : [],
+                    ...(codeWhispererService instanceof CodeWhispererServiceToken && {
```
volodkevych, line 380: I suggest to introduce separate if for this below, `fetchSupplementalContext` only for token-based server and set `requestContext.supplementalContexts`.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -406,7 +409,8 @@ export const CodewhispererServerFactory =
                     classifierResult: autoTriggerResult?.classifierResult,
                     classifierThreshold: autoTriggerResult?.classifierThreshold,
                     credentialStartUrl: credentialsProvider.getConnectionMetadata?.()?.sso?.startUrl ?? undefined,
-                    supplementalMetadata: supplementalContext,
+                    supplementalMetadata:
+                        codeWhispererService instanceof CodeWhispererServiceToken ? supplementalContext : undefined,
```
volodkevych, line 413: Then this check here would not be needed.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
francescoopiccoli: good idea, I implemented it, I utilized `supplementalContextPromise` construct to optimize performance and run in parallel, otherwise some tests were failing (e.g. Case 4. Inflight session is closed by subsequent completion request), due to a higher `timeToFirstRecommendation`  value (from 1260, to 2500)

volodkevych (approved):

PR #716, opened 2025-01-09
fix: make proxy nodejs specific logic available only for nodejs runtime

francescoopiccoli:
## Problem
**this is a short term fix, to fix the webworker bundling. We will implement a better solution very soon.**  
We want to avoid having nodejs specific logic (process.env.VARIABLE) in lsp-codewhisperer as it breaks the webworker bundling, for which `process` is not defined.

## Solution
If/else check to see if we are in node.js env
I tested the changes by testing the proxy

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/client/streamingClient/codewhispererStreamingClient.ts
+++ b/server/aws-lsp-codewhisperer/src/client/streamingClient/codewhispererStreamingClient.ts
@@ -17,8 +17,14 @@ export async function createStreamingClient(
     const creds = credentialsProvider.getCredentials('bearer')
 
     let clientOptions
-    const proxyUrl = process.env.HTTPS_PROXY ?? process.env.https_proxy
-    const certs = process.env.AWS_CA_BUNDLE ? [readFileSync(process.env.AWS_CA_BUNDLE)] : undefined
+    // short term solution to fix webworker bundling, broken due to this node.js specific logic in here
+    const isNodeJS: boolean = typeof process !== 'undefined' && process.release && process.release.name === 'node'
+    const proxyUrl = isNodeJS ? (process.env.HTTPS_PROXY ?? process.env.https_proxy) : undefined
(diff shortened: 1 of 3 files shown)


francescoopiccoli:
There is an occurrence of `process.env.HTTPS_PROXY`also in `servers/aws-lsp-identity/srs/sso`, I am not sure if I should implement the fix there to, as it does not affect the webworker bundling.

server/aws-lsp-codewhisperer/src/client/streamingClient/codewhispererStreamingClient.ts
```diff
@@ -17,8 +17,14 @@ export async function createStreamingClient(
     const creds = credentialsProvider.getCredentials('bearer')
 
     let clientOptions
-    const proxyUrl = process.env.HTTPS_PROXY ?? process.env.https_proxy
-    const certs = process.env.AWS_CA_BUNDLE ? [readFileSync(process.env.AWS_CA_BUNDLE)] : undefined
+    // short term solution to fix webworker bundling, broken due to this node.js specific logic in here
+    const isNodeJS: boolean = typeof process !== 'undefined' && process.release && process.release.name === 'node'
```
volodkevych, line 21: Why don't we pass `HTTPS_PROXY` and `AWS_CA_BUNDLE` as a config value from the runtime - the same way as we started doing for other env vars? It would not be that complex, but more clean.

francescoopiccoli: As this is very short term, I opted for the easiest solution that did the job, imo a more clean solution makes sense if it stays in the codebase indefinitely. Passing down via runtime is another option, not that complex but more work than this for sure (see https://github.com/aws/language-servers/pull/711/files). 
If you have a strong preference for the runtime option, I am good with implementing that solution here as well, let me know!

imykhai (approved):

PR #717, opened 2025-01-13
perf: dispose chat controller in chatController tests

francescoopiccoli:
## Problem
When running `npm run test` in `server/aws-lsp-codewhisperer`, all tests pass but the test task does not finish/close because we forgot to dispose some resources we created, in particular it's about the `chatController` object.

## Solution
Dispose via` chatController.dispose()`

## License


ege0zcan (approved):

ege0zcan:
Nice find

PR #725, opened 2025-01-15
fix: convert makeProxyConfig to sync to allow proxy configs to be loa…

pras0131:
convert makeProxyConfig function to sync as to allow proxy configurations to be loaded before being used. There has been an issue where proxy was not working for inline completions as the configurations were being loaded asyncronously and when the client was setting the configs, they were not set. The change aims to fix the issue and it has been tested on local environment to be working fine using Charles proxy for inline completions.

[[ THE CHANGE IS DEPENDENT ON THIS PR https://github.com/aws/language-server-runtimes/pull/299/files]

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/utils.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/utils.ts
@@ -131,7 +131,7 @@ export function getEndPositionForAcceptedSuggestion(content: string, startPositi
     return endPosition
 }
 
-export const makeProxyConfig = async (workspace: Workspace) => {
+export const makeProxyConfig = (workspace: Workspace) => {
     let additionalAwsConfig: ConfigurationOptions = {}
     // short term solution to fix webworker bundling, broken due to this node.js specific logic in here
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/utils.ts
```diff
@@ -140,7 +140,7 @@ export const makeProxyConfig = async (workspace: Workspace) => {
     if (proxyUrl) {
         const certs = isNodeJS
             ? process.env.AWS_CA_BUNDLE
-                ? [await workspace.fs.readFile(process.env.AWS_CA_BUNDLE)]
+                ? [workspace.fs.readFileSync(process.env.AWS_CA_BUNDLE).toString()]
```
natdurlik, line 143: Do we need to have a separate method for synchronous file read? Have you tried solutions with using 'readFile' async in a factory create method instead of the constructor?

pras0131: I found this as a clean way to do it. Factory create instead of constructor is not required here and it would lead to a lot of refactoring, which might not be required.

Issue #724, opened 2025-01-15
Q/codewhisperer does not support LSP standard "textDocument/completion" method


justinmk3:
# Problem

The LSP standard `textDocument/completion` method is not supported. Inline suggestions are currently provided by the custom `aws/textDocument/inlineCompletionWithReferences` method.

This means that:

- LSP clients have to build custom integration just to get suggestions from the Amazon Q LSP, even though that is a standard feature of LSP.
- in clients like Nvim or Emacs, every *completion plugin* must *also* be manually integrated.

# Expected behavior

Inline suggestions should be exposed from the standard LSP `textDocument/completion` (and related) methods, where possible. At least in some minimal/degraded form, if it's not possible to provide the full features.

PR #739, opened 2025-01-21
feat(amazonq): implement basic server for standard LSP completion API

viktorsaws:
## Problem
Q/codewhisperer does not support LSP standard "textDocument/completion" method https://github.com/aws/language-servers/issues/724

## Solution

- Added simple experimental `AmazonQCompletionServer` implementation to support `textDocument/completion` protocol. More docs available in `server/aws-lsp-codewhisperer/src/experimental/README.md`.
- Added new testing app in `app/aws-lsp-q-dev-completions`, which includes `AmazonQCompletionServer` and device SSO token provider among other servers.

To test the server: fetch this PR or https://github.com/viktorsaws/language-servers/tree/experimental_lsp_completions branch and rebuild the project:
```
npm install && npm run package
```

Bundled language server is available at `./app/aws-lsp-q-dev-completions/build/amazon-q-completions-and-sso.js` path.

6-fcf9be50f3

## License

### DIFF
--- a/app/aws-lsp-q-dev-completions/webpack.config.js
+++ b/app/aws-lsp-q-dev-completions/webpack.config.js
@@ -1,0 +1,41 @@
+var path = require('path')
+
+const baseConfig = {
+    mode: 'development',
+    output: {
+        path: __dirname,
+        filename: 'build/[name].js',
+        globalObject: 'this',
+        library: {
+            type: 'umd',
+        },
+    },
+    resolve: {
+        extensions: ['.ts', '.tsx', '.js'],
+    },
+    module: {
+        rules: [
+            {
+                test: /\.tsx?$/,
+                use: 'ts-loader',
+                exclude: /node_modules/,
+            },
+        ],
+    },
+}
+
+const completionsWithDeviceSsoConfig = {
+    ...baseConfig,
+    experiments: {
+        asyncWebAssembly: true,
+    },
+    entry: {
+        'amazon-q-completions-and-sso': path.join(__dirname, 'src/completions-with-device-sso.ts'),
+    },
+    resolve: {
+        ...baseConfig.resolve,
+    },
+    target: 'node',
+}
+
+module.exports = [completionsWithDeviceSsoConfig]
--- a/app/aws-lsp-q-dev-completions/README.md
+++ b/app/aws-lsp-q-dev-completions/README.md
@@ -1,0 +1,24 @@
+# Amazon Q Completions and Device SSO Token server
+
+This package provides example configuration to produce Amazon Q server implementation bundled with [Language Server Standalone Runtime](https://github.com/aws/language-server-runtimes).
+
+Language Server contains:
+- `SsoAuthServer`: support for Device authentication SSO flow from [`amzn/device-sso-auth-lsp`](../../server/device-sso-auth-lsp/) package.
+- `AmazonQCompletionServerToken`: experimental implementation of [Amazon Q Developer Code Completions](../../server/aws-lsp-codewhisperer/src/experimental/README.md) with standard LSP `textDocument/completion` protocol.
+- `CodeWhispererServerTokenProxy`: support of Amazon Q `InlineCompletionWithReferences`.
+- `QChatServerProxy`: support of Amazon Q Chat server.
+
+## Installation and usage
+
+To create compiled bundle run:
+```bash
+npm run package
+```
+
+This command will compile package and produce bundled Javascript server in `./out/` directory: 
+- `./out/completions-with-device-sso.js` - Amazon Q server bundled application.
+
+To verify compiled bundle can run, you can start it in your shell with NodeJS:
+```bash
+node ./out/completions-with-device-sso.js --stdio
+```
--- a/app/aws-lsp-q-dev-completions/package.json
+++ b/app/aws-lsp-q-dev-completions/package.json
@@ -1,0 +1,22 @@
+{
+    "name": "@amzn/aws-lsp-q-dev-completions",
+    "version": "0.0.1",
+    "description": "Amazon Q Developer Code Completions Language Server",
+    "main": "out/index.js",
+    "scripts": {
+        "clean": "rm -rf out/ bin/ tsconfig.tsbuildinfo",
+        "compile": "tsc --build",
+        "package": "npm run compile && npm run webpack",
+        "webpack": "webpack"
+    },
+    "dependencies": {
+        "@aws/language-server-runtimes": "^0.2.31",
+        "@aws/lsp-codewhisperer": "*",
+        "@amzn/device-sso-auth-lsp": "*"
+    },
+    "devDependencies": {
+        "ts-loader": "^9.4.4",
+        "webpack": "^5.94.0",
+        "webpack-cli": "^5.1.4"
+    }
+}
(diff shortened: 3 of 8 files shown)


server/aws-lsp-codewhisperer/src/experimental/README.md
```diff
@@ -0,0 +1,26 @@
+# AmazonQCompletionServer
```
justinmk3, line 1: given the layout of the rest of this project, the experimental/ directory seems unintuitive. how does it map to the existing pattern of the repo (which has `client/` and `language-server/` at the top level)?

app/aws-lsp-q-dev-completions/webpack.config.js
```diff
@@ -0,0 +1,41 @@
+var path = require('path')
+
+const baseConfig = {
```
justinmk3, line 3: why doesn't this inherit from some other part of this repo? does each "server" have its own webpack config?

server/aws-lsp-codewhisperer/src/experimental/amazonQCompletionServer.ts
```diff
@@ -0,0 +1,333 @@
...
+        return {
+            itemId: suggestion.itemId,
+            insertText: insertText,
+            range,
+            references: references?.length ? references : undefined,
+        }
+    })
+}
+
+export const AmazonQCompletionServer =
```
justinmk3, line 103: why isn't this part of the existing [codewhispererserver.ts](https://github.com/aws/language-servers/blob/9f447b2ca6a1a08242c591c1bdf91e1d3a9def3d/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts) ? can clients opt-in to the experiment by sending a `settings` flag on init?

app/aws-lsp-q-dev-completions/webpack.config.js
viktorsaws: Yeah, every package like this in `app` directory could be built using it's own config. Moving it out to common shared config would be an optimization, I decided to keep it more standalone for now.

server/aws-lsp-codewhisperer/src/experimental/README.md
viktorsaws: Yeah, I was not sure where to place it, given it's simple implementation. Moving to `language-server` would be more appropriate.

server/aws-lsp-codewhisperer/src/experimental/amazonQCompletionServer.ts
viktorsaws: I wanted to avoid polluting already overloaded [codewhispererserver.ts](https://github.com/aws/language-servers/blob/9f447b2ca6a1a08242c591c1bdf91e1d3a9def3d/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts) with this implementation. I'd look into toggling this completion behind a flag.

justinmk3: > I wanted to avoid polluting already overloaded [codewhispererserver.ts](https://github.com/aws/language-servers/blob/9f447b2ca6a1a08242c591c1bdf91e1d3a9def3d/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts)

I'm more worried about the pollution of numerous duplicate modules, configs, sub-packages. That's much harder to cognitively consume than 1 package with lots of code (which can use shared base-classes/modules).

But maybe I missed what you meant :)

server/aws-lsp-codewhisperer/src/experimental/amazonQCompletionServer.ts
```diff
@@ -0,0 +1,333 @@
...
+                            label: label,
+                            // insertText: item.insertText as string,
+                            textEdit: {
+                                range: {
+                                    start: params.position,
+                                    end: params.position,
+                                },
+                                newText: item.insertText as string,
+                            },
+                            kind: CompletionItemKind.Text,
```
justinmk3, line 278: Discussion with Neovim plugin authors leads me to believe that "snippet" is most appropriate for multiline completions. 


```suggestion
                            kind: CompletionItemKind.Snippet,
```

server/aws-lsp-codewhisperer/src/experimental/amazonQCompletionServer.ts
kmile: Can we just support both approaches at the same time? Clients that send us `textdocument/completion` requests will get answers with code-with-references filtered out. Clients that ask for a richer response including references, will get the enhanced response. They could use exactly the same handler without any duplication. We only need a small additional filter and transformation to make sure the `params` and `suggestions` match the right shape.

Are we worried clients might be sending both at the same time? They shouldn't, at least not to the same LSP server. While we could try to avoid this behavior with turning capabilities on/off dynamically, or adding some additional settings, it seems to me that the LSP server is responsible for answering client's requests, not determining whether the client's request is valid or not.

kmile: Would this (untested) code work instead?

```
        const onCompletionHandler = async (params: CompletionParams, token: CancellationToken) => {
            const inlineCompletionParams: InlineCompletionWithReferencesParams = {
                ...params,
                context: {
                    ...params.context,
                    triggerKind: InlineCompletionTriggerKind.Invoked,
                },
            }

            const inlineCompletionResult = await onInlineCompletionHandler(inlineCompletionParams, token)

            return {
                isIncomplete: false,
                items: inlineCompletionResult.items.filter(item => !item.references).map(item => {
                    return {
                        label: item.insertText.toString()
                    }
                }),
            }
        }
```

server/aws-lsp-codewhisperer/src/experimental/amazonQCompletionServer.ts
justinmk3: Actually based on feedback from blink.nvim users, `usekind: CompletionItemKind.Snippet` seems to work better for them, while `insertTextFormat: InsertTextFormat.Snippet` risks activating snippet "placeholder" handling (snippet mode).

server/aws-lsp-codewhisperer/src/experimental/amazonQCompletionServer.ts
viktorsaws: My main concern with using same handler as for `inlineSuggestionWithReferences` is that it's stateful - we have notion of shared completion session shared between Client and Server. Session lifecycle has been already quite complicated, and adding extra logic to handle standard completion could overcomplicate things.

I think it would be better alternative to move this separate handler to `CodeWhispererServer`, and reuse shared handler and LSP configuration state, as Justin suggested, but I would avoid mixing 2 calls.

server/aws-lsp-codewhisperer/src/experimental/amazonQCompletionServer.ts
viktorsaws: I'm not sure if using Snippet type would be correct, given it expects that returned text has richer syntax, defined in LSP https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax. 

I don't have clear overview of how different IDE/LSP Client handle Snippets, I suspect it might cause issues in some clients.

viktorsaws:
Need to implement support for cancellationTokens before merging

server/aws-lsp-codewhisperer/src/experimental/amazonQCompletionServer.ts
justinmk3: Further feedback from Neovim users has indicated that `Text` or `Reference` work better, because completion engines often filter-out `Snippet` (and sometimes `Text`, too).

PR #740, opened 2025-01-22
chore: improve vSCode client debug instructions and launch.json config

francescoopiccoli:
## Problem
Current debugging instructions (with breakpoints) are not complete/precise. In this PR, we specify step by step how to debug with breakpoints.
We also fix the task value set for the preLaunchTask option for the different configurations.
## License

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -24,7 +24,7 @@
             "env": {
                 "LSP_SERVER": "${workspaceFolder}/app/aws-lsp-buildspec-runtimes/out/index.js"
             },
-            "preLaunchTask": "npm: compile"
+            "preLaunchTask": "compile"
         },
         {
             "name": "CloudFormation Server",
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -24,7 +24,7 @@
             "env": {
                 "LSP_SERVER": "${workspaceFolder}/app/aws-lsp-buildspec-runtimes/out/index.js"
             },
-            "preLaunchTask": "npm: compile"
```
ege0zcan, line 27: could deleting this single line(https://github.com/aws/language-servers/blob/1c38b0c4b9f462df64e3ee74b8d8b196ae59dcc0/.vscode/tasks.json#L5) potentially have the same effect in fixing the issue?

francescoopiccoli: fixing the debugging issue you mean? or the failing npm: compile command?

ege0zcan: Both approaches are fine in my opinion, I was curious if there was a difference between the two approaches

ege0zcan (approved):

PR #742, opened 2025-01-23
refactor: change identity server to use core/runtimes server types and add noImplicitOverride

floralph:
## Problem

The identity server was not fully migrated to reuse core/runtimes server related types.  Abstract init methods on serverBase required overrides even if not needed.

## Solution

Removed redundant server types from identity server package and migrated code to reuse those types in core/runtimes.  Changed serverBase to provide default no-op impls of init methods and set noImplicitOverride (including updating a few existing servers that required the `override` keyword).

## License


ege0zcan (approved):

justinmk3 (approved):

justinmk3:
This is a "refactor", not a "chore"

PR #743, opened 2025-01-23
fix: move CW streaming client to tarball

natdurlik:
## Problem
We currently have dependency issues with codewhisperer-streaming client. Compiling step fails when package-lock.json is deleted.

## Solution
Packed streaming client compiled code into tarball (`npm pack`) and deleted source code from the repo.

## License


imykhai (approved):

imykhai:
As far as I understand, with this PR you should also include codewhisperer-streaming as a bundled dependency into the lsp-codewhisperer package.json 

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#bundledependencies

ege0zcan (approved):

PR #746, opened 2025-01-24
fix: removing duplicated nuget packages folder

pranav-firake:
## Problem
V1642391964
Currently when we are packaging artifacts.zip on the IDE, the nuget packages are part of 

1. default "packages" folder in the root solution/ source Code
2. newly created the references folder in artifactsWorkspace. 

This is causing increased artifacts size.

## Solution

 we are not using packages folder from sourceCode for transformations in Qnet CLI, we are only using nuget packages from references folder. 
So we can safely delete packages folder before zipping to save on space.

Simple example on saving space is 
Sample applications - Bobs Bookstore
Size of packages folder - 241MB

size of Artifacts zip BEFORE below changes of deduping - 179 MB

image

size of Artifacts zip AFTER below changes of deduping - 89 MB
image

we are saving almost 45% space in artifacts zipping and thereby for uploads. This will also benefit for customer issues with large codebase 

### UPDATE :
Also found that copyFile is async and we are zipping / deleting package folder even before all copying is complete.  so changed copyFile to Sync operation

## Testing 

Case : Updated a BobsBookstore application to bloat to 2.3 GB of artifacts.zip. 
with deduping, for same project artifacts size is - 1.2 GB

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
@@ -209,15 +209,15 @@ export class ArtifactManager {
     copyFile(sourceFilePath: string, destFilePath: string) {
         const dir = path.dirname(destFilePath)
         this.createFolderIfNotExist(dir)
-        fs.copyFile(sourceFilePath, destFilePath, error => {
-            if (error) {
-                if (!fs.existsSync(dir) && dir.includes(packagesFolderName)) {
-                    //Packages folder has been deleted to avoid duplicates in artifacts.zip
-                    return
-                }
-                this.logging.log('Failed to copy: ' + sourceFilePath + error)
+        try {
+            fs.copyFileSync(sourceFilePath, destFilePath)
+        } catch (err) {
+            if (!fs.existsSync(dir) && dir.includes(packagesFolderName)) {
+                //Packages folder has been deleted to avoid duplicates in artifacts.zip
+                return
             }
-        })
+            this.logging.log(`Failed to copy from ${sourceFilePath} and error is ${err}`)
+        }
     }
 
     calculateMD5Sync(filePath: string): string {
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
@@ -9,6 +9,7 @@ const artifactFolderName = 'artifact'
 const referencesFolderName = 'references'
 const zipFileName = 'artifact.zip'
 const sourceCodeFolderName = 'sourceCode'
+const packagesFolderName = 'packages'
 
 export class ArtifactManager {
     private workspace: Workspace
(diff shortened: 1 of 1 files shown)


marknfawaz:
Please add how this was tested.

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -41,6 +43,21 @@ export class ArtifactManager {
...
 
+    async removeDuplicateNugetPackagesFolder(request: StartTransformRequest) {
+        const packagesFolder = path.join(
+            this.workspacePath,
+            artifactFolderName,
+            sourceCodeFolderName,
+            packagesFolderName
+        )
+        if (!fs.existsSync(packagesFolder)) {
+            this.logging.log('Cannot find nuget packages folder in source code')
```
marknfawaz, line 54: Why do we need to log this?

jonlouie: +1. If we invert the `if` logic, we can simplify this method and reducing the logging becomes natural.

pranav-firake: agreed, it is causing unnecessary logs, will remove this. 

purpose for this if block was - 
copyFIle is async operation, so some copy function still goes on even after deleting packages folder. causing errors with packages/abc.pkg ENOENT errors. will add just return here instead of logging

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -41,6 +43,19 @@ export class ArtifactManager {
...
+    async removeDuplicateNugetPackagesFolder(request: StartTransformRequest) {
+        const packagesFolder = path.join(
+            this.workspacePath,
+            artifactFolderName,
+            sourceCodeFolderName,
+            packagesFolderName
+        )
+        if (fs.existsSync(packagesFolder)) {
+            fs.rmSync(packagesFolder, { recursive: true, force: true })
+            this.logging.log(`Removed nuget packages folder: ${packagesFolder}`)
```
jonlouie, line 55: One more thing: In the log message, we should say why it is removed (e.g. `"Removed packages folder {packagesFolder} from directory containing artifacts to be uploaded because it is a duplicate of {nugetFolder}"`)

pranav-firake: thanks for suggestion, updated logging message accordingly

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -41,6 +43,21 @@ export class ArtifactManager {
         }
     }
 
+    async removeDuplicateNugetPackagesFolder(request: StartTransformRequest) {
```
ege0zcan, line 46: as a potential optimization, is it possible to skip adding these files to the zip in the first place. If I understand correctly right now we spend extra time copying these files in the step before only to end up going through the folder again and deleting them. If we could create the artifacts in a single go that would also make the process more efficient. Is this possible?

ege0zcan:
Small reminder to remove the `[internal-review]` prefix before merging, otherwise the commit linter rules will prevent the merge

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
pranav-firake: yes Ege, thanks for suggestion. 
After this hotfix, we will take a stab in detail on optimizing entire zipping with ticket P194692593.

PR #749, opened 2025-01-24
feat: bugfix for artifacts upload

xueningl-az:
## Problem
Node limit for 2 GiB for fs.readFile .  ‘File size is greater than 2 GiB’ is common error in node with fs.readFile. 
Reference https://github.com/nodejs/node/issues/55864 

We are using fs.readFile in 3 places in language server code. 

## Solution

use fs.createReadStream instead.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
@@ -75,7 +75,7 @@ export class ArtifactManager {
             await sourceCodeFilePaths.forEach(async filePath => {
                 try {
                     this.copySourceFile(request.SolutionRootPath, filePath)
-                    var contentHash = await this.calculateMD5Sync(filePath)
+                    var contentHash = await this.calculateMD5Async(filePath)
                     var relativePath = this.normalizeSourceFileRelativePath(request.SolutionRootPath, filePath)
                     codeFiles.push({
                         contentMd5Hash: contentHash,
@@ -136,7 +136,7 @@ export class ArtifactManager {
     //use fs.createReadStream() to handle the file in smaller, manageable chunks rather than loading the entire file
     // into memory. This avoids hitting the 2 GiB limit that occurs when using fs.readFile(),
     // as it loads the entire file into memory.
-    static async getSha256(fileName: string): Promise<string> {
+    static async getSha256Async(fileName: string): Promise<string> {
         const hasher = crypto.createHash('sha256')
         const stream = fs.createReadStream(fileName)
 
@@ -212,17 +212,14 @@ export class ArtifactManager {
         })
     }
 
-    async calculateMD5Sync(filePath: string): Promise<string> {
+    async calculateMD5Async(filePath: string): Promise<string> {
         try {
             const hash = crypto.createHash('md5')
-            //const stream = fs.createReadStream(filePath, { highWaterMark: 64 * 1024 }); // 64 KB chunks
             const stream = fs.createReadStream(filePath)
             for await (const chunk of stream) {
                 hash.update(chunk)
             }
 
-            //const data = fs.readFileSync(filePath)
-            //const hash = crypto.createHash('md5').update(data)
             return hash.digest('hex')
         } catch (error) {
             this.logging.log('Failed to calculate hashcode: ' + filePath + error)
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -128,9 +132,18 @@ export class ArtifactManager {
         return zipPath
     }
 
-    static getSha256(fileName: string) {
+    //To read large files in chunks (greater than 2 GiB) using Node.js,
+    //use fs.createReadStream() to handle the file in smaller, manageable chunks rather than loading the entire file
+    // into memory. This avoids hitting the 2 GiB limit that occurs when using fs.readFile(),
+    // as it loads the entire file into memory.
+    static async getSha256(fileName: string): Promise<string> {
```
jonlouie, line 139: Rename to `getSha256Async`

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -199,10 +212,17 @@ export class ArtifactManager {
         })
     }
 
-    calculateMD5Sync(filePath: string): string {
+    async calculateMD5Sync(filePath: string): Promise<string> {
```
jonlouie, line 215: 1. Rename to `calculateMD5Async`
2. Do we need to `await` any calls to this function now that it is async?

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -150,8 +150,9 @@ export class TransformHandler {
     async uploadArtifactToS3Async(fileName: string, resp: CreateUploadUrlResponse, sha256: string) {
         const headersObj = this.getHeadersObj(sha256, resp.kmsKeyArn)
         try {
+            const fileStream = fs.createReadStream(fileName)
             const response = await got.put(resp.uploadUrl, {
-                body: fs.readFileSync(fileName),
+                body: fileStream,
```
kanikaul-amazon, line 155: Does this work? Seems like got has a separate API for streams

pranav-firake:
CI build is failing with message `AssertError: expected readFileSync to be called once but was called 0 times`
we will need to update test too

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
xueningl-az: This is the only place

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
xueningl-az: it's working for stream.

jonlouie (approved):

pranav-firake (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -128,9 +132,18 @@ export class ArtifactManager {
         return zipPath
     }
 
-    static getSha256(fileName: string) {
+    //To read large files in chunks (greater than 2 GiB) using Node.js,
+    //use fs.createReadStream() to handle the file in smaller, manageable chunks rather than loading the entire file
+    // into memory. This avoids hitting the 2 GiB limit that occurs when using fs.readFile(),
+    // as it loads the entire file into memory.
+    static async getSha256Async(fileName: string): Promise<string> {
```
ege0zcan, line 139: As a small optimization, is it possible/sensible to calculate the sha256 and md5 of files at the same time and store these details in the artifact manager, that way we can get rid of some extra overhead by reducing the number of times `fsCreateReadStream` gets called on a single file. 

If the timing of these hash calculations matters and if such an optimization is not feasible, feel free to ignore this comment

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
justinmk3: got is not browser-compatible. In the vscode extension we settled on `cross-fetch`, and are using a wrapper so that modules can avoid direct dependencies on whatever package we use: https://github.com/aws/aws-toolkit-vscode/blob/2200b1544d0c0c3e250b8ae2dc16d5f3e87c4270/packages/core/src/shared/request.ts#L6

PR #775, opened 2025-02-05
chore: fix failing tests for windows github actions

dogusata:
## Problem
Windows tests are failing (because of [breaking github action env. changes](https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/))

## Solution
Updated the tests to create a proper fileURL from any file uri.

## License


server/aws-lsp-codewhisperer/src/language-server/utilities/supplementalContextUtil/crossFileContextUtil.ts
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable import/no-nodejs-modules */
```
ege0zcan, line 1: I think we shouldn't disable eslint, for now unless they make the tests fail. The rule is configured to give warnings anyway which are useful reminders

dogusata: ?

PR #780, opened 2025-02-12
refactor: simplify initialize overrides in ServerBase and servers extending it

floralph:
## Problem
`initialize` method on ServerBase was not async and required implementations to handle catching errors and returning to meet the LSP return type of `HandlerResult<..., InitializeError>`.

Other changes:
* Added ServerState enum to ServerBase to support NotificationsServer
* Cleaned up showNotification impl in NotificationServer
* Fixed typo in CriteriaFilteringFetcher

## Solution
Simplified implementation of `initialize` method so error handling isn't necessary (handled by wrapper in ServerBase) and async code in initialize is now possible.  Updated IdentityServer for the initialize function declaration change.

Tested with `npm run test` and using vscode sample client IdentityServer function to see various identity server functions called/return values.

image

## License

### DIFF
--- a/server/aws-lsp-identity/src/language-server/identityServer.ts
+++ b/server/aws-lsp-identity/src/language-server/identityServer.ts
@@ -96,7 +96,7 @@ export class IdentityServer extends ServerBase {
             ...result,
             ...{
                 serverInfo: {
-                    name: 'AWS Toolkit Language Server for Identity',
+                    name: 'Language Server for AWS Identity Management',
                 },
             },
         }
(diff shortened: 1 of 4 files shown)


server/aws-lsp-identity/src/language-server/identityServer.ts
```diff
@@ -94,7 +92,14 @@ export class IdentityServer extends ServerBase {
 
         this.disposables.push(autoRefresher)
 
-        return { capabilities: {} }
+        return {
+            ...result,
+            ...{
+                serverInfo: {
+                    name: 'AWS Toolkit Language Server for Identity',
```
viktorsaws, line 99: nit: I imaging this server is not coupled to Toolkits and can be used by any client. Would more generic name be better, ex. "Language Server for AWS Identity Management"?

viktorsaws (approved):

aseemxs (approved):

PR #784, opened 2025-02-14
fix: prevent override of client request listeners in CodeWhispererServiceIAM

francescoopiccoli:
## Problem
Prevent override of client request listeners in CodeWhispererServiceIAM in case any listener was setup before. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
@@ -7,6 +7,7 @@ import {
 } from '@aws/language-server-runtimes/server-interface'
 import { AWSError, ConfigurationOptions, CredentialProviderChain, Credentials } from 'aws-sdk'
 import { PromiseResult } from 'aws-sdk/lib/request'
+import { Request } from 'aws-sdk/lib/core'
 import { v4 as uuidv4 } from 'uuid'
 import {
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -90,8 +91,13 @@ export class CodeWhispererServiceIAM extends CodeWhispererServiceBase {
             ]),
         }
         this.client = createCodeWhispererSigv4Client(options, sdkInitializator)
-        this.client.setupRequestListeners = ({ httpRequest }) => {
-            httpRequest.headers['x-amzn-codewhisperer-optout'] = `${!this.shareCodeWhispererContentWithAWS}`
+        // Avoid overwriting any existing client listeners
+        const clientRequestListeners = this.client.setupRequestListeners
+        this.client.setupRequestListeners = (request: Request<unknown, AWSError>) => {
```
viktorsaws, line 96: Do we need to do this with caching original listeners? Would using similar implementation as for `CodeWhispererServiceToken` work https://github.com/aws/language-servers/blob/aad4188b0b2dcedb54e21a7396beb854759f48ec/server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts#L145-L154?

imykhai (approved):

PR #814, opened 2025-03-03
feat(amazonq): centralize access to Q Service SDK instance and add support for Q Developer profiles

viktorsaws:
## Extended SSO connection and Amazon Q Developer Profiles management for Amazon Q Language Server

This PR introduces few major changes to handling connection, managing Codewhisperer Service and adds support for Q Developer profiles. Amazon Q Developer Profile is a new feature of Amazon Q, which allows users to configure which region is used for Amazon Q Service calls.

Key changes:

1. Introducing centralized CodeWhisperer Service instance management for SSO/Bearer Token based SDK client in [AmazonQTokenServiceManager.ts](https://github.com/aws/language-servers/pull/814/files#diff-b2a0c98ce7c15bf527343df014511e197074b03acfb519477967d78fcf3e333d).
2. Adding support for new `UpdateConfiguration` [Runtimes protocol](https://github.com/aws/language-server-runtimes/pull/350) for setting Amazon Q Developer Profile selection for IdentityCenter SSO connection.
3. Extending CodeWhisperer InlineCompletion server to use new Service manager. Extending `InlineCompletion` API handlers to return LSP Error responses with specific `awsErrorCode` in error data payload.

### AmazonQTokenServiceManager

AmazonQTokenServiceManager is a singleton class, that centralizes handling of current SSO connection, maintains single instance of CodeWhisperer SDK service client, and ensures that CodeWhisperer Client is up to date with current active Q Region selection, when enabled.

AmazonQTokenServiceManager will be used in Q Server implementations, which make calls to CodeWhisperer service. It exposes `AmazonQTokenServiceManager.getCodewhispererService()` method, which returns current active CodeWhisperer service client and ensures that client is correctly configured with current active SSO Token and region. [CodeWhispererServer](https://github.com/aws/language-servers/pull/814/files#diff-41618063d6869da8a8fcb2326f67e08e197e7a04c01cfab0b832996e320c653cR262-R285) is updated to use AmazonQTokenServiceManager for Token type of Service. IAM based Service token is not changed.

AmazonQTokenServiceManager subscribes to LSP changeConfiguration events to handle configuration changes by client and maintain CodeWhisperer Service configs in sync.

### Q Language Server API changes

1. New handler [UpdateConfiguration](https://github.com/aws/language-server-runtimes/blob/137654e610c1290a476e99fcb82da9af52abcbbd/runtimes/protocol/configuration.ts#L41-L47) request to select Q Developer Profile for IdentityCenter connection. Parameters:
```
params = {
    section: "aws.q",
    settings: {
        profileArn: $profileArn
    }
}
```

Response: void if change successful. 
Error: LSP ErrorResponse with `awsErrorCode` payload.

2. InlineCompletionWithReferences API may return LSP ResponseErrors with `awsErrorCode` payload.

#### AWS Error Codes
Calls to Q LSP may return LSP ResponseError with extended `data` payload
```
interface ResponseError {
	code: integer;
	message: string;
	data?: {
                awsErrorCode: AmazonQAwsErrorCode;
        }
}
```

Error code in `awsErrorCode` will indicated specific error occurred, based on state of CodeWhisperer Service connection. No payload may be returned in cases of regular request handling. Client should handle error codes, and may change it's own state or take action based on server error state

New `awsErrorCode` types introduced:
```
AmazonQAwsErrorCode:
    E_AMAZON_Q_NOT_INITIALIZED - CodeWhisperer service was not properly initialized.
    E_AMAZON_Q_PENDING_CONNECTION - Waiting for client to send SSO Bearer Token and expected SSO StartUrl
    E_AMAZON_Q_PENDING_PROFILE - (Only for IdentityCenter) waiting for client to send Q Developer profile selection
    E_AMAZON_Q_PENDING_PROFILE_UPDATE - (Only for IdentityCenter) waiting for server to finish Developer profile update
    E_AMAZON_Q_INVALID_PROFILE - (Only for IdentityCenter) Developer Profile is not valid
    E_AMAZON_Q_CONNECTION_NO_PROFILE_SUPPORT - (Only for no connection or BuilderId) current connection does not support Developer Profiles
```

### TODO
- [x] Tests
- [x] PR Description
- [x] E2E test scenarios

## License


server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -308,7 +325,7 @@ export const CodewhispererServerFactory =
         const onInlineCompletionHandler = async (
             params: InlineCompletionWithReferencesParams,
             token: CancellationToken
-        ): Promise<InlineCompletionListWithReferences> => {
+        ): Promise<InlineCompletionListWithReferences | AmazonQLspResponseError> => {
```
viktorsaws, line 328: TODO: Update protocol to support returning errors

server/aws-lsp-codewhisperer/package.json
```diff
@@ -18,6 +18,7 @@
         "copyServiceClient": "copyfiles -u 1 --error ./src/client/sigv4/*.json out && copyfiles -u 1 --error ./src/client/token/*.json out",
         "fix": "npm run fix:prettier",
         "fix:prettier": "prettier . --write",
+        "format": "npm run fix:prettier",
```
volodkevych, line 21: I so not think we need different names for the same scripts.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -276,6 +276,14 @@ export class CodeWhispererServiceToken extends CodeWhispererServiceBase {
         return this.client.listAvailableCustomizations(request).promise()
     }
 
+    async listAvailableProfiles() {
```
volodkevych, line 279: Why do we need it here?

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -249,16 +251,41 @@ export const CodewhispererServerFactory =
...
+        if (serviceType === 'CodeWhispererServiceToken') {
+            AmazonQServiceManager = AmazonQTokenServiceManager.getInstance({
+                lsp,
+                logging,
+                credentialsProvider,
+                sdkInitializator,
+                workspace,
+                runtime,
+            })
+            credentialsType = 'bearer'
```
volodkevych, line 274: Can we avoid hardcoding this?

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -407,7 +443,7 @@ export const CodewhispererServerFactory =
                     customizationArn: undefinedIfEmpty(codeWhispererService.customizationArn),
                 })
 
-                return codeWhispererService
+                return AmazonQServiceManager.getCodewhispererService()
```
volodkevych, line 446: I think we need to use one instance of service throughout one handler.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -268,7 +295,12 @@ export const CodewhispererServerFactory =
         )
 
         lsp.addInitializer((params: InitializeParams) => {
-            codeWhispererService.updateClientConfig({
+            AmazonQServiceManager.updateClientConfig({
```
volodkevych, line 298: As an option: Can we cache initParams in runtime and expose it as a feature to all the servers? Then we would not need this `updateClientConfig`.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -268,7 +295,12 @@ export const CodewhispererServerFactory =
...
+                userAgent: getUserAgent(params, runtime.serverInfo),
+            })
+
+            // TODO: Review configuration options expected in other features
+            fallbackCodeWhispererService.updateClientConfig({
```
volodkevych, line 303: I would prefer to hide `fallbackCodeWhispererService` behind `AmazonQServiceManager`. Given all Q servers will now depend on `AmazonQServiceManager`, we should make it simple to use.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,403 @@
...
+        return
+    }
+
+    public getCodewhispererService(): CodeWhispererServiceToken {
+        // Prevent initiating requests while profile change is in progress.
+        if (this.state === 'PENDING_Q_PROFILE_UPDATE') {
+            throw new AmazonQServicePendingProfileUpdateError()
+        }
+
+        this.handleSsoConnectionChange()
```
volodkevych, line 329: Is there any specific reason why we need to call it here? I would imaging we just need to reset `this.state` to `PENDING_CONNECTION` and remove cached service instance whenever `creds/.../remove` method is called.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,403 @@
...
+            return
+        }
+
+        this.logging?.log(`Amazon Q: switching service client region from ${oldRegion} to ${newRegion}`)
+
+        // Selected new profile is in different region. Re-initialize service and terminate inflight requests
+        this.resetCodewhispererService()
+        this.initializeCodewhispererService('identityCenter', newProfile.region)
+
+        this.activeIdcProfile = newProfile
```
volodkevych, line 318: nit: can we set `this.state = 'INITIALIZED'` here not inside of `initializeCodewhispererService` method. It's more readable.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,403 @@
...
+
+        throw new AmazonQServiceNotInitializedError()
+    }
+
+    private resetCodewhispererService() {
+        // TODO: terminate inflight requests for old service, if exist
+
+        this.cachedCodewhispererService = undefined
+        this.activeIdcProfile = undefined
+        this.state = 'PENDING_CONNECTION'
```
volodkevych, line 351: I do not think we should reset the state inside of this method. I think we should change to 'PENDING_CONNECTION' only on sign out.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
volodkevych, line 1: This class requires extensive unit tests.

volodkevych: Also comments on top with what this class is doing would be useful.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
viktorsaws: this is our currently standard way for exposing APIs from SDK clients in this Service class, so made it in line with other methods.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
viktorsaws: Will do as follow-up

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,403 @@
...
+            this.state = 'PENDING_Q_PROFILE_UPDATE'
+        }
+
+        const profiles = await listAvailableProfiles(this.features)
+
+        const newProfile = profiles.find(el => el.arn === newProfileArn)
+
+        if (!newProfile) {
+            // TODO: do more validation if necessary, for now, checking only existence of the profile
+            // TODO: do we need to reset service here if requested profile does not exist anymore?
```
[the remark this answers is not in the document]
viktorsaws: This is actually implemented, this TODO comment was outdated

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: Yes. I am adding this check here to proactively test connection status before we receive `creds/remove` call from client. This is to support cases when client does not implement Creds remove or does not send this event correctly

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
viktorsaws: Will follow-up

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,464 @@
...
+                const customizationArn = textUtils.undefinedIfEmpty(qConfig.customization)
+                this.configurationCache.set('customizationArn', customizationArn)
+                this.log(`Amazon Q Service Configuration updated to use ${customizationArn}`)
+
+                // aws.q.optOutTelemetry - telemetry optout option
+                const optOutTelemetryPreference = qConfig['optOutTelemetry'] === true ? 'OPTOUT' : 'OPTIN'
+                this.configurationCache.set('optOutTelemetryPreference', optOutTelemetryPreference)
+
+                if (this.cachedCodewhispererService) {
+                    this.cachedCodewhispererService.customizationArn = customizationArn
```
pras0131, line 229: I think we should also update `optOutTelemetryPreference` flag here in case of `cachedCodewhispererService` being non-null

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,464 @@
...
+                )
+
+                this.configurationCache.set(
+                    'includeSuggestionsWithCodeReferences',
+                    includeSuggestionsWithCodeReferences
+                )
+                this.configurationCache.set('shareCodeWhispererContentWithAWS', shareCodeWhispererContentWithAWS)
+
+                if (this.cachedCodewhispererService) {
+                    this.cachedCodewhispererService.shareCodeWhispererContentWithAWS = shareCodeWhispererContentWithAWS
```
pras0131, line 256: Should we also set `includeSuggestionsWithCodeReferences`?

client/vscode/src/activation.ts
```diff
@@ -213,5 +215,10 @@ export async function activateDocumentsLanguageServer(extensionContext: Extensio
         await registerNotification(client)
     }
 
+    const enableAmazonQProfiles = process.env.ENABLE_AMAZON_Q_PROFILES === 'true'
```
awschristou, line 218: Line 159 (creating the `initializationOptions` payload) needs to conditionally set `developerProfiles` based on `process.env.ENABLE_AMAZON_Q_PROFILES`

.vscode/launch.json
```diff
@@ -88,18 +88,20 @@
...
                 "ENABLE_CHAT": "true",
+                "ENABLE_CUSTOMIZATIONS": "true",
+                "ENABLE_AMAZON_Q_PROFILES": "true",
                 "ENABLE_AWS_Q_SECTION": "true"
                 // "HTTPS_PROXY": "http://127.0.0.1:8888",
                 // "AWS_CA_BUNDLE": "/path/to/cert.pem"
-            },
-            "preLaunchTask": "compile"
+            }
+            // "preLaunchTask": "compile"
```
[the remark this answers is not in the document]
viktorsaws: It's useless compilation which takes extra time to run test when nothing changes.

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
```diff
@@ -6,6 +6,7 @@ import { TelemetryService } from './telemetryService'
 import { getUserAgent, makeUserContextObject } from './utilities/telemetryUtils'
 import { DEFAULT_AWS_Q_REGION, DEFAULT_AWS_Q_ENDPOINT_URL } from '../constants'
 import { SDKInitializator } from '@aws/language-server-runtimes/server-interface'
+import { AmazonQTokenServiceManager } from './amazonQServiceManager/AmazonQTokenServiceManager'
```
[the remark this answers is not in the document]
viktorsaws: Yes, it will be used in all servers.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: no, includeSuggestionsWithCodeReferences is not used inside of `CodewhispererService`, only inside of `CodeWhispererServer` business logic. I added it here for completeness.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: Good callout, let me review how settings are set.

client/vscode/src/activation.ts
viktorsaws: good callout, thanks

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: I updated how configuration is handled in AmazonQTokenServiceManager in this commit https://github.com/aws/language-servers/pull/814/commits/791d21358240a34815ff05b180596adb21d04aaf. 

We actually can't hook in to `lsp` feature directly, as doing so overrides original handler set by consuming Server. I will think about how to optimize it, since it is not ideal setup, but I think could be done as follow-up

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: I will think about how to remove code duplication when we handle LSP Configuration in all different Servers as a followup

server/aws-lsp-codewhisperer/package.json
viktorsaws: addressed

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
pras0131: I think logs are added but `optOutTelemetryPreference`is not attached with the cachedCodewhispererService yet.

pras0131 (approved):

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,484 @@
...
+        }
+    }
+
+    private async handleProfileChange(newProfileArn: string): Promise<void> {
+        if (!this.enableDeveloperProfileSupport) {
+            this.log('Developer Profiles Support is not enabled')
+            return
+        }
+
+        if (!newProfileArn || newProfileArn.length === 0) {
```
awschristou, line 279: Toolkit will send a null arn to indicate "no profile". Signing out is one example of this. Is it possible for this to be handled as a non-error?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,484 @@
...
+                    this.cachedCodewhispererService.shareCodeWhispererContentWithAWS = shareCodeWhispererContentWithAWS
+                }
+            }
+        } catch (error) {
+            this.log(`Error in GetConfiguration: ${error}`)
+        }
+    }
+
+    private async handleProfileChange(newProfileArn: string): Promise<void> {
+        if (!this.enableDeveloperProfileSupport) {
```
awschristou, line 274: I have my client sending the enabled flag as part of the initialization handshake, but this remains false when I try to send a profile change. Can you please check that this service manager component is appropriately updating its "enable support" state?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,484 @@
...
+    private resetCodewhispererService() {
+        // TODO: terminate inflight requests for old service, if exist
+
+        this.cachedCodewhispererService = undefined
+        this.activeIdcProfile = undefined
+    }
+
+    private createCodewhispererServiceInstance(
+        connectionType: 'builderId' | 'identityCenter',
+        region: string = DEFAULT_AWS_Q_REGION
```
patrickvbeurden, line 399: I'm wondering if this should be `region?: string` without a default value. It seems that otherwise if nothing is explicitly passed here, we will always override this.features.runtime.getConfiguration('AWS_Q_REGION') with the default region on line 406 (because region is always defined).

viktorsaws:
Attack `profileArn` field when available to API calls done in CodeWhispererService, similar to `customizationArn`.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: On the signout (or when no credentials are detected, meaning client is not connected), language server will reset it's connection state to unconnected completely and reset profiles. In this case, sending empty profile is redundant, and not necessary.

If there is a case, when client needs to move from "Connection with IDC and Selected Profile" to "Connected with IDC - no profile" without Signout in between, then we can extend this handler to support it.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: Yeah, I haven't wired this codepath to that InitializationOption yet, going to do it today.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: good call, updating

pras0131 (approved):

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: Added wiring in latest version of PR

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: I am going to merge this PR with returning Error in this case. If different logic will be needed, I'll follow-up in separate PR

PR #817, opened 2025-03-10
feat(chat-client): handle 'openTab' requests

volodkevych:
## Problem

There is no way to request chat UI to open new or existing tab.

## Solution

The corresponding `openTab` logic was added and tested.
`@aws/language-server-runtimes` was updated to the latest release to contain new protocol.

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -278,8 +278,8 @@ export const createMynahUi = (
         return tabId ? mynahUi.getAllTabs()[tabId]?.store : undefined
     }
 
-    const createTabId = () => {
-        const tabId = mynahUi.updateStore('', tabFactory.createTab(false, disclaimerCardActive))
+    const createTabId = (needWelcomeMessages: boolean = false) => {
+        const tabId = mynahUi.updateStore('', tabFactory.createTab(needWelcomeMessages, disclaimerCardActive))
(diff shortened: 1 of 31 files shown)


chat-client/src/client/mynahUi.ts
```diff
@@ -399,11 +400,25 @@ ${params.message}`,
...
 
+    const openTab = (tabId?: string) => {
+        if (tabId && tabId !== mynahUi.getSelectedTabId()) {
+            mynahUi.selectTab(tabId)
+        }
+        if (!tabId) {
+            tabId = createTabId(true)
+        }
+
+        if (tabId) {
```
ege0zcan, line 411: Is there any chance that this if check returns false? Based on my understanding it is guaranteed to return true

ege0zcan (approved):

viktorsaws (approved):

chat-client/src/client/mynahUi.ts
volodkevych: If the tabs limit was reached.

ege0zcan: Makes sense, I forgot about the tab limit. What happens from the point of view of the server that requested the tab open in this case? If I am understanding the code correctly, we are not returning an error or a response to the server in this case and a server might be left waiting for a tab that will never open, is this the case or did I miss something?

PR #820, opened 2025-03-11
chore: adding http keepAlive agent in AWS SDK client

pras0131:
## Problem
Adding http keepalive agent in AWS SDK Client to help in reducing the latency of generateCompletions API.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
@@ -17,6 +17,7 @@ import {
     CodeWhispererTokenClientConfigurationOptions,
     createCodeWhispererTokenClient,
 } from '../client/token/codewhisperer'
+import * as https from 'node:https'
 
 // Define our own Suggestion interface to wrap the differences between Token and IAM Client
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -154,6 +155,13 @@ export class CodeWhispererServiceToken extends CodeWhispererServiceBase {
                     })
                 },
             ],
+            httpOptions: {
```
ege0zcan, line 158: I wonder if we should set this in language servers individually or if it is better to let the runtime handle it: https://github.com/aws/language-server-runtimes/blob/6131727c5d7cd7f7fd1883e1749975d87c3e0504/runtimes/runtimes/util/standalone/proxyUtil.ts#L37

This approach might also override the proxy config done by the runtimes and cause issues with proxies

francescoopiccoli: yes we should set this in runtimes. We want capabilities to be environment agnostic, here we are importing from node:https, implementing it here would probably break webworker/ ucaf runtime bundling

PR #822, opened 2025-03-12
Add support for listing available profiles

patrickvbeurden:
## Problem

We need a way to retrieve Q developer profiles from the `lsp` server for users to be able to select a specific profile during `identityCenter` authentication flow. 

## Solution

1. Defined a new section `aws.q.developerProfiles` for the existing `getConfigurationFromServer` request method to request available profiles.
2. Made requests to `aws.q` more granular by also adding `aws.q.customizations`. While clients cannot signal that they support profiles, `aws.q` will only return customizations.
3. Created a reusable method `getListAllAvailableProfilesHandler` to create a handler which iterates over the available `AWS_Q_ENDPOINTS`, aggregating the returned profiles from each endpoint.

testing:
1. Added two unit tests (1) to validate the aggregation and (2) to validate correct handling of connection type.
2. Added two unit tests to validate the pagination mechanism
3. Added unit tests for the `onGetConfigurationFromServer` handler to validate the correct resources are retrieved.
4. Added a test command to the bundled vs code client for testing the E2E flow **NOTE** prod endpoint currently still returns a throttling exception.

## License


server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/qDeveloperProfiles.ts
```diff
@@ -0,0 +1,72 @@
...
+        const qEndpoints = endpoints ?? AWS_Q_ENDPOINTS
+
+        try {
+            for (const [region, endpoint] of Object.entries(qEndpoints)) {
+                try {
+                    logging.debug(`Fetching profiles from region: ${region} and endpoint: ${endpoint}`)
+
+                    const codeWhispererService = service(region, endpoint)
+
+                    const response = await codeWhispererService.listAvailableProfiles({
```
volodkevych, line 46: Should pagination be supported for this endpoint, thinking about long-term?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/qDeveloperProfiles.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export type ListAllAvailableProfilesHandler = (
+    params: ListAllAvailableProfilesHandlerParams
+) => Promise<AmazonQDeveloperProfile[]>
+
+const MAX_Q_DEVELOPER_PROFILES = 10
+
+export const getListAllAvailableProfilesHandler =
+    (service: (region: string, endpoint: string) => CodeWhispererServiceToken): ListAllAvailableProfilesHandler =>
+    async ({ connectionType, logging, endpoints }) => {
```
volodkevych, line 30: Where do we have `FRA` endpoint configured?

patrickvbeurden: The idea would be to add it to the `AWS_Q_ENDPOINTS` constant with fast-follow, I think we don't have a public alias available yet so that's why it is left out at the moment.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/qDeveloperProfiles.ts
patrickvbeurden: Added pagination

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/qDeveloperProfiles.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (!connectionType || connectionType !== 'identityCenter') {
+            logging.debug('Connection type is not set or not identityCenter - returning empty response.')
+            return []
+        }
+
+        const allProfiles: AmazonQDeveloperProfile[] = []
+        const qEndpoints = endpoints ?? AWS_Q_ENDPOINTS
+
+        try {
+            for (const [region, endpoint] of Object.entries(qEndpoints)) {
```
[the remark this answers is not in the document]
patrickvbeurden: Added parallel calls as suggested.

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -53,6 +58,19 @@ export const QConfigurationServerToken =
 
                     return { customizations }
                 }
+
+                if (params.section === Q_DEVELOPER_PROFILES_CONFIGURATION_SECTION) {
```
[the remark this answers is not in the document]
patrickvbeurden: Adjusted the handler to support granular requests i.e. `aws.q`, `aws.q.customizations` and `aws.q.developerProfiles` and added unit tests to verify the different requests trigger the right resource fetching flows.

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -40,18 +53,71 @@ export const QConfigurationServerToken =
...
+                            ])
+                            return Q_DEVELOPER_PROFILES_ENABLED
+                                ? { customizations, developerProfiles }
+                                : { customizations }
+                        case Q_CUSTOMIZATIONS_CONFIGURATION_SECTION:
+                            customizations = await onGetConfigurationFromServerManager.listAvailableCustomizations()
+
+                            return customizations
+                        case Q_DEVELOPER_PROFILES_CONFIGURATION_SECTION:
+                            developerProfiles = await onGetConfigurationFromServerManager.listAvailableProfiles()
```
viktorsaws, line 105: I think we need check for `Q_DEVELOPER_PROFILES_ENABLED` flag here as well

server/aws-lsp-codewhisperer/src/language-server/configuration/onConfigurationFromServerManager.ts
```diff
@@ -0,0 +1,60 @@
...
+    getListAllAvailableProfilesHandler,
+    ListAllAvailableProfilesHandler,
+} from '../amazonQServiceManager/qDeveloperProfiles'
+import { CodeWhispererServiceToken } from '../codeWhispererService'
+import { Customizations } from '../../client/token/codewhispererbearertokenclient'
+import { Q_CUSTOMIZATIONS, Q_DEVELOPER_PROFILES } from './constants'
+
+const ON_GET_CONFIGURATION_FROM_SERVER_ERROR_PREFIX = 'Failed to fetch: '
+
+export class OnGetConfigurationFromServerManager {
```
viktorsaws, line 18: nit: I'd rename this class to something else from Manager, since it does not really manages anything, it's just a requests service.

Also it looks very small, perhaps we could move this code to QConfigurationServer, but I'll leave it up to you.

patrickvbeurden: Fair point, moved the class to QConfigurationServer and renamed to `ServerConfigurationProvider`

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
patrickvbeurden: Done, also moved the flag into `ServerConfigurationProvider`. Later we can also update and track the flag here when we expose client capabilities for profile support.

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -40,22 +55,131 @@ export const QConfigurationServerToken =
...
+                try {
+                    switch (section) {
+                        case Q_CONFIGURATION_SECTION:
+                            ;[customizations, developerProfiles] = await Promise.all([
+                                serverConfigurationProvider.listAvailableCustomizations(),
+                                serverConfigurationProvider.listAvailableProfiles(),
+                            ])
+
+                            return serverConfigurationProvider.qDeveloperProfilesEnabled
+                                ? { customizations, developerProfiles }
```
awschristou, line 99: How does a client caller know that the first object is a customizations array and the second object is developer profiles? Is there a differentiator somewhere that I overlooked?

patrickvbeurden: They will receive a dictionary like:

```
{
  'customizations': [...],
  'developerProfiles': [...]
}
```

So the dictionary keys should be the differentiator, and for the client caller it is important to know which key maps to which object instead of the order.

viktorsaws (approved):

PR #823, opened 2025-03-12
feat(identity): device code support

hayemaxi:
Follow up PRs will enable it for use in the identity server.

Flow:
- Shows a user code ack to client
  - can be cancelled
- Then opens the url and a progress bar on the client with the user code as a message on it
- Periodically polls for the user to complete login or the client to cancel the auth flow

Misc:
- modify eslint rule to support a common idiom (`while (true) {}`)
- move util function to identity server because it was unused where it was. Needs to be migrated to utils file, but not sure the of the best way to do this right now.

## License


server/aws-lsp-identity/src/sso/deviceCode/deviceCodeFlow.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    params.observability.logging.log('Storing and returning created SSO token.')
+    return UpdateSsoTokenFromCreateToken(result, params.clientRegistration, params.ssoSession)
+}
+
+function pollAuthorization(
+    fn: () => Promise<CreateTokenCommandOutput>,
+    expiresAt: Date,
+    token: CancellationToken,
+    intervalMs: number = backoffDelayMs
```
floralph, line 87: nit: For consistency with other identity code, consider using "Millis" instead of "Ms".

server/aws-lsp-identity/src/sso/utils.ts
```diff
@@ -100,3 +107,35 @@ export function UpdateSsoTokenFromCreateToken(
...
+export type ShowMessageRequest = (params: ShowMessageRequestParams) => Promise<MessageActionItem | null>
+export type ShowProgress = Lsp['sendProgress']
+
+export type SsoFlowParams = {
+    clientName: string
+    clientRegistration: SsoClientRegistration
+    ssoSession: SsoSession
+    handlers: {
+        showUrl: ShowUrl
+        showMsgRequest: ShowMessageRequest
```
floralph, line 135: nit: For consistency, consider naming this showMessageRequest.

server/aws-lsp-identity/src/sso/deviceCode/deviceCodeFlow.ts
```diff
@@ -0,0 +1,160 @@
...
+    ).catch(err => {
+        if (err instanceof AsyncTimeoutError) {
+            throw new AwsError(`User code was not acknowledged.`, AwsErrorCodes.E_TIMEOUT)
+        }
+
+        // Client threw an error
+        throw err
+    })
+
+    // Any response is acceptable
```
floralph, line 147: question: Because there is only one action at this time?

server/aws-lsp-identity/src/sso/utils.ts
```diff
@@ -100,3 +107,35 @@ export function UpdateSsoTokenFromCreateToken(
...
+        .finally(() => {
+            clearTimeout(timeoutHandle)
+        })
+}
+
+export type ShowUrl = (url: URL) => void
+export type ShowMessageRequest = (params: ShowMessageRequestParams) => Promise<MessageActionItem | null>
+export type ShowProgress = Lsp['sendProgress']
+
+export type SsoFlowParams = {
```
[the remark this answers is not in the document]
hayemaxi: I agree that is the next step there but it is not a P0.

server/aws-lsp-identity/src/sso/deviceCode/deviceCodeFlow.ts
hayemaxi: Yes. The expectation is that the client returns either null (canceled/user did not pick) or one of the provided actions. If it is not null then we just assume we got the only action that we provided back, no need to check.

server/aws-lsp-identity/src/sso/deviceCode/deviceCode.test.ts
```diff
@@ -0,0 +1,193 @@
+import { expect, use } from 'chai'
```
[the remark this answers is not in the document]
hayemaxi: Yes, oversight

floralph (approved):

app/aws-lsp-identity-runtimes/package.json
```diff
@@ -7,7 +7,7 @@
         "compile": "tsc --build"
     },
     "dependencies": {
-        "@aws/language-server-runtimes": "^0.2.40",
+        "@aws/language-server-runtimes": "^0.2.41",
```
pras0131, line 10: I suggest it's better to bump version once it's available in npmpm.

pras0131: https://npmpm.corp.amazon.com/pkg/@aws/language-server-runtimes

server/aws-lsp-identity/src/sso/deviceCode/deviceCodeFlow.ts
```diff
@@ -0,0 +1,160 @@
...
+            }
+            if (token.isCancellationRequested) {
+                throw new AwsError(`Client cancelled the auth flow.`, AwsErrorCodes.E_CANCELLED)
+            }
+
+            try {
+                return await fn()
+            } catch (err) {
+                if (err instanceof SlowDownException) {
+                    intervalMillis += backoffDelayMillis
```
pras0131, line 105: Do we want to keep polling forever or it's better to define X number of attempts?

hayemaxi: The way out of this is the `expiresAt` check a few lines above. The number of attempts is defined by the interval between checks. Both of these things are set by the OIDC client.

We can't add our own sanity check exit at X number of attempts. E.g. the expires is 10 minutes and the interval is 5 seconds. That is 120 attempts. If the OIDC client passes other values, the number of attempts changes. We can't just say exit after 10 or 200 attempts, both could be reasonable.

app/aws-lsp-identity-runtimes/package.json
hayemaxi: Why? Is merging this going to block some sort of internal build?

pras0131: Local build might fail as it will try to fetch from npmpm where the package is not available yet.

pras0131 (approved):

PR #839, opened 2025-03-17
feat(amazonq): handle client signalling support for q developer profiles

patrickvbeurden:
## Problem

Currently, we have functionality for fetching and returning Q developer profiles, but by default we have it disabled. Instead, iff the client signals support for Q developer profiles, we should return them in the applicable `GetConfigurationFromServer` requests (`aws.q`, `aws.q.developerProfiles`)

## Solution

At initialization, we check the `AWSInitializationOptions` in `InitializationParams` for the structure:

```
q: {
  developerProfiles: boolean;
}
```

If these params are present, and the `developerProfiles` flag is explicitly set to true, we update the corresponding flag in `ServerConfigProvider`, which then in turn enables fetching and returning the Q developer profiles.

Testing: added unit tests that verify the flag-checking mechanism, as well as the initialize handler correctly setting the `qDeveloperProfilesEnabled` flag. Also tested E2E flow with bundled vs client by signalling `developerProfiles: true`
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
@@ -1,5 +1,4 @@
 import {
-    AWSInitializationOptions,
     CancellationToken,
     CredentialsProvider,
     GetConfigurationFromServerParams,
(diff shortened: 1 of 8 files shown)


server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -183,3 +195,21 @@ export class ServerConfigurationProvider {
         return new ResponseError(LSPErrorCodes.RequestFailed, message)
     }
 }
+
+const AWSQCapabilitiesKey = 'q'
+const developerProfilesEnabledKey = 'developerProfiles'
+
+export function signalsAWSQDeveloperProfilesEnabled(initializationOptions: AWSInitializationOptions): boolean {
```
viktorsaws, line 202: I think we could move this function to separate helper, as it will be useful in another places.

viktorsaws (approved):

viktorsaws (approved):

PR #840, opened 2025-03-17
feat(identity): enable device code

hayemaxi:
- Add device code as selectable auth flow
- Refactor PKCE flow to use params structure
- Add IdentityServer to cw lsp activation and update version
- Add example message handlers in the sample vscode app

## License

### DIFF
--- a/client/vscode/src/identityActivation.ts
+++ b/client/vscode/src/identityActivation.ts
@@ -1,7 +1,11 @@
-import { commands, window } from 'vscode'
+import { compactDecrypt } from 'jose'
+import { ProgressLocation, commands, window } from 'vscode'
 import {
     AwsResponseError,
     GetSsoTokenParams,
+    GetSsoTokenProgress,
+    GetSsoTokenProgressToken,
+    GetSsoTokenProgressType,
     getSsoTokenRequestType,
     GetSsoTokenResult,
     IamIdentityCenterSsoTokenSource,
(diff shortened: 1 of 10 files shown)


client/vscode/src/identityActivation.ts
```diff
@@ -88,6 +137,10 @@ async function execTestCommand(client: LanguageClient): Promise<void> {
                     kind: SsoTokenSourceKind.IamIdentityCenter,
                     profileName: 'eclipse-q-profile',
                 } satisfies IamIdentityCenterSsoTokenSource,
+                options: {
+                    // authorizationFlow: 'Pkce',
```
jpinkney-aws, line 141: Would it make sense to register a sample test command for each of the flows? Or will this code never changed again unless people are specifically need pkce support?

client/vscode/src/identityActivation.ts
```diff
@@ -17,12 +21,57 @@ import {
...
+            const actions = params.actions?.map(a => a.title) ?? []
+            const response = await window.showInformationMessage(params.message, { modal: true }, ...actions)
+            return params.actions?.find(a => a.title === response) ?? null
+        }
+    )
+
+    let promise: Promise<void> | undefined
+    let resolver: () => void = () => {}
+    client.onProgress(GetSsoTokenProgressType, GetSsoTokenProgressToken, async (partialResult: GetSsoTokenProgress) => {
+        const decryptedKey = await compactDecrypt(partialResult as unknown as string, encryptionKey)
```
floralph, line 49: question:  I think I missed something earlier.  Why is GetSsoTokenProgress encrypted?

server/aws-lsp-identity/src/language-server/identityService.ts
```diff
@@ -64,14 +79,18 @@ export class IdentityService {
...
-                    this.showUrl,
+                    handlers: this.handlers,
                     token,
-                    this.observability
-                ).catch(reason => {
+                    observability: this.observability,
+                }
+
+                const flow =
+                    this.authFlows[params.options?.authorizationFlow ?? getSsoTokenOptionsDefaults.authorizationFlow]
```
floralph, line 92: Could this be more resilient as `const flow = this.authFlows[params.options?.authorizationFlow] ?? this.authFlows[getSsoTokenOptionsDefaults.authorizationFlow]` in the case params.options.authorizationFlow is truthy, but not valid?  It's also reasonable for a 'null is not a function' error below.

floralph (approved):

client/vscode/src/identityActivation.ts
floralph: The sample vscode client is used for development support, so if someone is working on and expecting PKCE, they would change the code.  The client does provide a quick smoke test, but the PKCE errors should pop in unit tests (or we have a test coverage miss).

natdurlik (approved):

client/vscode/src/identityActivation.ts
hayemaxi: You didn't miss anything, we aren't encrypting them within the Identity Service. Unlike the `window/*` protocols, the LSP by default is encrypting these requests before sending them to the server.

server/aws-lsp-identity/src/language-server/identityService.ts
hayemaxi: I have updated it to achieve what you suggested.

client/vscode/src/identityActivation.ts
hayemaxi: Yeah, I didn't put much effort into this file because I see it for local testing only, not a read-only sample or demo.
As it stands right now it needs some local configuration to work that I did not include in this PR.

floralph (approved):

jpinkney-aws (approved):

PR #842, opened 2025-03-18
chore: added auth listener to reset the service manager state in case of bearer token signout

pras0131:
Listener to update the local state of AmazonQTokenServiceManager class, so bearer token signout can reset the state to avoid any kind of race conditions.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
@@ -10,6 +10,7 @@ import {
     LSPErrorCodes,
     SsoConnectionType,
     CancellationToken,
+    ExecuteCommandParams,
 } from '@aws/language-server-runtimes/server-interface'
 import { DEFAULT_AWS_Q_ENDPOINT_URL, DEFAULT_AWS_Q_REGION, AWS_Q_ENDPOINTS } from '../../constants'
 import { CodeWhispererServiceToken } from '../codeWhispererService'
@@ -123,7 +124,19 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
     }
 
     private setupAuthListener(): void {
-        // TODO: listen on changes to credentials and signout events from client to manage state correctly.
+        this.features.lsp.onExecuteCommand(
+            async (params: ExecuteCommandParams, _token: CancellationToken): Promise<any> => {
+                this.log(`Received command: ${params.command}`)
+                switch (params.command) {
+                    case 'bearerCredentialsDeleteCommand':
+                        this.cachedCodewhispererService = undefined
+                        this.activeIdcProfile = undefined
+                        this.connectionType = 'none'
+                        break
+                }
+                return
+            }
+        )
     }
 
     private setupConfigurationListeners(): void {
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -123,7 +123,12 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
     }
 
     private setupAuthListener(): void {
-        // TODO: listen on changes to credentials and signout events from client to manage state correctly.
+        this.features.credentialsProvider.onBearerCredentialsDelete(() => {
```
patrickvbeurden, line 126: Should we also set the state back to `PENDING_CONNECTION`?

pras0131: Right, I forgot to reset the variable. Let me do that.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -138,7 +139,16 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
     }
 
     private setupAuthListener(): void {
-        // TODO: listen on changes to credentials and signout events from client to manage state correctly.
+        this.features.credentialsProvider.onCredentialsDeleted((type: CredentialsType) => {
+            this.log(`Received credentials delete event for type: ${type}`)
+            if (type === 'iam') {
```
patrickvbeurden, line 144: Can we be in a state where the client actually had IAM credentials and tries to delete them? Wondering if we should error if we receive this event?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -138,7 +139,16 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
     }
 
     private setupAuthListener(): void {
-        // TODO: listen on changes to credentials and signout events from client to manage state correctly.
+        this.features.credentialsProvider.onCredentialsDeleted((type: CredentialsType) => {
```
patrickvbeurden, line 142: Could you add a unit test to service manager for this function please?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
pras0131: Right now we are only using the service manager utility for bearer token but the event is emitted in both the scenarios. So to keep things simple, just not doing anything here in case of IAM.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -138,7 +139,16 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
...
+            this.log(`Received credentials delete event for type: ${type}`)
+            if (type === 'iam') {
+                return
+            }
+            this.cachedCodewhispererService = undefined
```
viktorsaws, line 147: As discussed, we'd have to call `this.resetCodewhispererService()`, as we reset inflight calls there with `this.cachedCodewhispererService?.abortInflightRequests()`

patrickvbeurden (approved):

PR #846, opened 2025-03-19
fix: disable slack notifications for PRs from forks

natdurlik:
## Problem
The Slack notification workflow fails on PRs from forked repositories. These failures are false positives and create unnecessary noise in our CI pipeline.
Reason is that GitHub Secrets are not passed to the runner when a workflow is triggered from a forked repository, because of security reasons.

## Solution
Disabling slack notification workflow for PRs coming from forks. We could think of some better solution that will work with forks in the future.

## License

### DIFF
--- a/.github/workflows/new_pr.yml
+++ b/.github/workflows/new_pr.yml
@@ -6,7 +6,7 @@ on:
 jobs:
     notify:
         name: Slack notification
-        if: github.event.pull_request.user.login != 'dependabot[bot]' && github.event.pull_request.draft == false
+        if: github.event.pull_request.user.login != 'dependabot[bot]' && github.event.pull_request.draft == false && github.event.pull_request.head.repo.fork == false
         runs-on: [ubuntu-latest]
         steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/new_pr.yml
```diff
@@ -6,7 +6,7 @@ on:
 jobs:
     notify:
         name: Slack notification
-        if: github.event.pull_request.user.login != 'dependabot[bot]' && github.event.pull_request.draft == false
+        if: github.event.pull_request.user.login != 'dependabot[bot]' && github.event.pull_request.draft == false && github.event.pull_request.head.repo.fork == false
```
pras0131, line 9: how does it prevent slack notifications from failing for branches as earlier we saw just a branch fork also fails.

natdurlik: cloud you send me an example? when I checked the Actions page, I only saw failed notifications from forked repositories.

ege0zcan (approved):

PR #848, opened 2025-03-19
chore: abort all inflight requests when resetCodewhispererService is invoked

pras0131:
Add the capability to track inflight requests for Code Whisperer client and abort them once resetCodeWhisperer service is invoked.

## License


server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -55,6 +55,23 @@ export abstract class CodeWhispererServiceBase {
...
+            request.abort()
+        })
+        this.inflightRequests.clear()
+    }
+
+    trackRequest(request: any) {
+        this.inflightRequests.add(request)
+    }
+
+    completeRequest(request: any) {
```
patrickvbeurden, line 71: Is it possible to add unit tests for these three new methods?

pras0131: can we do it in a follow up PR? will need to check how we can test specific additions introduced here.

patrickvbeurden: Yeah it might be tricky to cover (all of) them. Alternatively it might be more feasible to add a test command to the VS code client that tries to trigger the abort flow instead?

patrickvbeurden (approved):

server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -55,6 +56,23 @@ export abstract class CodeWhispererServiceBase {
     public profileArn?: string
     abstract client: CodeWhispererSigv4Client | CodeWhispererTokenClient
 
+    inflightRequests: Set<AWS.Request<any, AWSError> & RequestExtras> = new Set()
+
+    abortInflightRequests() {
+        this.inflightRequests.forEach(request => {
+            request.abort()
```
viktorsaws, line 63: Were you able to test what happens in LSP Handlers when the calls are aborted? Will they return specific error?

PR #849, opened 2025-03-19
feat(amazonq): update initialization flow for AmazonQTokenServiceManager

viktorsaws:
Updating initialization flow for AmazonQTokenServiceManager:

1. Now it must be initialized in consuming Server inside as part of `lsp.onInitialized` handler, to ensure that LSP connection was initialized and `InitializeParams` are cached in `lsp` feature.
2. Reading `q. developerProfiles` initialization option from cached `InitializeParams` to toggle Developer Profiles support in manager instead of passing a flag. 
3. Computing `customUserAgent` from cached `InitializeParams` instead of using setter in consuming code.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
@@ -10,11 +10,13 @@ import {
     LSPErrorCodes,
     SsoConnectionType,
     CancellationToken,
+    InitializeParams,
 } from '@aws/language-server-runtimes/server-interface'
 import { DEFAULT_AWS_Q_ENDPOINT_URL, DEFAULT_AWS_Q_REGION, AWS_Q_ENDPOINTS } from '../../constants'
 import { CodeWhispererServiceToken } from '../codeWhispererService'
 import {
     AmazonQError,
+    AmazonQServiceInitializationError,
     AmazonQServiceInvalidProfileError,
     AmazonQServiceNoProfileSupportError,
     AmazonQServiceNotInitializedError,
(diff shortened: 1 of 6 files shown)


server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -85,27 +91,42 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
...
+        }
+
         // Bind methods that are passed by reference to some handlers to maintain proper scope.
         this.serviceFactory = this.serviceFactory.bind(this)
         this.handleDidChangeConfiguration = this.handleDidChangeConfiguration.bind(this)
 
-        this.enableDeveloperProfileSupport = enableDeveloperProfileSupport
+        this.log('Reading enableDeveloperProfileSupport setting from AWSInitializationOptions')
+        if (this.features.lsp.getClientInitializeParams()?.initializationOptions?.aws) {
+            const awsOptions = this.features.lsp.getClientInitializeParams()?.initializationOptions?.aws || {}
```
patrickvbeurden, line 125: nit: I think if you create `awsOptions` before the if-check, then `|| {}` "check" is not necessary anymore

server/aws-lsp-codewhisperer/src/language-server/codeWhispererService.ts
```diff
@@ -52,6 +52,7 @@ export abstract class CodeWhispererServiceBase {
     protected readonly codeWhispererEndpoint
     public shareCodeWhispererContentWithAWS = false
     public customizationArn?: string
+    public profileArn?: string
```
pras0131, line 55: Rather than directly updating the instance variable from Service Manager, I suggest we have a setter instead as otherwise looking at this - it becomes difficult to tell how and when is being assigned.

viktorsaws: Can we do this as followup? We would need to refactor all `shareCodeWhispererContentWithAWS`, `customizationArn` and `profileArn` to use same pattern.

pras0131 (approved):

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -85,27 +91,42 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
...
+            )
         }
 
         this.features = features
         this.logging = features.logging
 
+        if (!this.features.lsp.getClientInitializeParams()) {
+            this.log('AmazonQTokenServiceManager initialized before LSP connection was not initialized.')
+            throw new AmazonQServiceInitializationError(
+                'AmazonQTokenServiceManager initialized before LSP connection was not initialized.'
```
patrickvbeurden, line 115: nit: I think the message should be `... before LSP connection was initialized.`

patrickvbeurden (approved):

PR #853, opened 2025-03-20
feat(amazonq): intergrate regionalization support into Q Chat server

viktorsaws:
## Problem
We need to add regionalization support to Q Chat Server.

## Solution
* Extend AmazonQTokenServiceManager to expose factory method to create instance of `CodeWhispererStreaming` client with currently active region and endpoint
* Integrate Q Chat server to use new method
* Remove obsolete configuration passed in various Q Chat server service classes.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
@@ -38,6 +38,7 @@ import {
     signalsAWSQDeveloperProfilesEnabled,
 } from './qDeveloperProfiles'
 import { getUserAgent } from '../utilities/telemetryUtils'
+import { getBearerTokenFromProvider } from '../utils'
 
 export interface Features {
     lsp: Lsp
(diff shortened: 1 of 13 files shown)


server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -488,6 +509,24 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
         return service
     }
 
+    private streamingClientFactory(region: string, endpoint: string): CodeWhispererStreaming {
+        const credentials = this.features.credentialsProvider.getCredentials('bearer') as BearerCredentials
```
patrickvbeurden, line 513: nit: not sure if this works, but maybe `as BearerCredentials | undefined`? Since we need to do the check afterwards anyways

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -488,6 +509,24 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
...
+
+        if (!credentials || !credentials.token) {
+            throw new Error(MISSING_BEARER_TOKEN_ERROR)
+        }
+
+        const streamingClient = this.features.sdkInitializator(CodeWhispererStreaming, {
+            region,
+            endpoint,
+            token: { token: credentials.token },
```
patrickvbeurden, line 522: Not completely sure how it works, but is the token set once at initialization? Or do we need to take into account this method might be called for token refresh for longer sessions?

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -104,7 +104,13 @@ export class ChatController implements ChatHandlers {
         const conversationIdentifier = session?.conversationId ?? 'New conversation'
         try {
             this.#log('Request for conversation id:', conversationIdentifier)
-            requestInput = this.#triggerContext.getChatParamsFromTrigger(params, triggerContext, this.#customizationArn)
+            const profileArn = AmazonQTokenServiceManager.getInstance(this.#features).getActiveProfileArn()
```
patrickvbeurden, line 107: nit: should we add the service manager as a dependency to this class?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
patrickvbeurden: nit: we could also consider to keep using the `getBearerTokenFromProvider` since it seems to have very similar logic

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: this is very good catch - token is pushed from client to server and can expire of change mid session. Caching client like this opens a problem of keeping stale client when we don't want it. I will review what to do here. I think easier could be to maintain initial approach and not cache streaming client for now.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
viktorsaws: this should not be necessary, since any code could initiate the singleton.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: I reverted the approach and don't cache streaming client for now.

PR #855, opened 2025-03-21
chore: language server runtimes version bump

vidkhath:
## Problem
Protocol is upgraded in Language Server Runtimes

## Solution
Bump version 

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -299,8 +299,11 @@ export const createMynahUi = (
     }
 
     const addChatResponse = (chatResult: ChatResult, tabId: string, isPartialResult: boolean) => {
+        const { type, ...chatResultWithoutType } = chatResult
+
         if (isPartialResult) {
-            mynahUi.updateLastChatAnswer(tabId, { ...chatResult })
+            // type for MynahUI differs from ChatResult types so we ignore it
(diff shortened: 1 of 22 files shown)


jpinkney-aws (approved):

jpinkney-aws (commented):
Looks like theres a couple conflicts

chat-client/src/client/mynahUi.ts
```diff
@@ -299,8 +299,11 @@ export const createMynahUi = (
     }
 
     const addChatResponse = (chatResult: ChatResult, tabId: string, isPartialResult: boolean) => {
+        const { type, ...chatResultWithoutType } = chatResult
+
         if (isPartialResult) {
-            mynahUi.updateLastChatAnswer(tabId, { ...chatResult })
+            // type for MynahUI differs from ChatResult types so we ignore it
+            mynahUi.updateLastChatAnswer(tabId, { ...chatResultWithoutType })
```
volodkevych, line 306: Alternatively, we could map what we get from protocol to mynahUI `type`. Then, we would enable prompts for based chat streaming too. I do not think it's something that will be used in practice, but it's a legit feature to have in general.

volodkevych (approved):

volodkevych (approved):

PR #856, opened 2025-03-21
feat(amazonq): integrate amazon q service manager with telemetry service

patrickvbeurden:
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
@@ -40,7 +40,7 @@ import { TelemetryService } from './telemetryService'
 import { AcceptedSuggestionEntry, CodeDiffTracker } from './telemetry/codeDiffTracker'
 import { DEFAULT_AWS_Q_ENDPOINT_URL, DEFAULT_AWS_Q_REGION } from '../constants'
 import { AmazonQTokenServiceManager } from './amazonQServiceManager/AmazonQTokenServiceManager'
-import { AmazonQError } from './amazonQServiceManager/errors'
+import { AmazonQError, AmazonQServiceInitializationError } from './amazonQServiceManager/errors'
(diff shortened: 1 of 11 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -681,6 +653,34 @@ export const CodewhispererServerFactory =
                 }
             }
 
+            const clientParams = safeGet(
+                lsp.getClientInitializeParams(),
```
viktorsaws, line 657: This case should never happen, as Runtime ensures that `onInitializedHandler` always called after initialization is done. 

If we want to keep a check like this, I'd suggest we expose specific flag on `lsp` to signal initialization state, instead of relying on `getClientInitializeParams()` call. This is not block merging this PR though, we could do it in follow-up

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
```diff
@@ -3,38 +3,22 @@ import { ChatController } from './chat/chatController'
...
-            logging,
-            workspace,
-            awsQRegion,
-            awsQEndpointUrl,
-            sdkInitializator
-        )
+        let telemetryService: TelemetryService;
 
         lsp.addInitializer((params: InitializeParams) => {
-            telemetryService.updateClientConfig({
```
viktorsaws, line 34: So nice that we can get rid of duplication like this

viktorsaws (approved):

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
patrickvbeurden: Agreed and having such a flag would definitely be useful, but in this case we do need the params itself as well, so as long as it can technically be undefined, typescript will complain unfortunately (unless we use e.g. `// @ts-ignore` of course)

PR #858, opened 2025-03-21
fix(amazonq): specify code analysis scope and scan name when running scans

shruti0085:
## Problem
Q developer scans were failing to execute due to a couple of issues. This change adds fixes to address them.
## Solution
Changes included:
* Scans were timing out due to a lower timeout limit. It has now been updated to 15mins matching other IDE's timeout limits for project scans. Reference [VSCode timeout](https://github.com/aws/aws-toolkit-vscode/blob/b048f97449b6265d97a51676acc3464aeefa4c06/packages/core/src/codewhisperer/models/constants.ts#L264)
* Missing parameters specifying the scope of the scan has now been added to the `UploadToS3` call and the `CreateScanJob` call of the scan process. At this time only `PROJECT` scope is supported. Support for file auto scans is out of scope. Reference [VSCode request](https://github.com/aws/aws-toolkit-vscode/blob/b048f97449b6265d97a51676acc3464aeefa4c06/packages/core/src/codewhisperer/service/securityScanHandler.ts#L293)
* A bug related to the presence of kms key in the upload call has been fixed.  
   * The null and undefined check for kmsKey always returned a true, that has now been fixed as an AND conditional check
   * The `CreateUploadUrlResponse` 's requestHeaders is defaulted  first to pick the request Headers. Reference [VSCode logic](https://github.com/aws/aws-toolkit-vscode/blob/b048f97449b6265d97a51676acc3464aeefa4c06/packages/core/src/codewhisperer/service/securityScanHandler.ts#L390).

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanHandler.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanHandler.test.ts
@@ -8,6 +8,7 @@ import { StartCodeAnalysisRequest } from '../../client/token/codewhispererbearer
 import { CodeWhispererServiceToken } from '../codeWhispererService'
 import { SecurityScanHandler } from './securityScanHandler'
 import { RawCodeScanIssue } from './types'
+import * as ScanConstants from './constants'
 
 const mockCodeScanFindings = JSON.stringify([
     {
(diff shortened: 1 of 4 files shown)


atonaamz (approved):

jpinkney-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanHandler.test.ts
```diff
@@ -74,7 +75,10 @@ describe('securityScanHandler', () => {
 
         it('returns correct source code', async () => {
             const expectedSourceCode = 'dummy-upload-id'
-            const res = await securityScanhandler.createCodeResourcePresignedUrlHandler(Buffer.from('dummy-data'))
+            const res = await securityScanhandler.createCodeResourcePresignedUrlHandler(
+                Buffer.from('dummy-data'),
```
suprajaven, line 79: nit - Since we already have a constants file, can we move some of these there as well?

suprajaven (approved):

server/aws-lsp-codewhisperer/src/language-server/securityScan/securityScanHandler.test.ts
shruti0085: I noticed there are constants file per server. For eg. there is a separate one for chat

atonaamz (approved):

jpinkney-aws (approved):

PR #861, opened 2025-03-24
feat: Cancel inflight update requests

pras0131:
The PR introduces changes to cancel in-flight update profile requests, so only the recent request is honored whereas the old request in progress gets cancelled. This is implemented using CancellationTokenSource, calling cancel() on the source instance changes the isCancellationRequested property of the token to true, hence enabling us to return an error for the old request.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
@@ -91,7 +91,7 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
     private configurationCache = new Map()
     private activeIdcProfile?: AmazonQDeveloperProfile
     private connectionType?: SsoConnectionType
-    public profileChangeTokenSource: CancellationTokenSource | undefined
+    private profileChangeTokenSource: CancellationTokenSource | undefined
     private region?: string
     private endpoint?: string
     /**
--- a/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
@@ -186,6 +186,11 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
                     }
                 } catch (error) {
                     this.log('Error updating profiles: ' + error)
+                    if (error instanceof AmazonQServiceProfileUpdateCancelled) {
+                        throw new ResponseError(LSPErrorCodes.ServerCancelled, error.message, {
+                            awsErrorCode: error.code,
+                        })
+                    }
                     if (error instanceof AmazonQError) {
                         throw new ResponseError(LSPErrorCodes.RequestFailed, error.message, {
                             awsErrorCode: error.code,
@@ -335,6 +340,7 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
         }
 
         this.logServiceState('UpdateProfile is requested')
+        const initialState = this.state
 
         // Test if connection type changed
         this.handleSsoConnectionChange()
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -173,8 +176,13 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
...
                     if (params.section === Q_CONFIGURATION_SECTION && params.settings.profileArn) {
                         this.log(`Profile update is requested for profile ${params.settings.profileArn}`)
+                        if (this.profileChangeTokenSource) {
+                            this.profileChangeTokenSource.cancel()
+                            this.profileChangeTokenSource.dispose()
+                        }
+                        this.profileChangeTokenSource = new CancellationTokenSource()
 
-                        await this.handleProfileChange(params.settings.profileArn)
+                        await this.handleProfileChange(params.settings.profileArn, this.profileChangeTokenSource.token)
```
viktorsaws, line 185: I think we should handle `AmazonQServiceProfileUpdateCancelled` exception in `catch` block and return `ResponseError` with `ServerCancelled` error code from `LSPErrorCodes.ServerCancelled` object. This way we signal to client that server made a cancellation.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -173,8 +176,13 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
                 try {
                     if (params.section === Q_CONFIGURATION_SECTION && params.settings.profileArn) {
                         this.log(`Profile update is requested for profile ${params.settings.profileArn}`)
+                        if (this.profileChangeTokenSource) {
+                            this.profileChangeTokenSource.cancel()
+                            this.profileChangeTokenSource.dispose()
```
viktorsaws, line 181: nit: `dispose` is always called in `finally` block

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -403,6 +424,11 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
         this.resetCodewhispererService()
 
         this.activeIdcProfile = newProfile
+
+        if (token.isCancellationRequested) {
```
viktorsaws, line 428: This cancellation should be before `this.resetCodewhispererService()` call on line 424 to keep change atomic. Otherwise it can leave manager in inconsistent state.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -359,6 +372,10 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
             throw new AmazonQServiceInvalidProfileError('Requested Amazon Q Profile does not exist')
         }
 
+        if (token.isCancellationRequested) {
+            throw new AmazonQServiceProfileUpdateCancelled('Requested profile update got cancelled')
```
viktorsaws, line 376: At line `357` state is set to `PENDING_Q_PROFILE_UPDATE` for the duration of this function.
We need to recover previous state to keep it consistent

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
pras0131: which is right I think as we would want to dispose the related resources after usage.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
pras0131: nice catch!

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
viktorsaws: yeah, I meant that in that case we don't need to call `.dispose()` here, since it is always called in `finally`

viktorsaws (approved):

PR #863, opened 2025-03-24
feat(amazonq): accept extra context for Q Inline Suggestions

imykhai:
## Problem
In some use cases, the client has more information about the context in which the code is being edited. For example, when the editor is run in a web environment on a specific page. Then providing this additional information to Q improves quality of the suggestions. And without the context the recommendations can be barely useful, especially when editing an empty file. 

## Solution
The additional information will be added to the file's left content of the request. If no extra context is passed, the left content is not modified.  
The extra context is passed to the server as workspace configuration under `aws.q.inlineSuggestions.extraContext`. 

Using the supplemental context field was considered, but it was later discovered that it's supported not for all languages. Thus, modifying the left content before making a request is a more reliable approach. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/README.md
+++ b/server/aws-lsp-codewhisperer/README.md
@@ -93,4 +93,4 @@ const params: InitializeParams = {
 ```
 
 ### Extra context for Q Inline Suggestions
-In cases when the client runs in a specific environment that requires customized suggestions, the server supports a `aws.q.extracontext` workspace configuration. This extra context will be passed to the left file content of the request in the beginning of the file.
+In cases when the client runs in a specific environment that requires customized suggestions, the server supports a `aws.q.extraContext` workspace configuration. This extra context will be passed to the left file content of the request in the beginning of the file.
--- a/server/aws-lsp-codewhisperer/README.md
+++ b/server/aws-lsp-codewhisperer/README.md
@@ -91,3 +91,6 @@ const params: InitializeParams = {
   }
 }
 ```
+
+### Extra context for Q Inline Suggestions
+In cases when the client runs in a specific environment that requires customized suggestions, the server supports a `aws.q.extracontext` workspace configuration. This extra context will be passed to the left file content of the request in the beginning of the file.
--- a/server/aws-lsp-codewhisperer/README.md
+++ b/server/aws-lsp-codewhisperer/README.md
@@ -93,4 +93,4 @@ const params: InitializeParams = {
 ```
 
 ### Extra context for Q Inline Suggestions
-In cases when the client runs in a specific environment that requires customized suggestions, the server supports a `aws.q.extraContext` workspace configuration. This extra context will be passed to the left file content of the request in the beginning of the file.
+In cases when the client runs in a specific environment that requires customized suggestions, the server supports a `aws.q.inlineSuggestions.extraContext` workspace configuration. This extra context will be passed to the left file content of the request in the beginning of the file.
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -395,6 +397,10 @@ export const CodewhispererServerFactory =
                     customizationArn: textUtils.undefinedIfEmpty(codeWhispererService.customizationArn),
                 })
 
+                // Add extra context to request context
+                if (extraContext) {
+                    requestContext.fileContext.leftFileContent = extraContext + '\n' + requestContext.fileContext.leftFileContent
```
francescoopiccoli, line 402: to double check, if context exceeds max size the extra context in leftFileContent is the first being cut out/not included, right?

francescoopiccoli (approved):

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
imykhai: That's correct

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -618,6 +624,7 @@ export const CodewhispererServerFactory =
                     // telemetryService.updateEnableTelemetryEventsToDestination(enableTelemetryEventsToDestination)
                     const optOutTelemetryPreference = qConfig['optOutTelemetry'] === true ? 'OPTOUT' : 'OPTIN'
                     telemetryService.updateOptOutPreference(optOutTelemetryPreference)
+                    extraContext = qConfig['extracontext']
```
viktorsaws, line 627: Please use `camelCase` for settings sections for consistency

viktorsaws (changes requested):

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
imykhai: yeah, will update the PR

viktorsaws: Can we use more descriptive name for this new setting? `extracontext` is very generic and can be misused and confusing. Since it is very specific to inline completion, could it be called like "completionContextPrefix" or something more self-explanatory?

And why do we need this prefix at all, what's the use case why we need it?

server/aws-lsp-codewhisperer/README.md
```diff
@@ -91,3 +91,6 @@ const params: InitializeParams = {
   }
 }
 ```
+
+### Extra context for Q Inline Suggestions
+In cases when the client runs in a specific environment that requires customized suggestions, the server supports a `aws.q.extracontext` workspace configuration. This extra context will be passed to the left file content of the request in the beginning of the file.
```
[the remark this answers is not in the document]
imykhai: responded in DM

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
imykhai: explained in the readme

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.test.ts
```diff
@@ -241,6 +241,35 @@ describe('CodeWhisperer Server', () => {
             sinon.assert.notCalled(service.generateSuggestions)
         })
 
+        it.only('should include extra context in recommendation request when extraContext is configured', async () => {
```
akoreman, line 244: You probably want to remove the `only` here to prevent the rest of the tests in this suite getting skipped

imykhai: Ah, good catch!

viktorsaws (approved):

viktorsaws (approved):

PR #865, opened 2025-03-24
chore: basic example how to use openTab on client/server side

volodkevych:
## License

### DIFF
--- a/client/vscode/src/chatActivation.ts
+++ b/client/vscode/src/chatActivation.ts
@@ -4,6 +4,7 @@ import {
     AUTH_FOLLOW_UP_CLICKED,
     CHAT_OPTIONS,
     COPY_TO_CLIPBOARD,
+    UiMessageResultParams,
 } from '@aws/chat-client-ui-types'
 import {
     ChatResult,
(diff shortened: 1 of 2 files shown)


client/vscode/src/chatActivation.ts
```diff
@@ -52,6 +64,35 @@ export function registerChat(languageClient: LanguageClient, extensionUri: Uri,
         languageClient.info(`[VSCode Client] Received telemetry event from server ${JSON.stringify(e)}`)
     })
 
+    languageClient.onRequest(openTabRequestType.method, async (params, _) => {
+        const mapErrorType = (type: string | undefined): number => {
```
kmile, line 68: Is it necessary to re-create this function every request?

client/vscode/src/chatActivation.ts
```diff
@@ -132,12 +173,30 @@ export function registerChat(languageClient: LanguageClient, extensionUri: Uri,
...
                 if (isServerEvent(message.command)) languageClient.sendNotification(message.command, message.params)
                 break
         }
     }, undefined)
 
+    async function waitForOpenTabResult(): Promise<UiMessageResultParams | undefined> {
+        const endTime = Date.now() + 60000 // 1 minute from now
+        const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms))
+        while (Date.now() < endTime) {
+            if (openTabResponseQueue.length > 0) {
```
kmile, line 189: Are tab requests and responses guaranteed to be in order? Can we index them by `requestId` instead?

client/vscode/src/chatActivation.ts
vidkhath: its only an example change so that I can start working on the open tab story

PR #868, opened 2025-03-25
feat(amazonq): return imports as InlineCompletionItemWithReferences to IDE client (import-adder feature)

Will-ShaoHua:
## Problem
To support Import adder feature (business logic is implemented in IDE client)

## License


jpinkney-aws (changes requested):
Looks like you have to update the test that calls this as well

Will-ShaoHua:
btw @jpinkney-aws is there a command or argument i can use for `npm runt test` but only run a specific file or dir other than all tests

Will-ShaoHua:
another question is which formatter we should use in this repo, it seems prettier will mess up the current format and bring in a lot of more code diff

jpinkney-aws:
> btw @jpinkney-aws is there a command or argument i can use for npm runt test but only run a specific file or dir other than all tests? Or something similar to vscode repo we can click on Run and Debug tab

I don't think so -- but i'm sure the flare team would welcome that feature if you wanted to add it 😄 

> another question is which formatter we should use in this repo, it seems prettier will mess up the current format and bring in a lot more code diff

It looks like theres a prettier config in the root, maybe that file just wasn't originally formatted using it?

jpinkney-aws (approved):
LGTM, will need a flare team review as well

zixlin7 (approved):

suprajaven (approved):

PR #869, opened 2025-03-25
fix(amazonq): handle exceptions in TelemetryService

viktorsaws:
## Problem
Calls to telemetry service may crash process when access to CodeWhispererService object are not handled properly.

## Solution
Handle rejections in TelemetryService

## License


patrickvbeurden (approved):
In TelemetryService there are some occurrences of `this.telemetry.emitMetric(...)`, should we add try/catch for those as well (or perhaps add a wrapper function with try/catch that we use instead of directly calling the methods on `this.telemetry`)?

viktorsaws:
> In TelemetryService there are some occurrences of `this.telemetry.emitMetric(...)`, should we add try/catch for those as well (or perhaps add a wrapper function with try/catch that we use instead of directly calling the methods on `this.telemetry`)?

`this.telemetry.emitMetric(...)` is not impacted by this, so we should be good

PR #870, opened 2025-03-25
feat(amazonq): add support for setting profile to null

patrickvbeurden:
## Problem

We received the request to support uninitialising profile selection in `updateConfiguration` by sending empty profile.

## Solution

Instead of listening just for a `string` in `params.settings.profileArn`, we now also listen for `null`. When we receive `null`, the code whisperer service is reset and the state is set back to `PENDING_Q_PROFILE`.

Testing: 
- Added unit tests for transitions:
  - `INITIALIZED` ->(null) `PENDING_Q_PROFILE`
  - `PENDING_Q_PROFILE_UPDATE` ->(null) `PENDING_Q_PROFILE`
- Tested E2E flow with the bundled vs code client

Additional fixes:
- Added missing cancellation of on going profile updates when we receive credentials event
- Altered request cancellation handling to not restore state to avoid race conditions when e.g. cancellation is initiated by credential deletion or null.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
@@ -10,7 +10,6 @@ import {
     LSPErrorCodes,
     SsoConnectionType,
     CancellationToken,
-    BearerCredentials,
     CredentialsType,
     InitializeParams,
     CancellationTokenSource,
@@ -45,7 +44,7 @@ import {
     signalsAWSQDeveloperProfilesEnabled,
 } from './qDeveloperProfiles'
 import { getUserAgent } from '../utilities/telemetryUtils'
-import { getBearerTokenFromProvider } from '../utils'
+import { getBearerTokenFromProvider, isStringOrNull } from '../utils'
 
 export interface Features {
     lsp: Lsp
@@ -159,6 +158,8 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
             if (type === 'iam') {
                 return
             }
+            this.cancelActiveProfileChangeToken()
+
             this.resetCodewhispererService()
             this.connectionType = 'none'
             this.state = 'PENDING_CONNECTION'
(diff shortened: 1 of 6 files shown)


server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -329,18 +333,30 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
...
     }
 
-    private async handleProfileChange(newProfileArn: string, token: CancellationToken): Promise<void> {
+    private cancelActiveProfileChangeToken() {
+        this.profileChangeTokenSource?.cancel()
+        this.profileChangeTokenSource?.dispose()
+        this.profileChangeTokenSource = undefined
+    }
+
+    private handleTokenCancellationRequest(token: CancellationToken) {
```
pras0131, line 342: Based on the updated code here, does it mean we should not be restoring the state?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -363,15 +379,20 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
             this.state = 'PENDING_Q_PROFILE_UPDATE'
         }
 
+        // Client sent an explicit null, indicating they want to reset the assigned profile (if any)
+        if (newProfileArn === null) {
+            this.resetCodewhispererService()
```
viktorsaws, line 384: nit: let's add logs here

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
patrickvbeurden: Discussed offline, but as summary: we should not let the cancelled request handle state modification, it could result in overriding state changes made by the request that initiated the cancellation, which is problematic in e.g. credentials delete flow or setting profile to null (introduced by this PR).

PR #878, opened 2025-03-26
deps: always install current monorepo dependencies

viktorsaws:
## Problem
Some packages declare dependencies on specific version range of other packages in this monorepo. When package version starts zeros for MAJOR and MINOR version parts (e.g. 0.0.1), npm threats them as major versions, and fails to install dependencies when version changes. This breaks automated release process, when packages with `0.0.x` versions are updated, ex. https://github.com/aws/language-servers/actions/runs/14081622706/job/39435699667?pr=851

## Solution

Use `*` to declare dependency version on subpackages in monorepo.

## License

### DIFF
--- a/server/aws-lsp-yaml/package.json
+++ b/server/aws-lsp-yaml/package.json
@@ -27,7 +27,7 @@
     },
     "dependencies": {
         "@aws/language-server-runtimes": "^0.2.48",
-        "@aws/lsp-core": "^0.0.2",
+        "@aws/lsp-core": "*",
         "vscode-languageserver": "^9.0.1",
         "vscode-languageserver-textdocument": "^1.0.8",
         "yaml-language-server": "1.13.0"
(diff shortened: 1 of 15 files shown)


suprajaven (approved):

server/aws-lsp-yaml/package.json
```diff
@@ -27,7 +27,7 @@
     },
     "dependencies": {
         "@aws/language-server-runtimes": "^0.2.48",
-        "@aws/lsp-core": "^0.0.2",
```
natdurlik, line 30: This makes sense locally in monorepo, but we do publish this package. What will happen when client will use version 0.1.1 of @aws/lsp-yaml, but we make backward incompatible changle in lsp-core in the future?

viktorsaws: This is good point, since we distribute compiled Javascript that references dependencies instead of bundling. I'll cancel this PR, let's follow-up. We'd need to either switch to proper Semver versions from `0.0.x` to prevent dependencies problems in future.

PR #881, opened 2025-03-26
feat: create copy of chat server and controller to use in agentic chat

kmile:
## Solution

Starting point for implementing agentic chat, separate from the existing chat for now.

## License

### DIFF
--- a/app/aws-lsp-codewhisperer-runtimes/src/agent-standalone.ts
+++ b/app/aws-lsp-codewhisperer-runtimes/src/agent-standalone.ts
@@ -1,0 +1,29 @@
+import { standalone } from '@aws/language-server-runtimes/runtimes'
+import { RuntimeProps } from '@aws/language-server-runtimes/runtimes/runtime'
+import {
+    CodeWhispererSecurityScanServerTokenProxy,
+    CodeWhispererServerTokenProxy,
+    QAgenticChatServerProxy,
+    QConfigurationServerTokenProxy,
+    QNetTransformServerTokenProxy,
+} from '@aws/lsp-codewhisperer/out/language-server/proxy-server'
+import { IdentityServer } from '@aws/lsp-identity'
+
+const MAJOR = 0
+const MINOR = 1
+const PATCH = 0
+const VERSION = `${MAJOR}.${MINOR}.${PATCH}`
+
+const props: RuntimeProps = {
+    version: VERSION,
+    servers: [
+        CodeWhispererServerTokenProxy,
+        CodeWhispererSecurityScanServerTokenProxy,
+        QConfigurationServerTokenProxy,
+        QNetTransformServerTokenProxy,
+        QAgenticChatServerProxy,
+        IdentityServer.create,
+    ],
+    name: 'AWS CodeWhisperer',
+}
+standalone(props)
--- a/server/aws-lsp-codewhisperer/src/index.ts
+++ b/server/aws-lsp-codewhisperer/src/index.ts
@@ -1,5 +1,6 @@
 export * from './language-server/codeWhispererSecurityScanServer'
 export * from './language-server/codeWhispererServer'
 export * from './language-server/qChatServer'
+export * as QAgenticChatServer from './language-server/agenticChat/qAgenticChatServer'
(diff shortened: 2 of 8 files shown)


jpinkney-aws (approved):
is agenticChatController 1-to-1 with chatController right now? Or were there any required changes?

kmile:
@jpinkney-aws the only change I made is that the agentic server uses the agentic controller.  API changes in `Server` interface coming in next PR.

server/aws-lsp-codewhisperer/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from './language-server/codeWhispererSecurityScanServer'
 export * from './language-server/codeWhispererServer'
 export * from './language-server/qChatServer'
+export * from './language-server/agenticChat/qAgenticChatServer'
```
justinmk3, line 4: off-topic: this means any symbols in any of these modules may conflict with each other. E.g. symbols from `qAgenticChatServer` may conflict with symbols from `qChatServer`. That then requires every symbol to do its own pseudo-namespacing. 

Instead, index.ts files can do this:

```
export * qAgenticChatServer from './language-server/agenticChat/qAgenticChatServer'
```

and that avoids confilcts and pseudo-namespacing, by using an actual namespace.

app/aws-lsp-codewhisperer-runtimes/src/agent-standalone.ts
```diff
@@ -0,0 +1,29 @@
...
+    version: VERSION,
+    servers: [
+        CodeWhispererServerTokenProxy,
+        CodeWhispererSecurityScanServerTokenProxy,
+        QConfigurationServerTokenProxy,
+        QNetTransformServerTokenProxy,
+        QAgenticChatServerProxy,
+        IdentityServer.create,
+    ],
+    name: 'AWS CodeWhisperer',
```
volodkevych, line 27: nit: we can replace `CodeWhisperer` with Q here

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
volodkevych, line 1: Should we also copy tests?

server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts
```diff
@@ -45,4 +46,6 @@ export const QNetTransformServerTokenProxy = QNetTransformServerToken(
 
 export const QChatServerProxy = QChatServer()
 
+export const QAgenticChatServerProxy = QAgenticChatServer()
```
viktorsaws, line 49: nit: I'd avoid adding these exports to `proxy-server.ts` file, ideally we will deprecate it completely

server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        chat.onSendFeedback(params => {
+            return chatController.onSendFeedback(params)
+        })
+
+        chat.onCodeInsertToCursorPosition(params => {
+            return chatController.onCodeInsertToCursorPosition(params)
+        })
+
+        logging.log('Q Chat server has been initialized')
```
volodkevych, line 113: nit: agentic chat? just for clarity

volodkevych:
Would it make sense to move all the shared chat files to a separate directory to make it more clear what functionality is shared between 2 chats?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: Done.

server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts
kmile: Let's do that when we remove the proxy exports.

volodkevych (approved):

PR #882, opened 2025-03-26
feat(amazonq): configuration support for vscode

zixlin7:
## Problem
merging main branch changes into the feature branch, the only new commit is: https://github.com/aws/language-servers/pull/882/commits/a1d43e4cd1d4b71187be34f359acdc02523cd24d and https://github.com/aws/language-servers/pull/882/commits/08426ad7955162ff80de7761a5f8ab9e6df5cb06

1. VS Code currently use a different section for configurations
2. the client name in ideCategory matching logic doesn't match the actual VS Code client name
3. Import currently doesn't take into consideration user settings

## Solution

1. update the logic to accommodate VSC configurations in language server
2. fix the client name for VS Code 
3. respect user's import settings

## License


server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -640,8 +641,21 @@ export const CodewhispererServerFactory =
...
-        const updateConfiguration = async () => {
+        const updateConfiguration = async (params?: DidChangeConfigurationParams) => {
             try {
+                const ideCategory = telemetryService.getUserContext()?.ideCategory
+                if (ideCategory === 'VSCODE') {
+                    const vscConfig = await lsp.workspace.getConfiguration(Q_CONFIGURATION_VSC)
+                    const optOutTelemetryPreference = vscConfig['shareContentWithAWS'] === true ? 'OPTOUT' : 'OPTIN'
+                    telemetryService.updateOptOutPreference(optOutTelemetryPreference)
+                    includeSuggestionsWithCodeReferences = vscConfig['showCodeWithReferences'] === true
+                    codeWhispererService.customizationArn = undefinedIfEmpty(params?.settings['customization'])
```
viktorsaws, line 652: There are more places that rely on reading there configuration in latest codebase. The target branch of this PR seems to be not up to date.

zixlin7: thanks for that info! Do we need to manually merge in main branch to the feature branches?

jpinkney-aws: yeah, IIRC flare doesn't have the github bot that the toolkits team has

zixlin7: Got it! I can pick up the additional change when I merge in new changes from mainline. We have a blocker right now with merging in main that was discussed offline and this change can be decoupled for now?

jpinkney-aws:
@zixlin7 is there any specific files you've changed that we need to take a look at for the merge?

zixlin7:
I think just reviewing the new commits should be sufficient for now. In case of unexpected merge error, there probably would be test failure and we should be able to catch it when we merge this feature branch back to main.


> @zixlin7 is there any specific files you've changed that we need to take a look at for the merge?

jpinkney-aws (approved):

viktorsaws:
It would be much easier to review, If actual functional changes would be in separate PR from rebases.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -0,0 +1,687 @@
...
+
+            return
+        }
+
+        this.logServiceState('Unknown Connection state')
+    }
+
+    public async handleDidChangeConfiguration(ideCategory?: string, params?: DidChangeConfigurationParams) {
+        try {
+            if (ideCategory === 'VSCODE') {
```
viktorsaws, line 287: I'd prefer not to add custom client-specific handling to Language Server, and delegate it to client to respond with expected configuration and read from client-specific configuration store. Adding more modes make it harder to test and manage language server logic.

VsCode extension could register custom handler on `ConfigurationRequest` method on LanguageClient and respond with what server expects following existing API. 

```
    client.onRequest(ConfigurationRequest.method, (section) => {
        // Get configuration from VSCode and return
    })
```

viktorsaws (changes requested):

server/aws-lsp-codewhisperer/src/language-server/utilities/telemetryUtils.ts
```diff
@@ -47,7 +47,7 @@ export const getUserAgent = (initializeParams: InitializeParams, serverInfo?: Se
 }
 
 const IDE_CATEGORY_MAP: { [key: string]: IdeCategory } = {
-    'AmazonQ-For-VSCode': 'VSCODE',
+    'AWS IDE Extensions for VSCode': 'VSCODE',
```
viktorsaws, line 50: This value should not be changed, because this format was specifically requested. Was there a change in the requirements from BI?

zixlin7: Could you help me understand how does this affect BI if we are just mapping client name to 'VSCODE'? I don't see a reverse mapping where we send `AmazonQ-For-VSCode` to telemetry. 
I can instead change the VS Code extension name to match this one.

zixlin7:
> It would be much easier to review, If actual functional changes would be in separate PR from rebases.

It would be easier if we can have auto merge to feature branches, similar to all toolkits repos

PR #883, opened 2025-03-27
feat: bundling and infra to deploy webworker in webpage on local webserver

francescoopiccoli:
## Problem
We want to check whether webworker bundling of lsp-codewhisperer iam fails at **runtime** time to prevent breaking it without noticing it, we already check for build time issue (https://github.com/aws/language-servers/pull/833). 

In this PR we create the infrastructure needed to run the webworker on a webpage locally on browser with a webpack dev server. This is needed to then create tests to do the check to enforce a functioning server on browser before merging.

As this monorepo uses commonjs, i had to override as use es module is tsconfig, as commonjs is not compatible with browser.

tests PR: https://github.com/aws/language-servers/pull/887

## Solution
Summary:
- Fix current bundling issues (mock-fs)
- Fix package to work on browser (tsconfig and webpack to use es module instead of commonjs)
- Create a basic webpage template public/index.html and host it on local webserver with webpack (devserver)
- Create a language client and setup initialization of language server inside webworker
- update doc

you can check for the correct functioning of webworker and language server in browser by:
```npm run package && npm run start``` and checking on `http://127.0.0.1:8080` that there are no console errors and a log with `Server initialized` logging initialization results.

## License

### DIFF
--- a/app/aws-lsp-codewhisperer-runtimes/tsconfig.json
+++ b/app/aws-lsp-codewhisperer-runtimes/tsconfig.json
@@ -2,7 +2,12 @@
     "extends": "../../tsconfig.packages.json",
     "compilerOptions": {
         "rootDir": "./src",
-        "outDir": "./out"
+        "outDir": "./out",
+        "esModuleInterop": true,
+        "allowJs": true,
+        "checkJs": true,
+        "module": "ESNext",
+        "target": "ES2022"
     },
     "include": ["src"]
 }
--- a/app/aws-lsp-codewhisperer-runtimes/README.md
+++ b/app/aws-lsp-codewhisperer-runtimes/README.md
@@ -24,3 +24,18 @@ or
 ```bash
 node ./out/iam-standalone.js --stdio
 ```
+
+## Development and Testing Web Worker Implementation
+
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Review by imykhai — APPROVED** [review]

File: app/aws-lsp-codewhisperer-runtimes/tsconfig.json [hunk 1]
```diff
@@ -2,7 +2,12 @@
     "extends": "../../tsconfig.packages.json",
     "compilerOptions": {
         "rootDir": "./src",
-        "outDir": "./out"
+        "outDir": "./out",
+        "esModuleInterop": true,
```
viktorsaws, line 6: This Typescript configuration breaks compiled standalone servers. When running `CodeWhisperer Server Token` command from VSCode `.launch.json`, it fails with 

```
node:76518) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/Users/vshcherb/Projects/Flare/language-servers/app/aws-lsp-codewhisperer-runtimes/out/token-standalone.js:1
import { standalone } from '@aws/language-server-runtimes/runtimes';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (node:internal/modules/cjs/loader:1378:20)
```

PR #886, opened 2025-03-28
feat(amazonq): optimize service manager reuse

patrickvbeurden:
## Problem

We currently need to maintain a fallback codewhisperer service in the inline completions server because it supports both token and iam credentials. Refactoring this server would enable us to optimize the reuse of the new amazon q service manager.

## Solution

- Centralize workspace configurations within the service manager
    - By adding the `getConfiguration(): Readonly<AWSQWorkspaceConfigs>` method to `BaseAmazonQServiceManager` we can refer to the up-to-date workspace configurations maintained within the service manager
    - Introduce an `AWSQConfigurationCache` wrapper which orchestrates manipulation of the cached configuration within the service manager in a type-safe manner
    - Factor our handling configuration update events to a reusable function
    - Factor out region and endpoint discovery to reusable function
- Introduce a factory to initialize a base service manager based on the server's credentials type 

Testing: existing unit tests pass, added more unit tests

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
@@ -48,6 +48,7 @@ import {
 } from '../../shared/amazonQServiceManager/errors'
 import { AmazonQTokenServiceManager } from '../../shared/amazonQServiceManager/AmazonQTokenServiceManager'
 import { TelemetryService } from '../../shared/telemetry/telemetryService'
+import { AmazonQWorkspaceConfig } from '../../shared/amazonQServiceManager/configurationUtils'
 
 type ChatHandlers = Omit<
     LspHandlers<Chat>,
@@ -562,24 +563,18 @@ export class ChatController implements ChatHandlers {
         return chatEventParser.getResult()
     }
 
-    updateConfiguration = async () => {
-        try {
-            const qConfig = await this.#features.lsp.workspace.getConfiguration(Q_CONFIGURATION_SECTION)
-            if (qConfig) {
-                this.#customizationArn = textUtils.undefinedIfEmpty(qConfig.customization)
-                this.#log(`Chat configuration updated to use ${this.#customizationArn}`)
-                /*
-                    The flag enableTelemetryEventsToDestination is set to true temporarily. It's value will be determined through destination
-                    configuration post all events migration to STE. It'll be replaced by qConfig['enableTelemetryEventsToDestination'] === true
-                */
-                // const enableTelemetryEventsToDestination = true
-                // this.#telemetryService.updateEnableTelemetryEventsToDestination(enableTelemetryEventsToDestination)
-                const optOutTelemetryPreference = qConfig['optOutTelemetry'] === true ? 'OPTOUT' : 'OPTIN'
-                this.#telemetryService.updateOptOutPreference(optOutTelemetryPreference)
-            }
-        } catch (error) {
-            this.#log(`Error in GetConfiguration: ${error}`)
-        }
+    updateConfiguration = (newConfig: AmazonQWorkspaceConfig) => {
+        this.#customizationArn = newConfig.customizationArn
+        this.#log(`Chat configuration updated to use ${this.#customizationArn}`)
+        /*
+            The flag enableTelemetryEventsToDestination is set to true temporarily. It's value will be determined through destination
+            configuration post all events migration to STE. It'll be replaced by qConfig['enableTelemetryEventsToDestination'] === true
+        */
+        // const enableTelemetryEventsToDestination = true
+        // this.#telemetryService.updateEnableTelemetryEventsToDestination(enableTelemetryEventsToDestination)
+        const updatedOptOutPreference = newConfig.optOutTelemetryPreference
+        this.#telemetryService.updateOptOutPreference(updatedOptOutPreference)
+        this.#log(`Chat configuration telemetry preference to ${updatedOptOutPreference}`)
     }
 
     #log(...messages: string[]) {
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -45,14 +45,13 @@ import { Metric } from '../../shared/telemetry/metric'
 import { getErrorMessage, isAwsError, isNullish, isObject } from '../../shared/utils'
 import { QChatTriggerContext, TriggerContext } from '../chat/contexts/triggerContext'
 import { HELP_MESSAGE } from '../chat/constants'
-import { Q_CONFIGURATION_SECTION } from '../configuration/qConfigurationServer'
-import { textUtils } from '@aws/lsp-core'
 import { TelemetryService } from '../../shared/telemetry/telemetryService'
 import {
     AmazonQServicePendingProfileError,
     AmazonQServicePendingSigninError,
 } from '../../shared/amazonQServiceManager/errors'
 import { AmazonQTokenServiceManager } from '../../shared/amazonQServiceManager/AmazonQTokenServiceManager'
+import { AmazonQWorkspaceConfig } from '../../shared/amazonQServiceManager/configurationUtils'
 
 type ChatHandlers = Omit<
     LspHandlers<Chat>,
@@ -567,24 +566,18 @@ export class AgenticChatController implements ChatHandlers {
         return chatEventParser.getResult()
     }
 
-    updateConfiguration = async () => {
-        try {
-            const qConfig = await this.#features.lsp.workspace.getConfiguration(Q_CONFIGURATION_SECTION)
-            if (qConfig) {
-                this.#customizationArn = textUtils.undefinedIfEmpty(qConfig.customization)
-                this.#log(`Chat configuration updated to use ${this.#customizationArn}`)
-                /*
-                    The flag enableTelemetryEventsToDestination is set to true temporarily. It's value will be determined through destination
-                    configuration post all events migration to STE. It'll be replaced by qConfig['enableTelemetryEventsToDestination'] === true
-                */
-                // const enableTelemetryEventsToDestination = true
-                // this.#telemetryService.updateEnableTelemetryEventsToDestination(enableTelemetryEventsToDestination)
-                const optOutTelemetryPreference = qConfig['optOutTelemetry'] === true ? 'OPTOUT' : 'OPTIN'
-                this.#telemetryService.updateOptOutPreference(optOutTelemetryPreference)
-            }
-        } catch (error) {
-            this.#log(`Error in GetConfiguration: ${error}`)
-        }
+    updateConfiguration = (newConfig: AmazonQWorkspaceConfig) => {
+        this.#customizationArn = newConfig.customizationArn
+        this.#log(`Chat configuration updated to use ${this.#customizationArn}`)
+        /*
+            The flag enableTelemetryEventsToDestination is set to true temporarily. It's value will be determined through destination
+            configuration post all events migration to STE. It'll be replaced by qConfig['enableTelemetryEventsToDestination'] === true
+        */
+        // const enableTelemetryEventsToDestination = true
+        // this.#telemetryService.updateEnableTelemetryEventsToDestination(enableTelemetryEventsToDestination)
+        const updatedOptOutPreference = newConfig.optOutTelemetryPreference
+        this.#telemetryService.updateOptOutPreference(updatedOptOutPreference)
+        this.#log(`Chat configuration telemetry preference to ${updatedOptOutPreference}`)
     }
 
     #log(...messages: string[]) {
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts
@@ -41,8 +41,8 @@ export const QAgenticChatServer =
         })
 
         const updateConfigurationHandler = async () => {
-            await amazonQServiceManager.handleDidChangeConfiguration()
-            await chatController.updateConfiguration()
+            const amazonQConfig = await amazonQServiceManager.handleDidChangeConfiguration()
+            chatController.updateConfiguration(amazonQConfig)
         }
 
         lsp.onInitialized(async () => {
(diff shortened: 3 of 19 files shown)


server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+        awsQRegion = region
+        // @ts-ignore
+        awsQEndpoint = AWS_Q_ENDPOINTS[awsQRegion]
+    } else {
+        const runtimeRegion = runtime.getConfiguration(AWS_Q_REGION_ENV_VAR)
+
+        if (runtimeRegion) {
+            logging.log(`Selecting region (found: ${runtimeRegion}) provided by runtime`)
+            awsQRegion = runtimeRegion
+            // @ts-ignore
```
viktorsaws, line 39: nit: is there a way we can stop `ts-ignore` in this function?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+export interface AWSQWorkspaceConfigs {
+    qConfig: QConfig
+    codeWhispererConfig: CodeWhispererConfig
+}
+
+export const CODE_WHISPERER_CONFIGURATION_SECTION = 'aws.codeWhisperer'
+
+export async function getAWSQRelatedWorkspaceConfigs(
```
viktorsaws, line 82: nit: better to use `AmazonQ` instead of `AWSQ` for consistency

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+        if (newQConfig) {
+            qConfig = {
+                customizationArn: textUtils.undefinedIfEmpty(newQConfig.customization),
+                optOutTelemetryPreference: newQConfig['optOutTelemetry'] === true ? 'OPTOUT' : 'OPTIN',
+                inlineSuggestions: {
+                    extraContext: textUtils.undefinedIfEmpty(newQConfig.inlineSuggestions?.extraContext),
+                },
+            }
+
+            cache?.setConfig('qConfig', qConfig)
```
viktorsaws, line 103: I'd move cache manipulation out of this function and make it responsibility on calling code. This way this method stays pure function with no side-effects.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+        extraContext: string | undefined // aws.q.inlineSuggestions.extraContext
+    }
+}
+
+export interface CodeWhispererConfig {
+    includeSuggestionsWithCodeReferences: boolean // aws.codeWhisperer.includeSuggestionsWithCodeReferences - return suggestions with code references
+    shareCodeWhispererContentWithAWS: boolean // aws.codeWhisperer.shareCodeWhispererContentWithAWS - share content with AWS
+}
+
+export interface AWSQWorkspaceConfigs {
```
volodkevych, line 75: I think we can put everything under single Q config. `aws.codeWhisperer` is just a legacy name for `aws.q`.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/factories.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export const initBaseServiceManager = (credentialsType: CredentialsType) => {
+    return (features: Features): BaseAmazonQServiceManager => {
+        if (credentialsType === 'bearer') {
+            features.logging.debug('Initializing token service manager')
+            return initTokenServiceManager(features)
+        } else {
+            features.logging.debug(`Falling back to IAM service manager (found credentialsType: ${credentialsType}`)
+            return initFallbackServiceManager(features)
```
viktorsaws, line 62: Given that we support only 2 kinds of services (iam and bearer), I think we simplify this factory and always explicitly initialize `bearer` or `iam` variants. I don't expect there'd be any new implementations soon, and if they come we can extend this factory.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -604,68 +564,32 @@ export const CodewhispererServerFactory =
...
-                logging.log(
-                    `Configuration updated to ${includeSuggestionsWithCodeReferences ? 'include' : 'exclude'} suggestions with code references`
-                )
-
-                fallbackCodeWhispererService.shareCodeWhispererContentWithAWS =
-                    config['shareCodeWhispererContentWithAWS'] === true
-                logging.log(
-                    `Configuration updated to ${fallbackCodeWhispererService.shareCodeWhispererContentWithAWS ? 'share' : 'not share'} Amazon Q content with AWS`
-                )
+                const { qConfig } = amazonQServiceManager.getConfiguration()
```
viktorsaws, line 569: Would it make sense to make `amazonQServiceManager.handleDidChangeConfiguration()` return new configuration values, instead of querying them here?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/factories.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    const configurationCache = new AWSQConfigurationCache()
+
+    return {
+        handleDidChangeConfiguration: async () => {
+            try {
+                const { qConfig, codeWhispererConfig } = await getAWSQRelatedWorkspaceConfigs(
+                    features.lsp,
+                    features.logging,
+                    configurationCache
```
volodkevych, line 31: nit: we pass`configurationCache` inside of the method just to set config into the cache, but we also return the config here, so we can just set the cache in this function - should be easier to understand the code

patrickvbeurden: Removed cache as a parameter and made caching responsibility of the caller (see Viktor's comment)

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
patrickvbeurden: Yes, by changing `AWS_Q_ENDPOINTS` to a `Map`. Alternatives are (1)`Record<string, string | undefined>`, but then during iteration we always need to check for `undefined`, or (2) enable `noUncheckedIndexedAccess`, which forces us to always check indexes for undefined, but that quickly increases the scope of refactor.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
patrickvbeurden: Yes that would be nice - added it. However, it would also be good to keep `getConfiguration` for cases where we don't need to trigger an update.

server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts
```diff
@@ -604,68 +564,32 @@ export const CodewhispererServerFactory =
...
-                // Fallback to default passed service factory for IAM credentials type
-                logging.log('Initialized Completion server with default service type')
-                amazonQServiceManager = {
-                    handleDidChangeConfiguration: async () => {},
-                    getCodewhispererService: () => {
-                        return fallbackCodeWhispererService
-                    },
-                }
-            }
+            amazonQServiceManager = serviceManager({
```
viktorsaws, line 585: I really like how initialization and handling of Configuration is simplified and consolidated in 1 place.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -454,24 +453,18 @@ export class ChatController implements ChatHandlers {
...
-                */
-                // const enableTelemetryEventsToDestination = true
-                // this.#telemetryService.updateEnableTelemetryEventsToDestination(enableTelemetryEventsToDestination)
-                const optOutTelemetryPreference = qConfig['optOutTelemetry'] === true ? 'OPTOUT' : 'OPTIN'
-                this.#telemetryService.updateOptOutPreference(optOutTelemetryPreference)
-            }
-        } catch (error) {
-            this.#log(`Error in GetConfiguration: ${error}`)
-        }
+    updateConfiguration = (newConfig: AmazonQWorkspaceConfig) => {
```
viktorsaws, line 456: Nice change to reduce round-trips to Client for configuration and decouple code from LSP client more.

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/factories.ts
```diff
@@ -0,0 +1,61 @@
...
+    getAmazonQRegionAndEndpoint,
+    getAmazonQRelatedWorkspaceConfigs,
+} from './configurationUtils'
+import { CodeWhispererServiceBase, CodeWhispererServiceIAM } from '../codeWhispererService'
+
+const initTokenServiceManager = (features: Features) => {
+    return AmazonQTokenServiceManager.getInstance(features)
+}
+
+export const initFallbackServiceManager = (features: Features, serviceOverride?: CodeWhispererServiceBase) => {
```
viktorsaws, line 15: I'd remove `serviceOverride` and make this explicitly IAM manager `initIAMServiceManager`

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
```diff
@@ -0,0 +1,181 @@
...
+        } else {
+            logging.log('Region not provided by caller or runtime, falling back to default region and endpoint')
+            amazonQRegion = DEFAULT_AWS_Q_REGION
+            amazonQEndpoint = DEFAULT_AWS_Q_ENDPOINT_URL
+        }
+    }
+
+    if (!amazonQEndpoint) {
+        logging.log(
+            `Unable to determine endpoint (found: ${amazonQEndpoint}) for region: ${amazonQRegion}, falling back to default region and endpoint`
```
viktorsaws, line 52: nit: let's log default region here, so it's clear if we need to debug this codepath

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
```diff
@@ -0,0 +1,181 @@
...
+        amazonQEndpoint = DEFAULT_AWS_Q_ENDPOINT_URL
+    }
+
+    return {
+        region: amazonQRegion,
+        endpoint: amazonQEndpoint,
+    }
+}
+
+interface QInlineSuggestions {
```
viktorsaws, line 64: supernit: I'd call this `QInlineSuggestionsConfig` to avoid confusion with functional code

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    return {
+        ...qConfig,
+        ...codeWhispererConfig,
+    }
+}
+
+export const defaultAmazonQWorkspaceConfigFactory = (): AmazonQWorkspaceConfig => {
+    return {
```
viktorsaws, line 145: It would be nice to document all supported configuration sections and their meaning in README. Not critical though, can be a follow-up

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/factories.ts
```diff
@@ -0,0 +1,61 @@
...
+            features.workspace,
+            region,
+            endpoint,
+            features.sdkInitializator
+        )
+
+    const configurationCache = new AmazonQConfigurationCache()
+
+    return {
+        handleDidChangeConfiguration: async (): Promise<Readonly<AmazonQWorkspaceConfig>> => {
```
viktorsaws, line 30: The logic for handling configuration gonna be the same for both Token and IAM based service managers, and should be kept in sync in future. Can we extract this into reusable method or maybe make a base class for Managers?

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -283,54 +281,32 @@ export class AmazonQTokenServiceManager implements BaseAmazonQServiceManager {
 
     public async handleDidChangeConfiguration() {
```
viktorsaws, line 282: I think we'll need to optimize calls to `handleDidChangeConfiguration`. Given that now `handleDidChangeConfiguration` is hooked into every consuming Server, it is called several times simultaneously. 
We didn't observe functional side-effect of this, but it will be better if we allow only 1 inflight call at a time. Today, it is called at least 4 times from 4 different servers

patrickvbeurden: I started working on a follow-up PR for this! It would be great to combine this together with creating a base class for the service manager

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/factories.ts
patrickvbeurden: Agreed, I think a base class would be ideal, is it ok to follow up in next PR, I think it would be good to combine with centralizing the `didChangeConfiguration` handling

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/configurationUtils.ts
patrickvbeurden: Will follow up

server/aws-lsp-codewhisperer/src/language-server/amazonQServiceManager/factories.ts
patrickvbeurden: Will follow up, this goes well with creating a base class in follow up PR

viktorsaws (approved):

PR #894, opened 2025-03-31
feat: port fs related tools from VSC. 

Hweinstock:
## Problem
Some agent tools were implemented in the VSC codebase. However, agentic chat is moving to use Flare implementation, so these tools should live here instead.

relevant: https://github.com/aws/language-server-runtimes/pull/393
## Solution
- Port [fsRead](https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/codewhispererChat/tools/fsRead.ts#L17) and [fsWrite](https://github.com/aws/aws-toolkit-vscode/blob/10bb1c7dc45f128df14d749d95905c0e9b808096/packages/core/src/codewhispererChat/tools/fsWrite.ts#L42).
- Ported from: 
   - https://github.com/aws/aws-toolkit-vscode/blob/10bb1c7dc45f128df14d749d95905c0e9b808096/packages/core/src/codewhispererChat/tools/fsWrite.ts#L42
   - https://github.com/aws/aws-toolkit-vscode/blob/10bb1c7dc45f128df14d749d95905c0e9b808096/packages/core/src/codewhispererChat/tools/fsRead.ts#L17
 - refactor error handling to allow errors to bubble up. 

## Testing 
- Avoid mocking the file system by adding a new test utility `TestFolder`. This is a simplified version of a [this](https://github.com/aws/aws-toolkit-vscode/blob/16477869525fb79f8dc82cb22e301aaea9c5e0c6/packages/core/src/test/testUtil.ts#L77) utility from the VSC codebase. 

## Future Work
- Update runtime dependency to `^0.2.49`, and add the tools to the agentic chat server [here](https://github.com/aws/language-servers/blob/1a84c5e4dd40ed564afcb457028ea49b535441d4/server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts#L19) 
- Does it make sense to move testing utilities to [runtimes](https://github.com/aws/language-server-runtimes/tree/main/runtimes/testing)? Should a non-mocked fs workspace be included?

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
```diff
@@ -0,0 +1,143 @@
...
+        this.fsPath = sanitized
+
+        let fileExists: boolean
+        try {
+            fileExists = await this.workspace.fs.exists(this.fsPath)
+            if (!fileExists) {
+                throw new Error(`Path: "${this.fsPath}" does not exist or cannot be accessed.`)
+            }
+        } catch (err) {
+            throw new Error(`Path: "${this.fsPath}" does not exist or cannot be accessed. (${err})`)
```
nkomonen-amazon, line 47: Where does workspace.fs come from? Assuming it is well written, it will already have a usable error code and message. So something like this seems redundant.

nkomonen-amazon:
@Hweinstock 

Any idea on what the actual implementation of `workspace.fs` is? Is it just the node fs?

> Avoid mocking the file system by adding a new test utility TestFolder

Mocking the filesystem for tests seems like a code smell, there isn't a strong reason to not just test against the filesystem. So I'm in agreement with your decision to not mock the filesystem and just have them use the TestFolder util.

> Does it make sense to move testing utilities to runtimes

What do you mean by this?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
Hweinstock: My understanding is that this is injected depending on the environment. For the common (non-web) use case, I believe its implemented here: https://github.com/aws/language-server-runtimes/blob/b31f851d7f16d290f29c93281d6646b50820b7fb/runtimes/runtimes/standalone.ts#L174 which doesn't have any error reporting or logging.

core/aws-lsp-core/src/test/testFolder.ts
```diff
@@ -0,0 +1,60 @@
...
+    private constructor(public readonly path: string) {}
+
+    async write(fileName: string, content: string): Promise<string> {
+        const filePath = path.join(this.path, fileName)
+        fs.writeFileSync(filePath, content)
+        return filePath
+    }
+
+    static async create() {
+        const tempDir = path.join(
```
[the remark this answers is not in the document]
Hweinstock: Yes, I agree. I noticed the VS Code implementation does this, and it would be a great feature. We do this by wrapping all the tests with our own test runner in VSC, but I don't see similar logic here yet, so this change would require some ground work.

Hweinstock:
> any idea why we have separate fsRead/Write files? I would have assumed it is better to have a single module with all FS related utils. In this case a wrapper over the basic fs module, but with some more complex utilities.

Not entirely sure why they made this design decision. My best guess is that they want to enforce the paradigm of each tool have a singular purpose. A general `filesystem` access tool would have many purposes. 

> Mocking the filesystem for tests seems like a code smell, there isn't a strong reason to not just test against the filesystem. So I'm in agreement with your decision to not mock the filesystem and just have them use the TestFolder util.

I agree. The number of filesystem quirks we've observed in VSC shows how easily bugs can slip through (especially on windows). 

> What do you mean by this?
Added a link to the description. I didn't see any testing utilities other than this `testing` folder in runtimes, so I am unsure if thats where this stuff should live.

jpinkney-aws:
>> any idea why we have separate fsRead/Write files? I would have assumed it is better to have a single module with all FS related utils. In this case a wrapper over the basic fs module, but with some more complex utilities.

> Not entirely sure why they made this design decision. My best guess is that they want to enforce the paradigm of each tool having a singular purpose. Separating it also seems like an easy way to manage granular access to the filesystem by only adding certain tools to an agent.

I believe its because agentic chat knows how to use tools is by looking up descriptions in the the [tool_index.json](https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/codewhispererChat/tools/tool_index.json#L2). The description/input for fsRead/fsWrite are different, so it would be harder to merge them together

jpinkney-aws (approved):
I'd probably also call out what commit you copied from in the description just so its easy to find later if you're looking at PR's

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
```diff
@@ -0,0 +1,143 @@
...
+    readRange?: number[]
+}
+
+export class FsRead {
+    private fsPath: string
+    private readonly readRange?: number[]
+    private readonly logging: Features['logging']
+    private readonly workspace: Features['workspace']
+
+    constructor(features: Pick<Features, 'workspace' | 'logging'>, params: FsReadParams) {
```
kmile, line 24: Why `Pick`, is there a reason to scope down the parameter type here?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
kmile: Correct. This is also why you get the linting errors. We try to avoid using Node platform specific filesystem libraries to make portability easier. If that is not possible in your use case, let's see if we can update `workspace.fs` to meet your requirements, or relax the linting rules for these specific tools.

Just know that if we do the latter, these tools will not be portable from one environment to the next, and that needs to be documented somewhere.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
```diff
@@ -0,0 +1,143 @@
...
+                throw new Error(`Path: "${this.fsPath}" does not exist or cannot be accessed.`)
+            }
+        } catch (err) {
+            throw new Error(`Path: "${this.fsPath}" does not exist or cannot be accessed. (${err})`)
+        }
+
+        this.logging.log(`Validation succeeded for path: ${this.fsPath}`)
+    }
+
+    public queueDescription(updates: Writable): void {
```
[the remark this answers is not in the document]
Hweinstock: It currently does not, but that seems like a useful feature. Original implementation lives [here](https://github.com/aws/aws-toolkit-vscode/blob/10bb1c7dc45f128df14d749d95905c0e9b808096/packages/core/src/codewhispererChat/tools/fsRead.ts#L50).  I see an example of how this could be done in Flare [here](https://github.com/aws/language-servers/blob/7ef13b585130e264f4fa9a2ba4fae2e923fb940a/server/aws-lsp-identity/src/sso/deviceCode/deviceCodeFlow.ts#L156), which would require changes to the runtime to add something like [this](https://github.com/aws/language-server-runtimes/blob/95cab0adac81d146d8988a7eb950d307b8947574/runtimes/protocol/identity-management.ts#L171) for file read progress. If that sounds like a good example to base this on I can mark it as a follow-up to add this for file read and write progress.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
Hweinstock: My intention was require as little as possible from the caller. However, this does force them to explicitly scope `Features` down so I changed it to `Pick<Features, 'workspace' | 'logging'> & Partial<Features>` to avoid this step.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
Hweinstock: I refactored it to avoid using these modules. I believe that relaxing the lint rules for specific tools could be problematic because it ignores indirect dependencies. For example, we use `sanitize` from `@aws/lsp-core/out/util/path` which requires these modules. It does have a [note](https://github.com/aws/language-servers/blob/7ef13b585130e264f4fa9a2ba4fae2e923fb940a/core/aws-lsp-core/src/util/path.ts#L1) at the top of the file, but lmk if I can document this elsewhere as well.

core/aws-lsp-core/src/index.ts
```diff
@@ -14,3 +14,4 @@ export * as textUtils from './util/text'
 export * as timeoutUtils from './util/timeoutUtils'
 export * from './util/awsError'
 export * from './base/index'
+export * from './test/testFolder'
```
justinmk3, line 17: Is it not likely that all these non-aliased `*` exports will eventually conflict? And that would then require quasi-namespacing. Instead of quasi-namespacing, we can solve the problem easily by aliasing, like this:

```suggestion
export * as testFolder from './test/testFolder'
```

core/aws-lsp-core/src/test/testFolder.ts
```diff
@@ -0,0 +1,60 @@
+import * as path from 'path'
+import * as fs from 'fs'
+import * as os from 'os'
+import * as crypto from 'crypto'
+
+/**
+ * Interface for working with temporary files.
+ * Simplified port of https://github.com/aws/aws-toolkit-vscode/blob/16477869525fb79f8dc82cb22e301aaea9c5e0c6/packages/core/src/test/testUtil.ts#L77
```
justinmk3, line 8: 👍 for now, this kind of precise cross-reference will help a lot, in each of these util files.

core/aws-lsp-core/src/util/path.ts
```diff
@@ -78,3 +78,16 @@ export function normalize(p: string): string {
     }
     return normalizeSeparator(path.normalize(p))
 }
+
+export function sanitize(inputPath: string): string {
```
justinmk3, line 82: In aws-toolkit-vscode this santize() was added recently, but I don't get why it was added: https://github.com/aws/aws-toolkit-vscode/pull/6840#discussion_r2018996426

Why is this needed instead of [our normalize()](https://github.com/aws/aws-toolkit-vscode/blob/708f5381c7627d06a6f63552b3f8074672351ed1/packages/core/src/shared/utilities/pathUtils.ts#L92) ?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
kmile: That makes sense. This would work or passing in the `Feature` that you need explicitly. Glad you're not taking a dependency on the whole `Features` set to avoid coupling everything together.

In order to implement the tool invocation I think you can remove the `params` here, since in the `Agent` feature the tool creation is separate from the invocation: https://github.com/aws/language-server-runtimes/tree/main/runtimes#agent-tools

core/aws-lsp-core/src/util/path.ts
Hweinstock: It looks like they do slightly different things that definitely overlap with `normalize`. For example, normalize doesn't remove white space or expand `~`. I agree they definitely overlap and could potentially be refactored to make this more clear.

imykhai:
One of the commits - a merge from `main` to the branch - doesn't comply with conventional commits.

nkomonen-amazon (approved):

PR #897, opened 2025-04-01
feat: Add support for inline chat

pras0131:
The PR introduces the changes to add support for inline chat.

## License


server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -54,17 +54,20 @@ export class ChatController implements ChatHandlers {
...
     #customizationArn?: string
     #telemetryService: TelemetryService
+    #amazonQServiceManager?: AmazonQTokenServiceManager
 
     constructor(
         chatSessionManagementService: ChatSessionManagementService,
         features: Features,
-        telemetryService: TelemetryService
+        telemetryService: TelemetryService,
+        amazonQServiceManager?: AmazonQTokenServiceManager
```
imykhai, line 63: We need the chat server compatible with the IAM credentials too. 
There is an example - https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts#L649

pras0131: we can make those changes as a separate follow up PR, if not urgent?

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
```diff
@@ -29,6 +29,53 @@ export class QChatTriggerContext {
...
+        params: InlineChatParams,
+        triggerContext: TriggerContext,
+        customizationArn?: string,
+        profileArn?: string
+    ): SendMessageCommandInput {
+        const { prompt } = params
+
+        const data: SendMessageCommandInput = {
+            conversationState: {
+                chatTriggerType: ChatTriggerType.INLINE_CHAT,
```
volodkevych, line 51: `chatTriggerType` is the only field that is different from `getChatParamsFromTrigger` method implementation. Can we reuse existing method?

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -215,7 +218,63 @@ export class ChatController implements ChatHandlers {
         params: InlineChatParams,
         token: CancellationToken
     ): Promise<ChatResult | ResponseError<ChatResult>> {
-        return {}
+        // TODO: This metric needs to be removed later, just added for now to be able to create a ChatEventParser object
```
volodkevych, line 221: can we remove it now?

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
pras0131: yes makes sense, I'll make the change.

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
```diff
@@ -29,6 +29,53 @@ export class QChatTriggerContext {
         }
     }
 
+    async getNewTriggerContextForInlineChat(params: InlineChatParams): Promise<TriggerContext> {
```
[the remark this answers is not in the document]
pras0131: yeah I checked since we are not using the tabId from ChatParams in the existing method, I can reuse the existing method.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
pras0131: Since ChatEventParser needs metric object as mandatory, hence kept it as is for now intentionally. The ChatEventParser needs to be refactored, which we can take up as a follow up?

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
volodkevych, line 1: As discussed, this also needs to be supported in `qAgenticChatServer` as this is the future version of our chat server.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -215,7 +218,63 @@ export class ChatController implements ChatHandlers {
...
+        const metric = new Metric<AddMessageEvent>({
+            cwsprChatConversationType: 'Chat',
+        })
+        const triggerContext = await this.#getInlineChatTriggerContext(params)
```
volodkevych, line 225: Have you considered an option to continue using `sendMessage` within a chatSession (as we use it for regular chat prompt) with a fake uuid instead of `tabId`? We would need a session per inline chat prompt. I'm thinking about code reuse.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
volodkevych: ok

server/aws-lsp-codewhisperer/src/language-server/qChatServer.ts
pras0131: yes, I am making similar changes in qAgenticServer as well.

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
pras0131: We discussed in the beginning that session controls might not be required for inline chat. Regarding code re-use, we don't have an actual tabId here in case of inline chat and session management has tabId as the main top level key. To avoid confusion while reading code, decided to keep it separate.

volodkevych:
We need to extend unit tests to cover the new functionality.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -407,6 +472,11 @@ export class AgenticChatController implements ChatHandlers {
         }
     }
 
+    async #getInlineChatTriggerContext(params: InlineChatParams) {
+        let triggerContext: TriggerContext = await this.#triggerContext.getNewTriggerContext(params)
+        return triggerContext
```
volodkevych, line 477: You can just return `this.#triggerContext.getNewTriggerContext(params)` - no need to await here.

pras0131: Let me add this update and tests in a follow up PR.

PR #899, opened 2025-04-01
fix: revert tsconfig changes in codewhisperer-runtimes

francescoopiccoli:
## Problem
quick fix to revert breaking changes in aws-lsp-codewhisperer-runtimes that caused the token standalone to break.
See https://github.com/aws/language-servers/pull/883/files#r2022907628

## License


app/aws-lsp-codewhisperer-runtimes/src/main.ts
```diff
@@ -44,4 +44,4 @@ const worker = new Worker(
 
 // Start Language Client
 const client = new LanguageClient(worker)
-await client.start()
+// await client.start()
```
francescoopiccoli, line 47: await requires es module

PR #901, opened 2025-04-02
feat(amazonq): add streaming client caching and inflight-request cancellation

patrickvbeurden:
## Problem

We need to cache the streaming client and be able to abort inflight requests on profile/connection changes.

## Solution

- initialize the streaming client with a token provider function that invokes the credentials provider before each request
- Add a wrapper around the streaming client that stores the abort controllers of inflight requests, similar to the solution for the regular codewhisperer service
    - This also allows us to add the active profile arn to incoming requests rather than making it the responsibility of the caller

Testing: new unit tests and validated chat still works in bundled minimial vs client.

## License


viktorsaws (approved):

ege0zcan:
We had an old POC for enabling chat cancellation: https://github.com/aws/language-servers/pull/433 

Should we close that open PR if this PR brings cancellation support?

PR #902, opened 2025-04-02
feat: context data selection support in chat-client

volodkevych:
## Problem

We need to support context commands sent from server to chat-client. This is expected to work through 'aws/chat/sendContextCommands' protocol.

## Solution

- Upgrade to the latest "@aws/language-server-runtimes" package
- Show context commands as quick actions in chat client (available via `@`)

Tested with fake data:
Screenshot 2025-04-02 at 14 45 36

## License

### DIFF
--- a/chat-client/src/client/chat.ts
+++ b/chat-client/src/client/chat.ts
@@ -21,7 +21,9 @@ import {
 } from '@aws/chat-client-ui-types'
 import {
     CHAT_REQUEST_METHOD,
+    CONTEXT_COMMAND_NOTIFICATION_METHOD,
     ChatParams,
+    ContextCommandParams,
     FEEDBACK_NOTIFICATION_METHOD,
     FOLLOW_UP_CLICK_NOTIFICATION_METHOD,
     FeedbackParams,
@@ -93,14 +95,16 @@ export const createChat = (
             case ERROR_MESSAGE:
                 mynahApi.showError((message as ErrorMessage).params)
                 break
+            case CONTEXT_COMMAND_NOTIFICATION_METHOD:
+                mynahApi.sendContextCommand(message.params as ContextCommandParams)
+                break
             case CHAT_OPTIONS: {
                 const params = (message as ChatOptionsMessage).params
                 const chatConfig: ChatClientConfig = params?.quickActions?.quickActionsCommandGroups
                     ? {
                           quickActionCommands: params.quickActions.quickActionsCommandGroups,
-                          disclaimerAcknowledged: config?.disclaimerAcknowledged ?? false,
                       }
-                    : { disclaimerAcknowledged: config?.disclaimerAcknowledged ?? false }
+                    : {}
 
                 tabFactory.updateDefaultTabData(chatConfig)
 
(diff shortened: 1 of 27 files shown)


chat-client/src/client/chat.ts
```diff
@@ -93,14 +95,16 @@ export const createChat = (
...
+                break
             case CHAT_OPTIONS: {
                 const params = (message as ChatOptionsMessage).params
                 const chatConfig: ChatClientConfig = params?.quickActions?.quickActionsCommandGroups
                     ? {
                           quickActionCommands: params.quickActions.quickActionsCommandGroups,
-                          disclaimerAcknowledged: config?.disclaimerAcknowledged ?? false,
                       }
-                    : { disclaimerAcknowledged: config?.disclaimerAcknowledged ?? false }
+                    : {}
```
vidkhath, line 107: is this intentional?

volodkevych: Yes, there was a bug due to this: `disclaimerAcknowledged` is not a property on mynah-ui and it throws error now. That's the reason why you do not see quick actions on your 1st chat tab.

chat-client/src/client/mynahUi.ts
```diff
@@ -423,12 +429,23 @@ ${params.message}`,
         }
     }
 
+    const sendContextCommand = (params: ContextCommandParams) => {
```
avi-alpert, line 432: is there a reason we call it `sendContextCommand` instead of `sendContextCommands` I see its plural in CONTEXT_COMMAND_NOTIFICATION_METHOD. I think it should be plural everywhere, since we can and likely always send multiple commands

volodkevych: renamed

vidkhath (approved):

chat-client/src/client/chat.ts
ege0zcan: I understand this is a fix but I want to make sure that this doesn't break any disclaimer logic that is expected, do we know if this might also have some implications?

client/vscode/src/chatActivation.ts
```diff
@@ -56,6 +57,13 @@ export function registerChat(languageClient: LanguageClient, extensionUri: Uri,
         languageClient.info(`[VSCode Client] Received telemetry event from server ${JSON.stringify(e)}`)
     })
 
+    languageClient.onNotification(contextCommandsNotificationType, params => {
```
ege0zcan, line 60: not important, but is it possible to do something generic here for the future:

```typescript
languageClient.onNotification(anyNotification, params => {
        panel.webview.postMessage({
            command: anyNotification.method,
            params: params,
        })
})
```

chat-client/src/client/chat.ts
volodkevych: Setting `disclaimerAcknowledged` property was never accepted in `mynahUi.updateStore`, so I do not think removing it breaks any functionality.

ege0zcan (approved):

client/vscode/src/chatActivation.ts
volodkevych: I think we can - I'll improve it in the next PR.

PR #906, opened 2025-04-03
feat(amazonq): centralize configuration handling to base service manager class

patrickvbeurden:
## Problem

Several calls are being made to `getConfiguration` from different servers at the same time using the same logic in different places. This is not ideal from a performance / redundancy perspective, and might also cause logical issues down the line. We should centralize these configuration updates

## Solution

1. Change `BaseAmazonQServiceManager` to an abstract class, already implementing the `handleDidChangeConfiguration` method
2. Add support for listeners that can hook into the service manager that will be executed at the end of any `handleDidChangeConfiguration` call.
    - This allows us to move hooking to the lsp.didChangeConfiguration notification inside the base manager as well, removing the need for duplication in all servers
3. Add a flag to the base manager class that prevents any additional calls to `lsp.workspace.getConfiguration` when a `handleDidChangeConfiguration` call is already in progress.

Testing: passing existing unit tests, added new tests, verified with bundle vs code client that there is only a single start up `handleDidChangeConfiguration` call that is actually processed (the other calls exit early), and after that only single calls at `didChangeConfiguration` events, as well as listeners being triggered correctly.

## License


server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
```diff
@@ -620,13 +618,18 @@ export const CodewhispererServerFactory =
                 }
             )
 
-            await updateConfiguration()
+            /*
+                NOTE: Only the code whisperer server should trigger the handleDidChangeConfiguration method on initialized.
```
viktorsaws, line 622: This is actually a problem we need to solve as part of this change IMO. Currently we can't rely on only CodeWhispererServer calling `amazonQServiceManager.handleDidChangeConfiguration()`, because it is not guaranteed that it will be always bundled in final Language Server application.

For example, it is possible to choose to bundle any on these separately:
https://github.com/aws/language-servers/blob/main/app/aws-lsp-codewhisperer-runtimes/src/token-standalone.ts#L20-L24. And if that final bundle does not include codeWhispererServer, configuration will nto be handled correctly anymore.

So we either need to keep calling `await amazonQServiceManager.handleDidChangeConfiguration()` in every server dependent on `amazonQServiceManager` or we need to do this refactoring as part of this PR.

I think easies will be to extract `amazonQServiceManager` into a separate Server, and then make it hard dependency for each dependent server. We could log some errors, when bundle does not this amazonQServiceManager Server.

server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.test.ts
```diff
@@ -107,7 +109,7 @@ describe('TelemetryService', () => {
         },
     }
 
-    beforeEach(() => {
+    beforeEach(async () => {
```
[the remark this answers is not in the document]
patrickvbeurden: Thanks, forgot to remove that!

server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
patrickvbeurden: We discussed it some more offline, and decided to implement the alternative in this PR, and keep calling `await amazonQServiceManager.handleDidChangeConfiguration()` at initialize in each server. Will follow up with another PR to create a separate Server

PR #857, opened 2025-03-21
refactor(amazonq): reorg amazonq servers by their function

wzxxing:
- **refactor(amazonq): move everything non-server out of language-server**

## Problem

AmazonQ server is contributed by many teams and there are utilities function scattered all over the `language-server` directory.

> We need to split aws-lsp-codewhisperer project into separate projects within language-servers monorepo.

There will be further discussion about whether separate project is worthy of the efforts. But we agreed that at least splitting servers using the directory is a good start.

## Solution

There will be further discussion about whether separate project is worthy of the efforts. But we agreed that at least splitting servers using the directory is a good start. A new directory `shared` is created to host the code that can be shared among different servers. Otherwise, each server should be restricted in their respective directories.

Tested by launching the codewhisperer server in the VSCode debug profile.

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #907, opened 2025-04-03
Move proxy server to shared directory

wzxxing:
## Problem

This is a follow up task from https://github.com/aws/language-servers/pull/857.

The proxy-server is shared in multiple places and is not tied to a particular LSP server.

## License


server/aws-lsp-codewhisperer/README.md
```diff
@@ -2,8 +2,20 @@
 
 ## Guides
 
+### Components
```
volodkevych, line 5: Can we please move it to CONTRIBUTING.md (new file if not there)? We can name `aws-lsp-codewhisperer` a capability and rename components to servers. Also, explain what has to be in `shared` and `client` subdirs. Also, mention that server cannot reference another server.
I would also explain what a server is and reference `Server` interface in runtimes package.

src/server -> src/language-server ?

wzxxing: Sure, will add a contributing guide line.

You mean 

- `agenticChat`
- `chat`
- `inline-completion`
- `netTransform`
- `securityScan`

Should be called "servers", the whole package is called a capability.

volodkevych (approved):

PR #910, opened 2025-04-03
feat: support create prompt form in chat-client

volodkevych:
## Problem

Custom form needs to be supported in chat-client as a UI to create prompt files.

## Solution

- If context item is selected with `'create-saved-prompt'` id, show the create prompt form
- If Submit button or corresponding "enter" key was pressed, send `createPrompt` request to extension/chat server.
Screenshot 2025-04-03 at 18 16 52

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -39,6 +39,13 @@ export interface InboundChatApi {
 
 type ContextCommandGroups = MynahUIDataModel['contextCommands']
 
+const ContextPrompt = {
+    CreateItemId: 'create-saved-prompt',
+    CancelButtonId: 'cancel-create-prompt',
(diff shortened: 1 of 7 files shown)


chat-client/src/client/mynahUi.ts
```diff
@@ -266,6 +273,44 @@ export const createMynahUi = (
...
+                mynahUi.showCustomForm(
+                    tabId,
+                    [
+                        {
+                            id: ContextPrompt.PromptNameFieldId,
+                            type: 'textinput',
+                            mandatory: true,
+                            autoFocus: true,
+                            title: 'Prompt name',
+                            placeholder: 'Enter prompt name',
```
dogusata, line 287: nit: At some point, we need to move these text to somewhere else for future, for translation.
**Not now.**

dogusata (approved):

PR #912, opened 2025-04-04
feat: port executeBash tool from VSC 

Hweinstock:
## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -0,0 +1,349 @@
...
+    HighRisk,
+    Destructive,
+}
+
+export const dangerousPatterns = new Set(['<(', '$(', '`'])
+export const splitOperators = new Set(['|', '&&', '||', '>'])
+export const splitOperatorsArray = Array.from(splitOperators)
+export const commandCategories = new Map<string, CommandCategory>([
+    // ReadOnly commands
+    ['ls', CommandCategory.ReadOnly],
```
justinmk3, line 20: Just for reference: this is, at best, an extremely crude heuristic. It can't give any sort of security, it can only help users avoid basic mistakes. 

If we want to go further with the "shell parsing", we'll need a robust library like https://www.shellcheck.net/ . And even then, parsing doesn't actually give guarantees about the binaries on the system or in the user's $PATH.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -0,0 +1,349 @@
...
+
+    private static truncateSafelyWithSuffix(str: string, maxLength: number): [string, boolean] {
+        if (str.length > maxLength) {
+            return [str.substring(0, maxLength), true]
+        }
+        return [str, false]
+    }
+
+    private static async whichCommand(logger: Logging, cmd: string): Promise<string> {
+        const cp = new processUtils.ChildProcess(logger, 'which', [cmd], {
```
justinmk3, line 305: The POSIX (i.e most-compatible) way is to use `command -v` (`sh -c 'command -v ...'`) or similar. Not sure if that matters, just FYI.

core/aws-lsp-core/src/util/path.ts
```diff
@@ -91,3 +91,7 @@ export function sanitize(inputPath: string): string {
     }
     return sanitized
 }
+
+export function getUserHomeDir(): string {
+    return os.homedir()
```
justinmk3, line 96: We'll probably need fs.ts even if it doesn't look like what's in VSC, for this kind of "system/environment-specific knowledge", which usually ends up being usefully separated from "pure path" operations.

justinmk3 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
Hweinstock: Yeah, I'm hoping this is temporary because it seems like a huge security vulnerability. I will reach out to the original author to confirm.

core/aws-lsp-core/src/util/path.ts
Hweinstock: This change might be more appropriate [here](https://github.com/aws/language-server-runtimes/blob/674c02696c150838b4bc93543fb0009c5982e7ad/runtimes/server-interface/workspace.ts#L17).

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
Hweinstock: That sounds like a better option. Additionally, I think these tests need to be refactored to avoid mocking the child process. This will allow us to confirm cross-platform support in CI.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolServer.ts
```diff
@@ -12,3 +13,9 @@ export const FsToolsServer: Server = ({ workspace, logging, agent }) => {
 
     return () => {}
 }
+
+export const BashToolsServer: Server = ({ logging, agent }) => {
+    const bashTool = new ExecuteBash(logging)
+    agent.addTool(bashTool.getSpec(), (input: ExecuteBashParams) => bashTool.invoke(input))
```
kmile, line 19: This invoke is likely where we want to do things like check for confirmation, report progress, etc. This really depends on the specific interaction and tool, so the framework does not prescribe any specific ways how to defer, confirm, or report on command execution.

kmile (approved):

justinmk3:
VSC logic was updated: https://github.com/aws/aws-toolkit-vscode/pull/6977

PR #919, opened 2025-04-07
feat: handle fileClick events

vidkhath:
## Problem
Handle file click events in chat client to support file click for context transparency.

## Solution
image

## License

### DIFF
--- a/chat-client/src/client/messager.ts
+++ b/chat-client/src/client/messager.ts
@@ -16,6 +16,7 @@ import {
     ChatParams,
     CreatePromptParams,
     FeedbackParams,
+    FileClickParams,
     FollowUpClickParams,
     InfoLinkClickParams,
     LinkClickParams,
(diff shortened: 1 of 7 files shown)


chat-client/src/client/messager.ts
```diff
@@ -165,4 +168,9 @@ export class Messager {
     onCreatePrompt = (promptName: string): void => {
         this.chatApi.createPrompt({ promptName })
     }
+
+    onFileClick = (params: FileClickParams): void => {
+        this.chatApi.telemetry({ ...params, name: FILE_CLICK_TELEMETRY_EVENT })
```
volodkevych, line 173: Do we also have telemetry for this on vscode side?

volodkevych: If no, we should double-check if it's really needed.

volodkevych (approved):

PR #923, opened 2025-04-08
feat: port readDirectoryRecursively from VSC

Hweinstock:
## Problem
This utility function is needed for the agentic tools and currently lives in the VSC repo. 

## Solution
- Port it to the flare. ([Source](https://github.com/aws/aws-toolkit-vscode/blob/dfee9f7a400e677e91a75e9c20d9515a52a6fad4/packages/core/src/shared/utilities/workspaceUtils.ts#L699))
- Add tests for it (currently missing in VSC). 
- refactor. Only major difference is that it no longer pushes stringified-errors to the results, and allows an option to fail on errors. 

## Notes
This PR contains some workarounds with TODO comments that can be properly addressed by https://github.com/aws/language-server-runtimes/pull/415. 

This same PR will also fix the warning on newer versions of node:
```
(node:40516) [DEP0178] DeprecationWarning: dirent.path is deprecated in favor of dirent.parentPath
(Use `node --trace-deprecation ...` to show where the warning was created
```

## License


core/aws-lsp-core/src/util/workspaceUtils.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        before(async function () {
+            tempFolder = await TestFolder.create()
+            testFeatures = new TestFeatures()
+            // Taken from https://github.com/aws/language-server-runtimes/blob/674c02696c150838b4bc93543fb0009c5982e7ad/runtimes/runtimes/standalone.ts#L216
+            testFeatures.workspace.fs.readdir = path => fs.readdir(path, { withFileTypes: true })
+            testFeatures.workspace.fs.exists = path =>
+                fs.access(path).then(
+                    () => true,
+                    () => false
+                )
```
kmile, line 27: `TestFeatures` already provides stubs for all namespaces, so this should technically not be necessary. Unless you want to 'undo' the stubbing and go back to the real filesystem.

See also: https://github.com/aws/language-server-runtimes/blob/674c02696c150838b4bc93543fb0009c5982e7ad/runtimes/testing/TestFeatures.ts#L53

kmile: Looking at the rest of the test, it's indeed easier just o defer to the actual filesystem than to stub out a bunch of recursive / nested calls to `TestFeatures`. 👍

kmile (approved):

core/aws-lsp-core/src/util/workspaceUtils.test.ts
Hweinstock: Yeah I am intending to 'undo' the stubbing here, but most ways felt pretty awkward. This is the best solution I came up with, but lmk if you see a better option.

PR #924, opened 2025-04-08
fix: pin typescript version and fix compile errors

rahmaniaam:
## Problem
Versions used in lockfiles sometimes does not get updated unless it's deleted then regenerated. The particular problem we have is that the typescript version used would be an older one, but when ported to another platform it would pull in the latest (because it's not using lockfile) causing new typescript compile errors that didn't exist before.

## Solution
Pin typescript to the minimum latest of both platforms, regenerate lockfile, then fix the new errors

## License


francescoopiccoli (approved):
have you considered also the option with `npm update` or specifically `npm update typescript`?

package.json
```diff
@@ -31,7 +31,7 @@
         "package": "npm run compile && npm run package --workspaces --if-present"
     },
     "dependencies": {
-        "typescript": "^5.5.3"
```
natdurlik, line 34: why we don't use ^ range anymore?

rahmaniaam: To make it more consistent between platforms. In this repo it's currently using 5.7.x, but if in another it's 5.8.x there is a potential of new typescript type checks that gets introduced

PR #932, opened 2025-04-08
Initial setup of sending Tool Spec to GenerateAssistantResponse

kmile:
## Problem

This is an example of a tool that can be sent to GenerateAssistantResponse. The Tool output is just being logged, no actual agent loop is happening.

This server can be run by picking the "CodeWhisperer Agentic Server Token" launch configuration. It does not impact the regular chat server.

## License


core/codewhisperer-streaming/amzn-codewhisperer-streaming-1.0.4.tgz
justinmk3, line 1: I guess this is off topic, but why isn't this a build step? Why is a package from this same repo required to be committed as a tgz?

justinmk3 (approved):

core/codewhisperer-streaming/amzn-codewhisperer-streaming-1.0.4.tgz
Hweinstock: I believe this package is generated from the internal smithy models, similar to what we do in VSC for our streaming clients. However, they package the client into a tarball to avoid all the build problems we struggled with. Feel free to correct me if I'm wrong @kmile. (looking at https://github.com/aws/language-servers/pull/743).

justinmk3: oh makes sense. I thought this was a lsp server, didn't read the name carefully enough.

volodkevych (approved):

server/aws-lsp-codewhisperer/src/shared/constants.ts
```diff
@@ -7,6 +7,7 @@ export const DEFAULT_AWS_Q_REGION = 'us-east-1'
 
 export const AWS_Q_ENDPOINTS = new Map([
     [DEFAULT_AWS_Q_REGION, DEFAULT_AWS_Q_ENDPOINT_URL],
+    ['us-east-1', 'https://codewhisperer.us-east-1.amazonaws.com/'],
```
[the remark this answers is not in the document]
kmile: Yes, and no. If you change the default to something other than `us-east-1`, certain parts of the code fail that expect a `us-east-1` record to always be there. So we either change the defaults to always point to an existing item in the list, or what I chose to do, add the 2 regions that are expected to always be there explicitly in the list to not break compilation.

core/codewhisperer-streaming/amzn-codewhisperer-streaming-1.0.4.tgz
kmile: @Hweinstock that's exactly correct!

PR #934, opened 2025-04-09
feat: add proper windows support for executeBash and remove mocks in tests.

Hweinstock:
## Problem
The tests for executeBash currently mock the underlying process providing little confidence the flow works properly, as child processes can behave in unexpected ways on different OS. 

After removing the mocks for executeBash the tests still pass on windows. I suspect this is because Git for Windows includes Git Bash which includes `which` and other unix utilities. However, this may not be the case on customer machines.

## Solution
- remove the mocks, and test the actual flow. 
- add proper windows support by using `where` instead since Node spawned processes [default to command prompt](https://nodejs.org/api/child_process.html#default-windows-shell), which supports `where` by default. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
@@ -220,6 +220,7 @@ export class ExecuteBash {
 
             let firstChunk = true
             let firstStderrChunk = true
+            const writer = updates?.getWriter()
             const childProcessOptions: processUtils.ChildProcessOptions = {
                 spawnOptions: {
                     cwd: cwd,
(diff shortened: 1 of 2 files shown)


justinmk3 (approved):

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -220,6 +220,7 @@ export class ExecuteBash {
 
             let firstChunk = true
             let firstStderrChunk = true
+            const writer = updates?.getWriter()
```
kmile, line 223: Where is this writer used? If it's used to communicate updates to the client then we might want to initialize the tool with a more comprehensive abstraction (I'm not saying tossing `Features` into it, but maybe something more explicit than an `any` stream?

PR #937, opened 2025-04-09
test: expand fsWrite test suite and sync with VSC

Hweinstock:
## Problem
There have been significant changes to the `fsWrite` tool since it was originally ported in https://github.com/aws/language-servers/pull/894. 

Additionally, the test suite is missing coverage for some core components and it has the properly that tests are dependent on each other. 

For example, one test creates a file that the next test uses. This is problematic because the tests become coupled and changes to one test could trickle down and cause unclear side effects. 

## Solution
- Port fsWrite latest changes from VSC. (as of https://github.com/aws/aws-toolkit-vscode/blob/8e00eefa33f4eee99eed162582c32c270e9e798e/packages/core/src/codewhispererChat/tools/fsWrite.ts#L42). 
- refactor tests so that they are isolated. 
- add coverage for the new diff mechanism. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
@@ -49,6 +49,32 @@ export class FsWrite {
         this.workspace = features.workspace
     }
 
+    public async validate(params: FsWriteParams): Promise<void> {
+        switch (params.command) {
+            case 'create':
+                if (!params.path) {
+                    throw new Error('Path must not be empty')
+                }
+                break
+            case 'strReplace':
+            case 'insert': {
+                const fileExists = await this.workspace.fs.exists(params.path)
+                if (!fileExists) {
+                    throw new Error('The provided path must exist in order to replace or insert contents into it')
+                }
+                break
+            }
+            case 'append':
+                if (!params.path) {
+                    throw new Error('Path must not be empty')
+                }
+                if (!params.newStr) {
+                    throw new Error('Content to append must not be empty')
+                }
+                break
+        }
+    }
+
     public async invoke(params: FsWriteParams): Promise<InvokeOutput> {
         const sanitizedPath = sanitize(params.path)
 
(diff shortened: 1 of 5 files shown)


justinmk3 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
```diff
@@ -1,8 +1,9 @@
 import { InvokeOutput } from './toolShared'
 import { Features } from '@aws/language-server-runtimes/server-interface/server'
 import { sanitize } from '@aws/lsp-core/out/util/path'
+import { Change, diffLines } from 'diff'
 
-// Port of https://github.com/aws/aws-toolkit-vscode/blob/10bb1c7dc45f128df14d749d95905c0e9b808096/packages/core/src/codewhispererChat/tools/fsWrite.ts#L42
+// Port of https://github.com/aws/aws-toolkit-vscode/blob/8e00eefa33f4eee99eed162582c32c270e9e798e/packages/core/src/codewhispererChat/tools/fsWrite.ts#L42
```
kmile, line 6: Thanks for the attention to detail! :D

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
```diff
@@ -68,6 +101,91 @@ export class FsWrite {
         }
     }
 
+    public async queueDescription(updates: WritableStream): Promise<void> {
+        const updateWriter = updates.getWriter()
+        // Write an empty string because FsWrite should only show a chat message with header
```
kmile, line 106: This is something we might want/need to reconsider, I am assuming those map to the `TODO`s in the `toolServer` to add responses back to the user?

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
Hweinstock: Yes, this is how its currently done on the VSC side, but unsure if this exact setup makes sense in the LSP. 
relevant: https://github.com/aws/language-servers/pull/934/files/8476a7e6f87972a329a47c58d29ebd4c87177e1a#r2039648303

justinmk3 (approved):

PR #939, opened 2025-04-09
feat: add proper windows support for executeBash (WIP)

Hweinstock:
Built on https://github.com/aws/language-servers/pull/934

## Problem
After removing the mocks for executeBash the tests still pass on windows. I suspect this is because Git for Windows includes Git Bash which includes `which` and other unix utilities. However, this may not be the case on customer machines.

## Solution
- add proper windows support by using `where` instead. 

## License


Hweinstock:
merged into https://github.com/aws/language-servers/pull/934

PR #945, opened 2025-04-10
feat: add iam support in q chat server and q agentic server

francescoopiccoli:
## Problem
In this PR we add IAM auth support to q chat server and q agentic server. It required some refactoring of the chat servers and the StreamingClientService, as previously chat servers were coupled with token auth only. To do this, for consistency and clarity I followed the same patterns as we do for inline-completion servers (CodewhispererServerFactory) and for inline-completion services (CodeWhispererServiceBase).

To manually test follow the updated instructions in readme and contributing.md, this pr does not include any new tests, some basic tests have been added in https://github.com/aws/language-servers/pull/953. 

The PR was initially meant to be reviewed commit by commit, after requested changes this no longer holds. I would still suggest to review based on which files were modified in each initial commit.

## License


server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/BaseAmazonQServiceManager.ts
```diff
@@ -13,24 +15,26 @@ import {
...
-
+import { StreamingClientServiceBase } from '../streamingClientService'
 export interface Features {
     lsp: Lsp
     logging: Logging
     runtime: Runtime
     credentialsProvider: CredentialsProvider
     sdkInitializator: SDKInitializator
     workspace: Workspace
+    chat?: Chat
```
patrickvbeurden, line 26: Are these new ones necessary for the service manager? Maybe I missed something, but I don't think they are used

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQIAMServiceManager.ts
```diff
@@ -33,6 +43,30 @@ export class AmazonQIAMServiceManager extends BaseAmazonQServiceManager<CodeWhis
 
         return this.cachedCodewhispererService
     }
+
+    public getStreamingClient() {
+        const iamService = this.getCodewhispererService()
+
+        if (!iamService || !this.region || !this.endpoint) {
```
patrickvbeurden, line 50: I think you could leave these checks out, as well as the factory. They are present in the token service manager because the user could change profile/region during the session so there `getCodewhispererService` method does some connection / state change checks, but with IAM this is not the case

nit: the new region and endpoint properties can be reused in `getCodewhispererService` as well

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -126,7 +128,12 @@ export class AgenticChatController implements ChatHandlers {
         const conversationIdentifier = session?.conversationId ?? 'New conversation'
         try {
             this.#log('Request for conversation id:', conversationIdentifier)
-            const profileArn = AmazonQTokenServiceManager.getInstance(this.#features).getActiveProfileArn()
+
+            let profileArn: string | undefined
+
+            if (this.#amazonQServiceManager instanceof AmazonQTokenServiceManager) {
```
patrickvbeurden, line 134: nit: the token streaming client adds the profileArn to the request if it is isn't undefined, so I believe this could be left out entirely

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
```diff
@@ -30,7 +30,7 @@ import {
     ChatTelemetryEventName,
     CombinedConversationEvent,
 } from '../../shared/telemetry/types'
-import { Features, LspHandlers, Result } from '../types'
```
patrickvbeurden, line 33: nit: I would keep using `Features` from `../types` here (and in other places where it was being used), perhaps it would be good to consider changing the name of `Features` in the service manager eventually since it is a bit confusing

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.test.ts
```diff
@@ -29,7 +29,7 @@ describe('Chat Session Service', () => {
     }
 
     beforeEach(() => {
-        codeWhispererStreamingClient = stubInterface<StreamingClientService>()
+        codeWhispererStreamingClient = stubInterface<StreamingClientServiceToken>()
```
patrickvbeurden, line 32: nit: would consider adding coverage for the IAM client here as well, since e.g. abort being called correctly is specific to the different implementations

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/BaseAmazonQServiceManager.ts
francescoopiccoli: Good point, they are not, I added to have compatibility and to access chat from 
`export const QChatServerFactory =
    (serviceManager: (features: Features) => AmazonQBaseServiceManager): Server =>`
    since Features type that I am using is the one coming from `BaseAmazonQServiceManager`, I will change that to be the Features type from ../types (i.e. from the runtimes definition).

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQIAMServiceManager.ts
francescoopiccoli: thanks, yea i added them for consistency but they are indeed not needed, will remove

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
francescoopiccoli: not sure i got this, my idea was to make sure it was not added for `AmazonQIAMServiceManager`. From my understanding `profileArn` is token only. Is this correct? (and what about `customizationArn`)?

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.ts
francescoopiccoli: agreed, its confusing to have the same name for the two types, will keep the import from `../types` and rename the `Features` type defined in `BaseAmazonQServiceManager` to `QServiceManagerFeatures`

patrickvbeurden (approved):

server/aws-lsp-codewhisperer/src/shared/proxy-server.ts
```diff
@@ -26,8 +26,10 @@ export const QNetTransformServerTokenProxy = QNetTransformServerToken(
     }
 )
 
-export const QChatServerProxy = QChatServer()
+export const QChatServerTokenProxy = QChatServerFactory(initBaseTokenServiceManager)
+export const QChatServerIAMProxy = QChatServerFactory(initBaseIAMServiceManager)
```
viktorsaws, line 30: We should stop using `Proxy` suffix for servers, it is meaningless and will be deprected

app/aws-lsp-codewhisperer-runtimes/src/token-standalone.ts
```diff
@@ -21,7 +21,7 @@ const props: RuntimeProps = {
         CodeWhispererSecurityScanServerTokenProxy,
         QConfigurationServerTokenProxy,
         QNetTransformServerTokenProxy,
-        QChatServerProxy,
+        QChatServerTokenProxy,
```
viktorsaws, line 24: Don't forget to change all consuming code after this rename. We have some config in internal RBC package

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -551,6 +557,6 @@ export class AmazonQTokenServiceManager extends BaseAmazonQServiceManager<CodeWh
     }
 }
 
-export const initBaseTokenServiceManager = (features: Features): AmazonQBaseServiceManager => {
+export const initBaseTokenServiceManager = (features: QServiceManagerFeatures): AmazonQBaseServiceManager => {
```
viktorsaws, line 560: nit: why do we need to change the type from `Features`? I think it's cleaner to keep using previous one

server/aws-lsp-codewhisperer/src/shared/streamingClientService.ts
```diff
@@ -75,11 +113,41 @@ export class StreamingClientService {
...
+            endpoint: endpoint,
+            credentialProvider: new CredentialProviderChain([
+                () => credentialsProvider.getCredentials('iam') as Credentials,
+            ]),
+            retryStrategy: new ConfiguredRetryStrategy(0, (attempt: number) => 500 + attempt ** 10),
         })
-        this.inflightRequests.clear()
+    }
+
+    public async sendMessage(
```
viktorsaws, line 137: Looks like common sendMessage logic could also be consolidated in base class, it's only 1 parameter that's different, which could be passed to base method.

viktorsaws (approved):

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
francescoopiccoli: With @patrickvbeurden we agreed (see [comment](https://github.com/aws/language-servers/pull/945#discussion_r2038995908)) that it was confusing having two types having the same name but different properties (i.e. the `Features` defined in runtime and `Features` type defined in `BaseAmazonQServiceManager`), one being a subset of the other. Imo QServiceManagerFeatures gives a better idea of the fact that it is a subset of the general `Features` type.

server/aws-lsp-codewhisperer/src/shared/streamingClientService.ts
francescoopiccoli: Initially I agreed, but i tried modifying but it ends up making the code harder to read, we need to pass a child class specific param (profileArn) as optional to base class, and the call this.client.sendMessage has to be done by if / else on the streaming client to make sure params of the call are compatible. I will keep current implementation

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
```diff
@@ -61,9 +64,15 @@ export class ChatSessionService {
 
         const client = this.#amazonQServiceManager.getStreamingClient()
 
-        const response = await client.generateAssistantResponse(request, this.#abortController)
-
-        return response
+        if (client instanceof StreamingClientServiceToken) {
```
viktorsaws, line 67: Can we avoid relying of Class instance here? You can check availability of `generateAssistantResponse` method on the StreamingClient class instead, and keep `ChatSessionService` independent from type of streaming service used.

server/aws-lsp-codewhisperer/src/shared/proxy-server.ts
viktorsaws: We will need to do both: remove Proxy suffix from all these functions, and remove `proxy-server` file altogether. I'm ok with keeping current change for consistency.

server/aws-lsp-codewhisperer/src/language-server/chat/qChatServerIAM.test.ts
```diff
@@ -0,0 +1,113 @@
+import { CancellationToken, Server } from '@aws/language-server-runtimes/server-interface'
+import { TestFeatures } from '@aws/language-server-runtimes/testing'
+import sinon from 'ts-sinon'
+import { ChatController } from './chatController'
+import { ChatSessionManagementService } from './chatSessionManagementService'
+import { QChatServerFactory } from './qChatServer'
+import { AmazonQIAMServiceManager } from '../../shared/amazonQServiceManager/AmazonQIAMServiceManager'
+
+describe('QChatServerIAM', () => {
```
viktorsaws, line 9: Cna we optimize tests setup and avoid having 2 identical test suite files? Given QChatServer does not have any specific implementation based on type of service used, we should be able to run same test suite for both service manager variants.

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
francescoopiccoli: checking availability of `generateAssistantResponse` method still requires casting with specific streaming service (`StreamingClientServiceToken`), as  `client` is an instance of the base class which does not have `generateAssistantResponse` method
```
if ('generateAssistantResponse' in client) {
            const tokenClient = client as StreamingClientServiceToken
            const response = await tokenClient.generateAssistantResponse(request, this.#abortController)
```

without casting it throws a typescript compile error: 'client.generateAssistantResponse' is of type 'unknown'.

viktorsaws (approved):

viktorsaws (approved):

PR #948, opened 2025-04-10
feat(amazonq): support context commands in agentic chat

zixlin7:
## Problem
We need to support files/folders/code symbols/prompts as context command in agentic chat, and also show user the context referenced in chat response
## Solution
Screenshot 2025-04-17 at 9 46 32 AM

note telemetry update will be followup in next PR

related VS Code change: https://github.com/aws/aws-toolkit-vscode/pull/6999

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -3,6 +3,7 @@
  * Will be deleted or merged.
  */
 
+import * as path from 'path'
 import {
     ChatResponseStream,
     CodeWhispererStreaming,
@@ -36,6 +37,8 @@ import { DEFAULT_HELP_FOLLOW_UP_PROMPT, HELP_MESSAGE } from '../chat/constants'
 import { TelemetryService } from '../../shared/telemetry/telemetryService'
 import { AmazonQTokenServiceManager } from '../../shared/amazonQServiceManager/AmazonQTokenServiceManager'
 import { TabBarController } from './tabBarController'
+import { getUserPromptsDirectory } from './context/contextUtils'
+import { AdditionalContextProvider } from './context/addtionalContextProvider'
 
 describe('AgenticChatController', () => {
     const mockTabId = 'tab-1'
@@ -97,11 +100,13 @@ describe('AgenticChatController', () => {
 
     let sendMessageStub: sinon.SinonStub
     let generateAssistantResponseStub: sinon.SinonStub
+    let additionalContextProviderStub: sinon.SinonStub
     let disposeStub: sinon.SinonStub
     let activeTabSpy: {
         get: sinon.SinonSpy<[], string | undefined>
         set: sinon.SinonSpy<[string | undefined], void>
     }
+    let fsWriteFileStub: sinon.SinonStub
     let removeConversationSpy: sinon.SinonSpy
     let emitConversationMetricStub: sinon.SinonStub
 
@@ -144,13 +149,14 @@ describe('AgenticChatController', () => {
             })
 
         testFeatures = new TestFeatures()
+        fsWriteFileStub = sinon.stub()
 
         testFeatures.workspace.fs = {
             ...testFeatures.workspace.fs,
             getServerDataDirPath: sinon.stub().returns('/mock/server/data/path'),
             mkdir: sinon.stub().resolves(),
             readFile: sinon.stub().resolves(),
-            writeFile: sinon.stub().resolves(),
+            writeFile: fsWriteFileStub.resolves(),
             rm: sinon.stub().resolves(),
         }
 
@@ -161,6 +167,8 @@ describe('AgenticChatController', () => {
             addTool: sinon.stub().resolves(),
         }
 
+        additionalContextProviderStub = sinon.stub(AdditionalContextProvider.prototype, 'getAdditionalContext')
+        additionalContextProviderStub.resolves([])
         // @ts-ignore
         const cachedInitializeParams: InitializeParams = {
             initializationOptions: {
@@ -173,6 +181,7 @@ describe('AgenticChatController', () => {
                 },
             },
         }
+        testFeatures.lsp.window.showDocument = sinon.stub()
         testFeatures.lsp.getClientInitializeParams.returns(cachedInitializeParams)
         setCredentials('builderId')
 
@@ -998,6 +1007,25 @@ describe('AgenticChatController', () => {
                     }
                 )
             })
+        })
+    })
+
+    describe('onCreatePrompt', () => {
+        it('should create prompt file with given name', async () => {
+            const promptName = 'testPrompt'
+            const expectedPath = path.join(getUserPromptsDirectory(), 'testPrompt.prompt.md')
+
+            await chatController.onCreatePrompt({ promptName })
+
+            sinon.assert.calledOnceWithExactly(fsWriteFileStub, expectedPath, '', { mode: 0o600 })
+        })
+
+        it('should create default prompt file when no name provided', async () => {
+            const expectedPath = path.join(getUserPromptsDirectory(), 'default.prompt.md')
+
+            await chatController.onCreatePrompt({ promptName: '' })
+
+            sinon.assert.calledOnceWithExactly(fsWriteFileStub, expectedPath, '', { mode: 0o600 })
         })
     })
 
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
@@ -1,0 +1,154 @@
+import { FileDetails, QuickActionCommand, FileList } from '@aws/language-server-runtimes/protocol'
+import { AdditionalContextPrompt, ContextCommandItem, ContextCommandItemType } from 'local-indexing'
+import * as path from 'path'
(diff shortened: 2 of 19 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
```diff
@@ -0,0 +1,147 @@
...
+import { URI } from 'vscode-uri'
+import { getUserHomeDir } from '@aws/lsp-core/out/util/path'
+
+export type AdditionalContextLengths = {
+    fileContextLength: number
+    promptContextLength: number
+    ruleContextLength: number
+}
+
+export class ContextCommandsProvider {
```
justinmk3, line 15: does this codebase need a `Disposable` interface?
```suggestion
export class ContextCommandsProvider implements Disposable {
```

zixlin7: I guess we can use the one from `vscode-jsonrpc`, will do

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
```diff
@@ -173,6 +181,7 @@ describe('AgenticChatController', () => {
                 },
             },
         }
+        testFeatures.lsp.window.showDocument = sinon.stub()
```
kmile, line 184: Is this stub missing in `TestFeatures`?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+                    rulesFiles.push(path.join(rulesPath, entry.name))
+                }
+            }
+        }
+        return rulesFiles
+    }
+
+    getContextType(prompt: AdditionalContextPrompt): string {
+        if (prompt.filePath.endsWith(promptFileExtension)) {
+            if (pathUtils.isInDirectory(path.join('.amazonq', 'rules'), prompt.relativePath)) {
```
kmile, line 45: Why does `promptsDirectory` get its own handler ahd `rulesDirectory` does not?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+                return 'rule'
+            } else if (pathUtils.isInDirectory(getUserPromptsDirectory(), prompt.filePath)) {
```
kmile, line 47: Why do we use `relativePath` for rules but `filePath` for prompts? Is there a meaningful distinction?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+        const workspaceRules = await this.collectWorkspaceRules(triggerContext)
+        let workspaceFolderPath = triggerContext.workspaceFolder?.uri
+            ? URI.parse(triggerContext.workspaceFolder.uri).fsPath
+            : LocalProjectContextController.getInstance().getRootDirectory()
+
+        if (workspaceRules.length > 0) {
+            additionalContextCommands.push(
+                ...workspaceRules.map(file => ({
+                    workspaceFolder: workspaceFolderPath,
+                    type: 'file' as ContextCommandItemType,
```
kmile, line 68: This hides the type safety of the `type` type. It might be safer to set the return type on the `map` of the upper array, so the compiler actually complains when the `file` is not actually a `ContextCommandItemType`.

Right now, this would compile:

```
type: 'definitelyNotAnItemType' as ContextCommandItemType
```

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+            } else if (pathUtils.isInDirectory(getUserPromptsDirectory(), prompt.filePath)) {
+                return 'prompt'
+            }
+        }
+        return 'file'
+    }
+
+    async getAdditionalContext(
+        triggerContext: TriggerContext,
+        context?: any[]
```
kmile, line 56: Does this have to be `any[]`? Is there no more specific type we can use here?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+        const fileList: FileList = {
+            filePaths: context.map(item => item.relativePath),
+            details: fileDetails,
+        }
+        return fileList
+    }
+
+    mapToContextCommandItems(context: any[], workspaceFolderPath: string): ContextCommandItem[] {
```
kmile, line 132: I don't really get this method, instead of using `any` and converting it using this method, can't we use `ContextCommandItem` directly where we call this method?

The `context: any` always seems to have this type:

```
{ icon: 'file' | 'folder' | 'code-block' | string, command: string | undefined, id: string | undefined }
```
which combined with `workspaceFolderPath` is essentially the same as `ContextCommandItem` already.

This method maps the `any` to the specific type and filters out the known `icon`s?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
```diff
@@ -0,0 +1,134 @@
+import { TriggerType } from '@aws/chat-client-ui-types'
```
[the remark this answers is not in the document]
kmile: Would be good to add this to the header of the file as well since this PR comment will get lost once merged. That way we can track the files' origins once we merge and clean up the `chat` and `agenticChat` folders.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
```diff
@@ -0,0 +1,171 @@
+import * as path from 'path'
+import { FSWatcher, watch } from 'chokidar'
```
kmile, line 2: We're using `workspace.fs` to access the filesystem, but in this case I think `chokidar` uses the underlying `fs` directly.

Could this use https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles instead?

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -20,15 +24,24 @@ export class LocalProjectContextController {
...
+        workspace: Workspace
+    ) {
         this.fileExtensions = Object.keys(languageByExtension)
         this.workspaceFolders = workspaceFolders
         this.clientName = clientName
         this.log = logging
+        this.contextCommandsProvider = new ContextCommandsProvider(logging, chat, workspace)
     }
 
     public static getInstance() {
```
kmile, line 47: Do we use this anywhere? It seems like we're just instantiating the class, and calling it in the `Server` (which is great, I much prefer that over singletons)

kmile:
What is the actual mechanism for sending context items to the client? I don't think I saw it in `LocalProjectContextServer`.

kmile (approved):

zixlin7:
> What is the actual mechanism for sending context items to the client? I don't think I saw it in `LocalProjectContextServer`.

it is send in ContextCommandsProvider here: https://github.com/aws/language-servers/blob/98aa254f0541b7d508106b9c063370c065278475/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts#L71

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
zixlin7: Ah I forgot to dispose it, fixed

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
zixlin7: I think the main reason we have to do this is because context is missing `type` which is used in the call to get prompt from the local context server, but makes sense to create a type for it instead of using any, will follow up

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
zixlin7: because prompts are dynamically updated in the chat context menu, rules are just sent along with api requests

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
zixlin7: yes I think so

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
zixlin7: because rules are in the workspace but prompts are in the `.aws` folder

jpinkney-aws (approved):

jpinkney-aws (approved):

PR #949, opened 2025-04-11
feat: initial support for local project context

breedloj:
## Problem
Local project context is not currently available for Q capabilities that are able to use it to enrich processing.

## Solution
Introduces a LSP component for local project context which can be consumed by Q components such as chat and inline. The LSP server handles file indexing and update events from LSP clients and proxies to a controller abstracting calls to the underlying context library. Components can directly retrieve context without looping back to clients. The library itself will be included as a runtime dependency with relevant Flare bundles. Types from the context library are bundled to aid in development of the Flare component.

Consumption example for chat:

```
    private async extractProjectContext(prompt?: string): Promise<RelevantTextDocument[] | undefined> {
        if (prompt === undefined) {
            return undefined
        }

        const vectorResult = await LocalProjectContextController.getInstance().queryVectorIndex({
            query: prompt
        })

        return chunksToRelevantTextDocuments(vectorResult.chunks)
    }
```

Outside of passing additional configuration (WIP), client destinations will not need to make additional changes to introduce this support to supported capabilities. Context support is entirely optional and is designed to fail gracefully when unavailable.

This feature is by default **off** and will not affect any existing server functionality. Enablement can be toggled via the LSP configuration:

```
aws.q {
      ...
        projectContext: {
            enableIndexing: true | false,
        }
}
```

Subsequent PRs will wire context up to various Q capabilities.

## License


server/aws-lsp-codewhisperer/types/types-local-indexing-1.0.0.tgz
volodkevych, line 1: What would be the process to update those?

volodkevych:
Will this solution work in broswer?

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextController.ts
```diff
@@ -0,0 +1,203 @@
...
+                }
+            }
+        }
+        this.log.info(`Found ${workspaceSourceFiles.length} source files`)
+        return workspaceSourceFiles
+    }
+
+    private async getCodeSourceFiles(dir: string): Promise<string[]> {
+        try {
+            const files = await fs.readdir(dir, { withFileTypes: true })
```
volodkevych, line 154: This will not work in browser. We have `workspace.fs` feature, but it's currently mainly noop for browser. We need to implement reading ws folders for browser.

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
```diff
@@ -0,0 +1,129 @@
...
+            const newPaths = event.files.map(file => file.oldUri.replace('file:', ''))
+
+            await localProjectContextController.updateIndex(oldPaths, 'remove')
+            await localProjectContextController.updateIndex(newPaths, 'add')
+        } catch (error) {
+            logging.error(`Error handling rename event: ${error}`)
+        }
+    })
+
+    lsp.onDidSaveTextDocument(async event => {
```
volodkevych, line 119: How will this server be used? I only see the LPS methods that trigger the update of the index.

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextController.ts
```diff
@@ -0,0 +1,203 @@
...
+        try {
+            const resp = await this._vecLib?.queryInlineProjectContext(request.query, request.filePath, request.target)
+            return resp ?? []
+        } catch (error) {
+            this.log.error(`Error in queryInlineProjectContext: ${error}`)
+            return []
+        }
+    }
+
+    public async queryVectorIndex(request: QueryRequest): Promise<Chunk[]> {
```
volodkevych, line 119: Interesting to know: how is it planned to work for [context](https://github.com/aws/aws-toolkit-vscode/blob/f1dded7edcf7d4d95410f31cfb87da208e91fb27/packages/core/src/codewhispererChat/controllers/chat/controller.ts#L564) given `getContextCommandItems` and `getContextCommandPrompt` are not exposed

breedloj: The currently exposed APIs represent what was needed for existing inline and chat use cases that we were explicitly targeting. However the others can easily be added on a case by case basis.

server/aws-lsp-codewhisperer/types/types-local-indexing-1.0.0.tgz
breedloj: We are adding this step to the SOP for updating the vector lib for publishing to Flare. Longer term the goal is to migrate the entire library to an NPM dependency to avoid any disconnect. The base types in the library rarely change.

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextController.ts
breedloj: It is not currently intended to work in the browser. This would require significant changes to the underlying library to work within the context of a browser file system and web assemblies. Not in scope for this change - is there anything additionally that needs to be done to exclude this from browser distributions?

volodkevych: Yes, excluding it from bundle and handling as optional dependency in other servers should be enough. Some webpack polyfills may be needed in app runtime packages.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/configurationUtils.ts
```diff
@@ -67,10 +67,15 @@ interface QInlineSuggestionsConfig {
     extraContext: string | undefined // aws.q.inlineSuggestions.extraContext
 }
 
+interface QProjectContextConfig {
+    enableIndexing: boolean // aws.q.projectContext.enableIndexing
```
volodkevych, line 71: nit: should this flag or section be named specific to client-side project context? When we add server-side project context - that may need settings too.

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextController.ts
```diff
@@ -0,0 +1,204 @@
...
+                existing => !removed.some(removal => this.areWorkspaceFoldersEqual(existing, removal))
+            )
+
+            const merged = [...afterRemovals]
+            for (const addition of added) {
+                if (!merged.some(existing => this.areWorkspaceFoldersEqual(existing, addition))) {
+                    merged.push(addition)
+                }
+            }
+            if (this._vecLib) {
```
volodkevych, line 96: Shouldn't we check for this on the top of this method - I do not think any code needs to be executed if library is disabled.

volodkevych: In general, you could introduce NoOp controller for cases when indexing is not supported - it would be more readable and less error-prone. Or even check for the `enabled` flag earlier - in server.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/configurationUtils.ts
breedloj: Good call - qualified it: `enableLocalIndexing`

volodkevych (approved):

PR #952, opened 2025-04-12
fix(amazonq): add regionalization support to .NET Transform server

awschristou:
This change updates the .NET Transform server to use `AmazonQTokenServiceManager` when producing the service client. This way, the service calls are able to provide the developer profile arn or region details for IdC based connections.

This change is based on #859, which made a similar change for security scanning.

I worked on this change with @shruti0085 , and we tested it by using the Visual Studio Toolkit to perform a .NET Transform using a local build of the server. We were able to see the `profileArn` field being passed along with all of the .NET Transform related service calls.

## License


server/aws-lsp-codewhisperer/src/language-server/netTransform/transformHandler.ts
```diff
@@ -350,10 +356,10 @@ export class TransformHandler {
         } as GetTransformResponse
     }
 
-    async downloadExportResultArchive(cwStreamingClient: CodeWhispererStreaming, exportId: string, saveToDir: string) {
+    async downloadExportResultArchive(cwStreamingClient: StreamingClientService, exportId: string, saveToDir: string) {
```
viktorsaws, line 359: It's better to use `amazonQServiceManager.getStreamingClient()` instead of passing `StreamingClient` object from caller `netTransformServer`, since `amazonQServiceManager` provides access already and makes sure streaming client properly configured.

viktorsaws (approved):

pranav-firake (approved):

PR #960, opened 2025-04-14
feat: expose configuration for GPU acceleration and index worker threads in context server

breedloj:
## Problem
Clients may want to configure the optional GPU acceleration and index worker thread parameters of the local indexing library. VSCode and Jetbrains support these options in their user facing preference panes.

## Solution
Add these options to the LSP server's supported set of configurables. By default, GPU acceleration is off and index worker threads are not explicitly set.

## License


server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
```diff
@@ -119,7 +119,10 @@ export const LocalProjectContextServer = (): Server => features => {
         try {
             logging.log(`Setting project context enabled to ${updatedConfig.projectContext?.enableLocalIndexing}`)
             updatedConfig.projectContext?.enableLocalIndexing
-                ? await localProjectContextController.init()
+                ? await localProjectContextController.init({
+                      enableGpuAcceleration: updatedConfig?.projectContext?.enableGpuAcceleration,
+                      indexWorkerThreads: updatedConfig?.projectContext?.indexWorkerThreads,
+                  })
                 : await localProjectContextController.dispose()
```
kmile, line 126: At this point should this just be an `if` statement? ;)

kmile, line 126: Can `init` and `dispose` be done multiple times in a row? What happens if (for whatever reason) 2 `init`s or 2 `dispose`s happen in a row?

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/configurationUtils.ts
```diff
@@ -112,6 +114,8 @@ export async function getAmazonQRelatedWorkspaceConfigs(
                 },
                 projectContext: {
                     enableLocalIndexing: newQConfig.projectContext?.enableLocalIndexing === true,
+                    enableGpuAcceleration: newQConfig.projectContext?.enableGpuAcceleration === true,
+                    indexWorkerThreads: newQConfig.projectContext?.indexWorkerThreads ?? -1,
```
kmile, line 118: Does `-1` have a special meaning? Or would `0` threads mean the same thing?

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
kmile: Reading along, it seems like 2x `dispose` is pretty safe, but 2x `init` would result in orphaned resources.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/configurationUtils.ts
breedloj: It would be the same - I just considered `-1` to be more of an effective no-op/disabled state.

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
breedloj: I will address this in a subsequent PR.

breedloj: Made init no-op if already initialized.

kmile (approved):

PR #966, opened 2025-04-15
chore: sync FsWrite tool with vsc version

Hweinstock:
## Problem

## Solution
Sync up to most recent changes: https://github.com/aws/aws-toolkit-vscode/blob/16aa8768834f41ae512522473a6a962bb96abe51/packages/core/src/codewhispererChat/tools/fsWrite.ts#L42

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
@@ -3,10 +3,11 @@ import { Features } from '@aws/language-server-runtimes/server-interface/server'
 import { sanitize } from '@aws/lsp-core/out/util/path'
 import { Change, diffLines } from 'diff'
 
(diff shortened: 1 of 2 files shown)


jpinkney-aws (approved):

Hweinstock:
I've re-synced this PR with the latest changes.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
```diff
@@ -41,33 +41,49 @@ export interface FsWriteBackup {
 }
 
 export class FsWrite {
-    private readonly logging: Features['logging']
     private readonly workspace: Features['workspace']
 
-    constructor(features: Pick<Features, 'workspace' | 'logging'> & Partial<Features>) {
-        this.logging = features.logging
+    constructor(features: Pick<Features, 'workspace'> & Partial<Features>) {
```
kmile, line 46: I love this approach so much, maybe we should introduce

```
type SomeRequiredKeys<T, K extends keyof T> = Pick<T, K> & Partial<Omit<T, K>>
type SomeFeatures<K extends keyof Features> = SomeRequiredKeys<Features, K>
```

to avoid having to re-create this pattern all the time. It's essentially our dependency injection signature.

kmile (approved):

PR #967, opened 2025-04-15
feat: add configurable file indexing logic

taldekar:
## Problem
When the LSP is initialized, local project context needs to crawl the workspaces, build a repo map, and create an index with it that will be used to provide additional context to Q. The workspace need to be crawled and indexed and what files to include in the index needs to be configurable.

## Solution
File indexing options are exposed to the user through the `workspace/configuration` call. This information is used to determine what files to include in the index by the `processWorkspaceFolders` method in `localProjectContextController`.

## Justification for newly introduced libraries
- [**fdir**](https://www.npmjs.com/package/fdir): small (< 2KB), well-tested library that significantly [speeds up](https://github.com/thecodrr/fdir/blob/HEAD/BENCHMARKS.md#:~:text=11.3.244.8%2Dnode.25-,Asynchronous,-4865%20files%20%26%20745) directory crawling (claims to be capable of crawling 1 million files in < 1 second) when building the initial file directory.
- [**picomatch**](https://www.npmjs.com/package/picomatch): [required by fdir](https://github.com/thecodrr/fdir/blob/HEAD/documentation.md#:~:text=Uses%20picomatch%20underneath.%20To%20keep%20fdir%20dependency%20free%2C%20it%20is%20up%20to%20the%20user%20to%20install%20picomatch%20manually.) under the hood to apply a glob filter. This is used to limit the results of the crawl to file extensions supported by the language server.
- [**ignore**](https://www.npmjs.com/package/ignore): well tested, small (~63 kB) package that allows us to enforce file exclusion rules in the `.gitignore` convention. These rules are configurable by different IDEs and are passed in through the `workspace/configuration` call. It is also used to parse the users `.gitignore` files for further filtering.


server/aws-lsp-codewhisperer/package.json
```diff
@@ -46,7 +46,10 @@
         "lokijs": "^1.5.12",
         "uuid": "^11.0.5",
         "diff": "^7.0.0",
-        "vscode-uri": "^3.1.0"
+        "vscode-uri": "^3.1.0",
+        "fdir": "^6.4.3",
+        "ignore": "^7.0.3",
+        "picomatch": "^4.0.2"
```
justinmk3, line 52: why are these new deps needed?  New deps always should include explanations.

is "ignore" for gitignore? we used a different lib in VSC, after several revisions: https://github.com/aws/aws-toolkit-vscode/blob/dc43ad2fb564c70da9a4fdbcbe1e5ce600b7f21b/packages/core/package.json#L528

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextController.ts
```diff
@@ -126,48 +173,136 @@ export class LocalProjectContextController {
...
+                        }
+                        const userGitIgnoreFilterByFile = new Map(
+                            await Promise.all(
+                                localGitIgnoreFiles.map(async filePath => {
+                                    const lines: string[] = []
+                                    try {
+                                        const fileStream = fs.createReadStream(filePath)
+                                        const rl = readline.createInterface({
+                                            input: fileStream,
+                                            crlfDelay: Infinity,
```
justinmk3, line 257: ?? why is readline needed ?

server/aws-lsp-codewhisperer/package.json
taldekar: @justinmk3 Oh gotcha, I'll update the description. I was using ignore because I wanted to enforce the `.gitignore` convention in the ignore patterns that are passed in to the lsp, and ignore let me do that easily and was very well tested. I'll look into gitignore-parser, but it looks like it's more for parsing actual .gitignore files.

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextController.ts
taldekar: @justinmk3 I think this would be easier done with gitignore-parser, but I had to parse the file because ignore doesn't support this functionality _:(

taldekar: gitignore-parser would be really useful here, but it doesn't support adding a list of rules which is important for the rules we get from the IDEs.

justinmk3: > but I had to parse the file because ignore doesn't support this functionality _:(

readline should not be needed for that.

taldekar: Gotcha, I'll look into other web-compatible options

taldekar: @justinmk3 local context isn't meant to be web-compatible at this time, but I have removed the `readline` dependency anyways for a more concise approach.

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -136,76 +187,135 @@ export class LocalProjectContextController {
...
+
+        const controller = new AbortController()
+
+        const workspaceSourceFiles = await Promise.all(
+            workspaceFolders.map(async (folder: WorkspaceFolder) => {
+                const absolutePath = path.resolve(URI.parse(folder.uri).fsPath)
+                const localGitIgnoreFiles: string[] = []
+
+                const crawler = new fdir()
+                    .withSymlinks({ resolvePaths: !includeSymLinks })
```
justinmk3, line 247: symlinks can cause loops, or just really poor performance on complex directories. I assume `includeSymLinks` decides whether symlinks are followed. It should probably be false by default.

justinmk3 (approved):

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -11,26 +12,55 @@ import type {
...
-    private readonly fileExtensions: string[]
     private readonly clientName: string
-    private readonly log: Logging
+    private readonly log?: Logging
 
-    constructor(clientName: string, workspaceFolders: WorkspaceFolder[], logging: Logging) {
-        this.fileExtensions = Object.keys(languageByExtension)
+    private ignoreFilePatterns?: string[]
+    private includeSymlinks?: boolean
+    private maxFileSizeMb?: number
```
jpinkney-aws, line 53: should this by MB? or is are they actually representing bits?

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -11,26 +12,55 @@ import type {
...
+    private maxIndexSizeMb?: number
+    private respectUserGitIgnores?: boolean
+    private indexCacheDirPath: string = path.join(homedir, '.aws', 'amazonq', 'cache')
+
+    private readonly fileExtensions: string[] = Object.keys(languageByExtension)
+    private readonly DEFAULT_MAX_INDEX_SIZE = 2048
+    private readonly DEFAULT_MAX_FILE_SIZE = 10
+    private readonly MB_TO_BYTES = 1024 * 1024
+
+    constructor(clientName: string, workspaceFolders: WorkspaceFolder[], logging?: Logging) {
```
jpinkney-aws, line 63: why has the logger now become optional?

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
taldekar: Yup it's false by default. Not sure if we want to surface this setting to the user yet, but loops and poor performance are definitely a possible risk. We have an upper bound on the index size though, so we wouldn't explicitly break anything.

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
taldekar: Oh I see, yup it should be MB. I'll make the change.

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
taldekar: I made this change for testing on my local. I'll revert this change before merging.

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
taldekar: @jpinkney-aws fixed this!

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
taldekar: @jpinkney-aws fixed this as well!

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -11,26 +12,55 @@ import type {
 } from 'local-indexing'
 import { URI } from 'vscode-uri'
 
-const fs = require('fs').promises
+const ignore = require('ignore')
+const { fdir } = require('fdir')
+const fs = require('fs')
```
kmile, line 17: are these necessary to be `require`d or can they be `import` as well?

taldekar: I believe these should be importable as well. I can make the change.

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -43,13 +73,28 @@ export class LocalProjectContextController {
         return this.instance
     }
 
-    public async init(vectorLib?: any): Promise<void> {
+    public async init({
```
kmile, line 76: Do we have a way to deliver a no-op version of the entire workspace indexing that has no dependencies on `os` and `fs` and `vectorLib`? We need to make sure that this bundle can be loaded in a browser environment eventually, with local indexing either disabled or replaced by something suitable for the browser. This is not urgent but we'll need to come up with a solution eventually.

kmile (approved):

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
taldekar: I believe if the vector lib cannot be resolved (in the browser environment for instance), all calls to the controller are effectively no-ops. But @breedloj would be able to answer that better.

kmile (approved):

PR #968, opened 2025-04-15
chore: sync executeBash tool with vsc version

Hweinstock:
## Problem

## Solution
Sync up to most recent changes: https://github.com/aws/aws-toolkit-vscode/blob/741c2c481bcf0dca2d9554e32dc91d8514b1b1d1/packages/core/src/codewhispererChat/tools/executeBash.ts#L132

## Notes
There are two items marked for future work:
- expose a function to check if workspace contains a file (in runtimes). 
- generalize a proper way to handle tool cancellation. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
@@ -1,4 +1,4 @@
-// Port from VSC https://github.com/aws/aws-toolkit-vscode/blob/093d5bcbce777c88cf18c76b52738610263d1fc0/packages/core/src/codewhispererChat/tools/executeBash.ts#L134
+// Port from VSC https://github.com/aws/aws-toolkit-vscode/blob/741c2c481bcf0dca2d9554e32dc91d8514b1b1d1/packages/core/src/codewhispererChat/tools/executeBash.ts#L134
 
 import { InvokeOutput } from './toolShared'
 import { split } from 'shlex'
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
@@ -3,16 +3,20 @@
 import { InvokeOutput } from './toolShared'
 import { split } from 'shlex'
 import { Logging } from '@aws/language-server-runtimes/server-interface'
-import { processUtils } from '@aws/lsp-core'
+import { processUtils, workspaceUtils } from '@aws/lsp-core'
+import { CancellationToken } from 'vscode-languageserver'
 import { getUserHomeDir, sanitize } from '@aws/lsp-core/out/util/path'
+import { ChildProcess, ChildProcessOptions } from '@aws/lsp-core/out/util/processUtils'
+// eslint-disable-next-line import/no-nodejs-modules
+import { isAbsolute, join } from 'path' // Safe to import on web since this is part of path-browserify
+import { Features } from '../../types'
 
 export enum CommandCategory {
     ReadOnly,
-    HighRisk,
+    Mutate,
     Destructive,
 }
 
-export const dangerousPatterns = new Set(['<(', '$(', '`'])
 export const splitOperators = new Set(['|', '&&', '||', '>'])
 export const splitOperatorsArray = Array.from(splitOperators)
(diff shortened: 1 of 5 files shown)


jpinkney-aws (approved):
Can we put the todos in a sim somewhere? just so they're tracked

Hweinstock:
Good call out. I have one for the workspace file check, but I'll add one for the other, I'm unsure if @kmile has already started thinking about cancellation.

kmile:
We currently check each agent loop whether it's cancelled:

https://github.com/aws/language-servers/pull/970/files#diff-0da876fbe798d89b1ae25052ceca6949c8b334ed02f207f1f5c45b3d5dbeebf0R226

This should probably move to each tool use as well.

I don't know if we can / should cancel tools themselves. I can imagine longer-running, multi-step tools have a need to be signaled that their output is no longer needed. The semantics of cancellation would be up to each tool to implement. I think we can start with checking before starting the tool use, and later expand to cancellation mid-tool.

Hweinstock:
Verified these are still the latest changes (as of writing this comment).

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolServer.ts
```diff
@@ -29,8 +29,8 @@ export const FsToolsServer: Server = ({ workspace, logging, agent }) => {
     return () => {}
 }
 
-export const BashToolsServer: Server = ({ logging, agent }) => {
-    const bashTool = new ExecuteBash(logging)
+export const BashToolsServer: Server = ({ logging, workspace, agent }) => {
+    const bashTool = new ExecuteBash({ logging, workspace })
     agent.addTool(bashTool.getSpec(), (input: ExecuteBashParams) => bashTool.invoke(input))
```
kmile, line 34: I'm assuming we'd wrap the user acceptance if required around the `invoke` here?

kmile: Or inside the `agenticChatController` if we have access to everything we need there.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -1,18 +1,21 @@
-// Port from VSC https://github.com/aws/aws-toolkit-vscode/blob/093d5bcbce777c88cf18c76b52738610263d1fc0/packages/core/src/codewhispererChat/tools/executeBash.ts#L134
+// Port from VSC https://github.com/aws/aws-toolkit-vscode/blob/741c2c481bcf0dca2d9554e32dc91d8514b1b1d1/packages/core/src/codewhispererChat/tools/executeBash.ts#L134
 
 import { InvokeOutput } from './toolShared'
 import { split } from 'shlex'
 import { Logging } from '@aws/language-server-runtimes/server-interface'
-import { processUtils } from '@aws/lsp-core'
-import { getUserHomeDir, sanitize } from '@aws/lsp-core/out/util/path'
+import { processUtils, workspaceUtils } from '@aws/lsp-core'
+import { CancellationToken } from 'vscode-languageserver'
```
kmile, line 7: I don't think we want to take a dependency on LSP cancellation here, but instead wrap it in our own type to prevent pulling in `vscode-languageserver` in tools.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -180,76 +195,162 @@ export class ExecuteBash {
...
-    public async invoke(params: ExecuteBashParams, updates?: WritableStream): Promise<InvokeOutput> {
-        await this.validate(this.logger, params.command)
-        const cwd = params.cwd ? sanitize(params.cwd) : getUserHomeDir()
-        this.logger.info(`Invoking bash command: '${params.command}' in cwd: '${cwd}'`)
+    private looksLikePath(arg: string): boolean {
+        return arg.startsWith('/') || arg.startsWith('./') || arg.startsWith('../')
+    }
+
+    // TODO: generalize cancellation logic for tools.
+    public async invoke(
```
kmile, line 236: Going over the signature of `invoke`, this doesn't match the signature in `useTool` anymore (well, the optional parameters will never be supplied).

So that means that the `cancellationToken` needs to be inserted inside `BashToolServer:34` which then needs to grab it from the `chatController`?

kmile (approved):

PR #969, opened 2025-04-15
chore: resync listDir tool with VSC

Hweinstock:
## Problem

## Solution
- Sync up to most recent changes: https://github.com/aws/aws-toolkit-vscode/blob/741c2c481bcf0dca2d9554e32dc91d8514b1b1d1/packages/core/src/codewhispererChat/tools/listDirectory.ts#L19
- In VSC, they re-implement a regex type checker. Instead of doing that I simplified the code to use regex directly. See [here](https://github.com/aws/aws-toolkit-vscode/blob/0eea1d8ca6e25243609a07dc2a2c31886b224baa/packages/core/src/codewhispererChat/tools/listDirectory.ts#L19) for the VSC implementation. 
- Centralize path validation object, rather than having two identical implementations. 

## License

### DIFF
--- a/core/aws-lsp-core/src/util/workspaceUtils.ts
+++ b/core/aws-lsp-core/src/util/workspaceUtils.ts
@@ -10,6 +10,7 @@ export async function readDirectoryRecursively(
     folderPath: string,
     options?: {
         maxDepth?: number
+        excludePatterns?: (string | RegExp)[]
         customFormatCallback?: (entry: Dirent) => string
         failOnError?: boolean
     }
@@ -46,10 +47,13 @@ export async function readDirectoryRecursively(
             features.logging.warn(errMsg)
             continue
         }
-
+        // TODO: add tests here
         for (const entry of entries) {
-            results.push(formatter(entry))
             const childPath = getEntryPath(entry)
+            if (options?.excludePatterns?.some(pattern => new RegExp(pattern).test(childPath))) {
+                continue
+            }
+            results.push(formatter(entry))
             if (entry.isDirectory() && (options?.maxDepth === undefined || depth < options?.maxDepth)) {
                 queue.push({ filepath: childPath, depth: depth + 1 })
             }
@@ -78,4 +82,10 @@ export function formatListing(entry: Dirent): string {
 
 export function getEntryPath(entry: Dirent) {
     return path.join(entry.parentPath, entry.name)
+}
+
+// TODO: port this to runtimes?
+export async function inWorkspace(workspace: Features['workspace'], filepath: string) {
+    const r = await workspace.getTextDocument(filepath)
+    return r !== undefined
 }
--- a/core/aws-lsp-core/src/util/workspaceUtils.ts
+++ b/core/aws-lsp-core/src/util/workspaceUtils.ts
@@ -48,7 +48,6 @@ export async function readDirectoryRecursively(
             features.logging.warn(errMsg)
             continue
         }
-        // TODO: add tests here
         for (const entry of entries) {
             const childPath = getEntryPath(entry)
             if (options?.excludePatterns?.some(pattern => new RegExp(pattern).test(childPath))) {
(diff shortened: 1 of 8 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/listDirectory.ts
```diff
@@ -18,8 +19,24 @@ export class ListDirectory {
         this.workspace = features.workspace
     }
 
-    public async queueDescription(params: ListDirectoryParams, updates: WritableStream) {
+    public async validate(params: ListDirectoryParams): Promise<void> {
+        if (params.maxDepth !== undefined && params.maxDepth < 0) {
+            throw new Error('MaxDepth cannot be negative.')
+        }
```
kmile, line 25: I think we can embed this in the JSON schema as well to define the input boundaries in one place.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/listDirectory.ts
```diff
@@ -18,8 +19,24 @@ export class ListDirectory {
...
+        await validatePath(params.path, this.workspace.fs.exists)
+    }
+
+    public async queueDescription(params: ListDirectoryParams, updates: WritableStream, requiresAcceptance: boolean) {
```
kmile, line 29: It's still a bit unclear how we're going to use this method.

core/aws-lsp-core/src/util/workspaceUtils.ts
```diff
@@ -60,22 +64,27 @@ export async function readDirectoryRecursively(
...
     return `${typeChar} ${path.join(entry.parentPath, entry.name)}`
 }
 
 export function getEntryPath(entry: Dirent) {
     return path.join(entry.parentPath, entry.name)
 }
+
+// TODO: port this to runtimes?
+export async function inWorkspace(workspace: Features['workspace'], filepath: string) {
+    return (await workspace.getTextDocument(URI.file(filepath).toString())) !== undefined
```
kmile, line 89: What does this do? I believer the `getTextDocument` method will only work for files that have their contents synced to the LSP server. i.e., files that are currently open.

The path also looks to be a folder, while `getTextDocument` is expecting file paths.

kmile: I assume the purpose of this check, given its usage for determining validation, is that we want the `listDirectory` tool to always succeed if it's looking for subfolders of the workspace folder?

If that's the case, you probably want to compare against `workspaceFolders?` inside https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize

core/aws-lsp-core/src/util/workspaceUtils.ts
```diff
@@ -46,10 +48,12 @@ export async function readDirectoryRecursively(
             features.logging.warn(errMsg)
             continue
         }
-
         for (const entry of entries) {
-            results.push(formatter(entry))
             const childPath = getEntryPath(entry)
+            if (options?.excludePatterns?.some(pattern => new RegExp(pattern).test(childPath))) {
```
kmile, line 53: nit: we can get regex literals or cache the `new Regex` here instead of re-compiling the same list on every pattern evaluation?

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/listDirectory.ts
Hweinstock: I don't think it will be used at this point given how we've structured our UI. If we make the decision to allow tools to write directly to chat, this is what could be used, however I agree with your earlier assessment that it isn't the right approach down the line.

core/aws-lsp-core/src/util/workspaceUtils.ts
Hweinstock: Thanks for calling that out, and that alternative solution makes sense. I'll also track a follow up to runtimes to add this info in a docstring to avoid future confusion.

core/aws-lsp-core/src/util/workspaceUtils.ts
Hweinstock: Not a nit, good call out. I will address this in a follow up.

PR #970, opened 2025-04-16
Agentic chat

kmile:
## Problem
Tool Results need to be passed back into the model for the next request.

## Solution
Add user and assistant responses to the history, and add tool results to the current message.

## License

### DIFF
--- a/app/aws-lsp-codewhisperer-runtimes/src/agent-standalone.ts
+++ b/app/aws-lsp-codewhisperer-runtimes/src/agent-standalone.ts
@@ -8,6 +8,7 @@ import {
     QNetTransformServerTokenProxy,
 } from '@aws/lsp-codewhisperer'
 import { IdentityServer } from '@aws/lsp-identity'
+import { FsToolsServer } from '@aws/lsp-codewhisperer/out/language-server/agenticChat/tools/toolServer'
 
 const MAJOR = 0
 const MINOR = 1
@@ -23,6 +24,7 @@ const props: RuntimeProps = {
         QNetTransformServerTokenProxy,
         QAgenticChatServerProxy,
         IdentityServer.create,
+        FsToolsServer,
     ],
     name: 'AWS CodeWhisperer',
 }
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -429,7 +429,7 @@ describe('AgenticChatController', () => {
             assert.deepStrictEqual(
                 secondCallArgs.conversationState?.currentMessage?.userInputMessage?.userInputMessageContext
                     ?.toolResults[0].content[0].json,
-                { result: mockToolResult }
+                mockToolResult
             )
 
             // Verify that the history was updated correctly
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -442,6 +442,141 @@ describe('AgenticChatController', () => {
             assert.deepStrictEqual(chatResult, expectedCompleteChatResult)
         })
 
+        it('propagates tool execution errors to the model in toolResults', async () => {
+            // First response includes a tool use request
+            const mockToolUseId = 'mock-tool-use-id-error'
+            const mockToolName = 'mock-tool-name'
+            const mockToolInput = JSON.stringify({ param1: 'value1' })
+            const mockErrorMessage = 'Tool execution failed with an error'
+
+            const mockToolUseResponseList: ChatResponseStream[] = [
+                {
+                    messageMetadataEvent: {
+                        conversationId: mockConversationId,
+                    },
+                },
+                {
+                    assistantResponseEvent: {
+                        content: 'I need to use a tool that will fail. ',
+                    },
+                },
+                {
+                    toolUseEvent: {
+                        toolUseId: mockToolUseId,
+                        name: mockToolName,
+                        input: mockToolInput,
+                        stop: true,
+                    },
+                },
+            ]
+
+            // Second response after tool execution error
+            const mockFinalResponseList: ChatResponseStream[] = [
+                {
+                    messageMetadataEvent: {
+                        conversationId: mockConversationId,
+                    },
+                },
+                {
+                    assistantResponseEvent: {
+                        content: 'I see the tool failed with error: ',
+                    },
+                },
+                {
+                    assistantResponseEvent: {
+                        content: mockErrorMessage,
+                    },
+                },
+            ]
+
+            // Reset the stub and set up to return different responses on consecutive calls
+            generateAssistantResponseStub.restore()
+            generateAssistantResponseStub = sinon.stub(CodeWhispererStreaming.prototype, 'generateAssistantResponse')
+
+            generateAssistantResponseStub.onFirstCall().returns(
+                Promise.resolve({
+                    $metadata: {
+                        requestId: mockMessageId,
+                    },
+                    generateAssistantResponseResponse: createIterableResponse(mockToolUseResponseList),
+                })
+            )
+
+            generateAssistantResponseStub.onSecondCall().returns(
+                Promise.resolve({
+                    $metadata: {
+                        requestId: mockMessageId,
+                    },
+                    generateAssistantResponseResponse: createIterableResponse(mockFinalResponseList),
+                })
+            )
+
+            // Reset the runTool stub and make it throw an error
+            const runToolStub = testFeatures.agent.runTool as sinon.SinonStub
+            runToolStub.reset()
+            runToolStub.rejects(new Error(mockErrorMessage))
+
+            // Make the request
+            const chatResultPromise = chatController.onChatPrompt(
+                { tabId: mockTabId, prompt: { prompt: 'Hello with failing tool' } },
+                mockCancellationToken
+            )
+
+            const chatResult = await chatResultPromise
+
+            // Verify that generateAssistantResponse was called twice
+            sinon.assert.calledTwice(generateAssistantResponseStub)
+
+            // Verify that the tool was executed
+            sinon.assert.calledOnce(runToolStub)
+            sinon.assert.calledWith(runToolStub, mockToolName, JSON.parse(mockToolInput))
+
+            // Verify that the second request included the tool error in the toolResults with status 'error'
+            const secondCallArgs = generateAssistantResponseStub.secondCall.args[0]
+            assert.ok(
+                secondCallArgs.conversationState?.currentMessage?.userInputMessage?.userInputMessageContext?.toolResults
+            )
+            assert.strictEqual(
+                secondCallArgs.conversationState?.currentMessage?.userInputMessage?.userInputMessageContext?.toolResults
+                    .length,
+                1
+            )
+            assert.strictEqual(
+                secondCallArgs.conversationState?.currentMessage?.userInputMessage?.userInputMessageContext
+                    ?.toolResults[0].toolUseId,
+                mockToolUseId
+            )
+            assert.strictEqual(
+                secondCallArgs.conversationState?.currentMessage?.userInputMessage?.userInputMessageContext
+                    ?.toolResults[0].status,
+                'error'
+            )
+            assert.deepStrictEqual(
+                secondCallArgs.conversationState?.currentMessage?.userInputMessage?.userInputMessageContext
+                    ?.toolResults[0].content[0].json,
+                { error: mockErrorMessage }
+            )
+
+            // Verify that the history was updated correctly
+            assert.ok(secondCallArgs.conversationState?.history)
+            assert.strictEqual(secondCallArgs.conversationState?.history.length, 2)
+            assert.ok(secondCallArgs.conversationState?.history[0].userInputMessage)
+            assert.ok(secondCallArgs.conversationState?.history[1].assistantResponseMessage)
+
+            // Create expected result format matching the actual format
+            const expectedErrorChatResult: ChatResult = {
+                messageId: mockMessageId,
+                body: 'I see the tool failed with error: Tool execution failed with an error',
+                canBeVoted: true,
+                codeReference: undefined,
+                followUp: undefined,
+                relatedContent: undefined,
+            }
+
+            // Verify the final result includes both messages
+            assert.deepStrictEqual(chatResult, expectedErrorChatResult)
+        })
+
         it('handles multiple iterations of tool uses with proper history updates', async () => {
             // First response includes a tool use request
             const mockToolUseId1 = 'mock-tool-use-id-1'
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -154,6 +154,13 @@ describe('AgenticChatController', () => {
             rm: sinon.stub().resolves(),
         }
 
+        // Add agent with runTool method to testFeatures
+        testFeatures.agent = {
+            runTool: sinon.stub().resolves({}),
+            getTools: sinon.stub().returns([]),
+            addTool: sinon.stub().resolves(),
+        }
+
         // @ts-ignore
         const cachedInitializeParams: InitializeParams = {
             initializationOptions: {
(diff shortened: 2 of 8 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts
```diff
@@ -103,6 +103,66 @@ export const QAgenticChatServer =
...
+                },
+                required: ['prompt'],
+            },
+        } as const, async input => {
+            return input.prompt?.length
+        })
+
+        agent.addTool({
+            name: 'get_open_workspace_files',
+            description: 'Use the LSP document synchronization to read a list of all open documents in the current workspace',
```
jpinkney-aws, line 124: I wasn't at the bug bash last week, were these the starter set of lsp tools that we needed? or do more need to be added?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts
```diff
@@ -103,6 +103,66 @@ export const QAgenticChatServer =
             return chatController.onEndChat(...params)
         })
 
+        agent.addTool({
```
jpinkney-aws, line 106: is this tool needed or is it just an example tool?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatEventParser.ts
```diff
@@ -128,7 +128,15 @@ export class AgenticChatEventParser implements ChatResult {
                 }
 
                 if (toolUseEvent.stop) {
-                    console.log(`ToolUseEvent: ${toolUseId} ${name} ${this.toolUses[toolUseId]?.input}`)
+                    const parsedInput =
```
jpinkney-aws, line 131: Could this be simplfied by doing something like
```
const input = this.toolUses[toolUseId].input;
const parsedInput = typeof input === 'string' ? JSON.parse(input || '{}') : input;
```

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -165,141 +153,360 @@ export class AgenticChatController implements ChatHandlers {
...
+                break
             }
 
-            this.#log(`Q api request error ${err instanceof Error ? err.message : 'unknown'}`)
-            return new ResponseError<ChatResult>(
-                LSPErrorCodes.RequestFailed,
-                err instanceof Error ? err.message : 'Unknown request error'
-            )
-        }
+            // Phase 3: Request Execution
```
jpinkney-aws, line 241: are we missing a phase 2?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -165,141 +153,360 @@ export class AgenticChatController implements ChatHandlers {
...
+        session: ChatSessionService,
+        metric: Metric<CombinedConversationEvent>,
+        partialResultToken?: string | number,
+        conversationIdentifier?: string,
+        token?: CancellationToken
+    ): Promise<Result<AgenticChatResultWithMetadata, string>> {
+        let currentRequestInput = { ...initialRequestInput }
+        let finalResult: Result<AgenticChatResultWithMetadata, string> | null = null
+        let iterationCount = 0
+        const maxIterations = 10 // Safety limit to prevent infinite loops
```
jpinkney-aws, line 227: Does the VSCode implementation have this safety? or do they just let it run?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -165,141 +153,360 @@ export class AgenticChatController implements ChatHandlers {
...
             })
-            // Save question/answer interaction to chat history
-            if (params.prompt.prompt && session.conversationId && result.data?.chatResult.body) {
-                this.#chatHistoryDb.addMessage(params.tabId, 'cwc', session.conversationId, {
-                    body: params.prompt.prompt,
-                    type: 'prompt' as any,
+        }
+
+        if (iterationCount >= maxIterations) {
+            this.#log('Agent loop reached maximum iterations limit')
```
jpinkney-aws, line 302: would be nice in the future to have telemetry for this, that way we could see if agentic chat was hitting this too often

jpinkney-aws (commented):
Looks basically like I expected after having a walkthrough with the agentic chat yesterday. just minor comments

server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts
kmile: These are a first stab at providing LSP tools. In my experience the `Fs` based tools took a bit of searching in order to find what they were looking for, and just adding these for open files made it a lot more reliable and faster.

I expect that we either continue to improve `Fs` based tools, or provide a set of `LSP` based tools in a separate `Server`

kmile: One I definitely see a lot of value in is `applyWorkspaceEdit`.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: yeah, these have been moved around and refactored quite a bit, I'll make them consistent again.

The idea being:

1. prepare prompt
2. prepare request (with context)
3. run request
4. process response (with tools)
5. finalize results

With 2-4 being the agent loop, and 1+5 being the before and after.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: I think they let it run. But in the VSCode implementation a lot of tools interrupt the loop and ask for user input (something we haven't implemented yet for these specific cases) so there might not be that much iteration before the user has a chance to interrupt.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatEventParser.ts
kmile: Not entirely. When calling tools without inputs, then the input in the model response can either be missing, an empty string, or an empty object.

The Schema validation of Tool Usage however always expects an object at the top level, so if the string is empty, we replace it with `{}` instead.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
```diff
@@ -305,6 +312,326 @@ describe('AgenticChatController', () => {
...
+                },
+                {
+                    assistantResponseEvent: {
+                        content: '!',
+                    },
+                },
+            ]
+
+            // Reset the stub and set up to return different responses on consecutive calls
+            generateAssistantResponseStub.restore()
```
volodkevych, line 368: can we do these restores in `afterEach`?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -155,141 +143,360 @@ export class AgenticChatController implements ChatHandlers {
...
+    /**
+     * Prepares the initial request input for the chat prompt
+     */
+    async #prepareRequestInput(
+        params: ChatParams,
+        session: ChatSessionService,
+        triggerContext: TriggerContext
+    ): Promise<GenerateAssistantResponseCommandInput> {
+        this.#log('Preparing request input')
+        const profileArn = AmazonQTokenServiceManager.getInstance(this.#features).getActiveProfileArn()
```
viktorsaws, line 185: nit: instead of calling `getInstance` we can use `this.#amazonQServiceManager` object, as it must be passed at controller initialization anyway

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -155,141 +143,360 @@ export class AgenticChatController implements ChatHandlers {
...
+        let currentRequestInput = { ...initialRequestInput }
+        let finalResult: Result<AgenticChatResultWithMetadata, string> | null = null
+        let iterationCount = 0
+        const maxIterations = 10 // Safety limit to prevent infinite loops
+
+        metric.recordStart()
+
+        while (iterationCount < maxIterations) {
+            iterationCount++
+            this.#log(`Agent loop iteration ${iterationCount} for conversation id:`, conversationIdentifier || '')
```
viktorsaws, line 223: We should log this (and other logs in this loop) with `.debug` log level, it's going to be very noisy and pollute log stream we send to extension

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -155,141 +143,360 @@ export class AgenticChatController implements ChatHandlers {
...
                 }),
-                params.partialResultToken
+                partialResultToken
             )
 
-            session.conversationId = result.data?.conversationId
-            this.#log('Session conversation id:', session.conversationId || '')
+            // Store the conversation ID from the first response
+            if (iterationCount === 1 && result.data?.conversationId) {
+                session.conversationId = result.data.conversationId
```
viktorsaws, line 248: Is it an error if we don't have `conversationId` on first response?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolServer.ts
```diff
@@ -14,14 +14,14 @@ export const FsToolsServer: Server = ({ workspace, logging, agent }) => {
         // TODO: fill in logic for handling invalid tool invocations
         // TODO: implement chat streaming via queueDescription.
         await fsReadTool.validate(input)
-        await fsReadTool.invoke(input)
+        return await fsReadTool.invoke(input)
```
volodkevych, line 17: `await` is not needed in this syntax, right?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts
volodkevych: Yeah, we could move those tools to a separate server - for readability. We could also include this example tools server for testing but exclude for prod bundle.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatEventParser.ts
```diff
@@ -128,7 +128,15 @@ export class AgenticChatEventParser implements ChatResult {
...
+                        typeof this.toolUses[toolUseId].input === 'string'
+                            ? JSON.parse(this.toolUses[toolUseId].input === '' ? '{}' : this.toolUses[toolUseId].input)
```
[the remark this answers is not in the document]
kmile: Yes. These specific errors are currently handled 'outside' the agent loop, which means we're not re-prompting the agent with this error as a tool result, since errors during parsing indicate a server error, the input received from the model is not valid JSON. We should expect all inputs from the model to match the JSON schema.

Errors during tool execution for instance are handled inside the agent loop, and they are submitted to the model as tool results, so those are potentially recoverable.

Do you think we should handle these specific parse errors a different way?

Do you think we should

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -155,141 +143,360 @@ export class AgenticChatController implements ChatHandlers {
...
+                toolResultContent = { json: { items: toolResult.result } }
+            } else if (typeof toolResult.result === 'object') {
+                toolResultContent = { json: toolResult.result }
+            } else toolResultContent = { text: JSON.stringify(toolResult.result) }
+
+            this.#log(`ToolResult: ${JSON.stringify(toolResultContent)}`)
+            updatedRequestInput.conversationState!.currentMessage!.userInputMessage!.userInputMessageContext!.toolResults.push(
+                {
+                    toolUseId: toolResult.toolUseId,
+                    status: 'success',
```
volodkevych, line 380: If there are errors in `toolResults`, shouldn't we propagate it as an error here?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -155,141 +143,360 @@ export class AgenticChatController implements ChatHandlers {
...
-                    })
+            const currentMessage = currentRequestInput.conversationState?.currentMessage
 
-                    this.#telemetryController.emitStartConversationMetric(params.tabId, metric.metric)
-                }
+            // Process tool uses and update the request input for the next iteration
+            const toolResults = await this.#processToolUses(pendingToolUses)
+            currentRequestInput = this.#updateRequestInputWithToolResults(currentRequestInput, toolResults)
+
+            if (!currentRequestInput.conversationState!.history) {
```
Hweinstock, line 266: Is there a case where `converstationState` is undefined?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: You're right. We don't know whether a result is a success or failure from tools themselves, but we can catch the exceptions [here](https://github.com/aws/language-servers/pull/970/files/ed266b71f952c97c17195e53ab97463454553719#diff-0da876fbe798d89b1ae25052ceca6949c8b334ed02f207f1f5c45b3d5dbeebf0R339) and add this `result` property.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -155,141 +143,360 @@ export class AgenticChatController implements ChatHandlers {
...
+                    userInputMessageContext: currentMessage?.userInputMessage?.userInputMessageContext,
                 },
             })
-            // Save question/answer interaction to chat history
-            if (params.prompt.prompt && session.conversationId && result.data?.chatResult.body) {
-                this.#chatHistoryDb.addMessage(params.tabId, 'cwc', session.conversationId, {
-                    body: params.prompt.prompt,
-                    type: 'prompt' as any,
+
+            currentRequestInput.conversationState!.history.push({
```
Hweinstock, line 279: I see that we add the tool results to the history, but where does this get sent to the client?

app/aws-lsp-codewhisperer-runtimes/src/agent-standalone.ts
```diff
@@ -23,6 +24,7 @@ const props: RuntimeProps = {
         QNetTransformServerTokenProxy,
         QAgenticChatServerProxy,
         IdentityServer.create,
+        FsToolsServer,
```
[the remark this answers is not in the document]
kmile: Technically speaking, a later server can overwrite an earlier one based on Tool name:

https://github.com/aws/language-server-runtimes/blob/d914bec8ce1fc841c9b73fb622c68559be44917f/runtimes/runtimes/agent.ts#L32

Other than that, there is no impact on the `Agent` namespace. There are a couple others that treat servers in priority order, mainly here: https://github.com/aws/language-server-runtimes/blob/d914bec8ce1fc841c9b73fb622c68559be44917f/runtimes/runtimes/lsp/router/lspRouter.ts#L183

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -155,141 +143,360 @@ export class AgenticChatController implements ChatHandlers {
...
+        Array<{
+            toolUseId: string
+            name: string
+            result: any
+        }>
+    > {
+        const results = []
+
+        for (const toolUse of toolUses) {
+            if (!toolUse.name || !toolUse.toolUseId) continue
```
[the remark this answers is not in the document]
kmile: This is what we get back from the model. The [spec](https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/src.gen/%40amzn/codewhisperer-streaming/src/models/models_0.ts#L446) says `string | undefined` for some reason. Might not be smart to silently ignore them, it might be better to add some telemetry, but there's nothing we can do if this happens for whatever issue on the upstream API.

So either we `continue` here, throw an exception, or fail somewhere downstream at an unspecified moment.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatEventParser.ts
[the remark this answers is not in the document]
volodkevych: That's fine if the error is not recoverable on retry. However, if we try/catch and log here, it would be a bit more transparent and easier to troubleshoot chat requests failures.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: I don't believe so (https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts#L43)

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
volodkevych: I can see that `#processToolUses` already catches errors and we get those errors in `toolResult.result`. So that's I guess what can be used to define status here.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: Here: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts#L611

We currently show the `AssistantResposne` text in the chat,  but we could decide otherwise. VsCode hides this from the chat, and let's the tools use an [explanation](https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts#L611) parameter to show to the customer.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: Updated to reflect this.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
```diff
@@ -305,6 +312,326 @@ describe('AgenticChatController', () => {
...
+            )
+
+            const chatResult = await chatResultPromise
+
+            // Verify that generateAssistantResponse was called twice
+            sinon.assert.calledTwice(generateAssistantResponseStub)
+
+            // Verify that the tool was executed
+            sinon.assert.calledOnce(runToolStub)
+            sinon.assert.calledWith(runToolStub, mockToolName, JSON.parse(mockToolInput))
```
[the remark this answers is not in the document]
kmile: That actually [does not exist](https://sinonjs.org/releases/latest/assertions/#sinonassertcalledoncewithexactlyspyorspycall-arg1-arg2-) for some reason?.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: Done.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: Makes sense, sounds like the same situation as the tool name.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/qAgenticChatServer.ts
kmile: Will follow up with an `LspToolsServer` with a couple "naive" tools we can iterate on.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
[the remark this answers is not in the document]
kmile: Yeah, I'll add a followup to track that.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: This gives the models response back to the client, but does not give the tool results back to the client. It relies on the model to summarize or capture the tool's results in its response. However, I assume this is an intentional decision since for example, we don't want to send the contents of a large file in chat. Is that a correct understanding?

kmile: Yes. The model summarizes the tool response in its output.

We could do something like (oversimplified):

```
this.#features.lsp.sendProgress(chatRequestType, partialResultToken, {
    body: `${toolResults.length > 0 ? '## Tools\n' : ''}${toolResults.join('\n')}${result.body}`
})
```

Or any other composition of the `ChatResult`.

volodkevych (approved):

PR #971, opened 2025-04-16
TEAM REVIEW : TESTING NEEDED : Make RemoveDuplicateNugetPackage failure a non-blocker for transformation

tomar-gunjan:
## Problem
When an error is thrown from RemoveDuplicateNugetPackage it causes the entire transformation to fail

## Solution
Adding a try catch anf just logging the error since RemoveDuplicateNugetPackage is not a necessary step for transformation.

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
@@ -61,17 +61,21 @@ export class ArtifactManager {
     }
 
     async removeDuplicateNugetPackagesFolder(request: StartTransformRequest) {
-        const packagesFolder = path.join(
-            this.workspacePath,
-            artifactFolderName,
-            sourceCodeFolderName,
-            packagesFolderName
-        )
-        if (fs.existsSync(packagesFolder)) {
-            fs.rmSync(packagesFolder, { recursive: true, force: true })
-            this.logging.log(
-                `Removed packages folder ${packagesFolder} from source code directory to be uploaded because it is a duplicate of references folder from artifacts`
+        try {
+            const packagesFolder = path.join(
+                this.workspacePath,
+                artifactFolderName,
+                sourceCodeFolderName,
+                packagesFolderName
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -61,17 +61,21 @@ export class ArtifactManager {
...
+                packagesFolderName
             )
+            if (fs.existsSync(packagesFolder)) {
+                fs.rmSync(packagesFolder, { recursive: true, force: true })
+                this.logging.log(
+                    `Removed packages folder ${packagesFolder} from source code directory to be uploaded because it is a duplicate of references folder from artifacts`
+                )
+            }
+        } catch (error) {
+            this.logging.log('Failed to remove duplicate nugget package folder: ' + error)
```
jonlouie, line 78: Replace: nugget package -> packages (keeps messaging consistent with the log message above])

huawenm (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -61,17 +61,21 @@ export class ArtifactManager {
...
+                packagesFolderName
             )
+            if (fs.existsSync(packagesFolder)) {
+                fs.rmSync(packagesFolder, { recursive: true, force: true })
+                this.logging.log(
+                    `Removed packages folder ${packagesFolder} from source code directory to be uploaded because it is a duplicate of references folder from artifacts`
+                )
+            }
+        } catch (error) {
+            this.logging.log('Failed to remove packages folder: ' + error)
```
pranav-firake, line 78: can we include {`packagesFolder`} here in log? 
also can we reword it somewhat like 'failed to remove packages folder {packagesFolder}, error received is  ' + error? 
This way customers will know that this is not error that may cause transformation to fail, it was just clean up error.

pranav-firake (approved):

PR #973, opened 2025-04-16
fix(chat-client): disable click event for empty history list item

viktorsaws:
## Problem
Click on empty history list item open chat window and sends `conversationClicked` event to language server.

## Solution
* Do not invoke click event on empty history item in chat client
* Do not handle click event on "empty" history item in language server

Screenshot 2025-04-16 at 13 25 21

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -20,6 +20,8 @@ import { Features } from '@aws/language-server-runtimes/server-interface/server'
 import { ConversationItemGroup } from '@aws/language-server-runtimes/protocol'
 import { getUserHomeDir } from '@aws/lsp-core/out/util/path'
 
+export const EMPTY_CONVERSATION_LIST_ID = 'empty'
```
viktorsaws, line 23: We could move `empty` key to `types` package to properly share it between Chat Client and language server.

volodkevych (approved):

PR #983, opened 2025-04-16
feat: enable inline project context in suggestion requests

nborges-aws:
## Problem
Supplemental Context included in suggestions requests currently does not include inline project context
## Solution
Port over vscode toolkit implementation to include project context in supplemental context sent in suggestions request. Leverages vector library and local project context addition to the language server component.
 
_**Note:**_ This code is reliant on PR #982 re-enabling the localProjectContextController's ability to find the vector library.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
@@ -72,21 +72,7 @@ export async function fetchSupplementalContextForSrc(
     if (supplementalContextConfig === undefined) {
         return supplementalContextConfig
     }
-
-    if (supplementalContextConfig === 'OpenTabs_BM25') {
-        const opentabsContextPromise = waitUntil(
-            async function () {
-                return await fetchOpenTabsContext(document, position, workspace, cancellationToken)
-            },
-            { timeout: supplementalContextTimeoutInMs, interval: 5, truthy: false }
-        )
-        const supContext = (await opentabsContextPromise) ?? []
-        return {
-            supplementalContextItems: supContext,
-            strategy: supContext.length === 0 ? 'Empty' : 'OpenTabs_BM25',
-        }
-    }
-
+    //TODO: add logic for other strategies once available
     if (supplementalContextConfig === 'codemap') {
         return await codemapContext(document, position, workspace, cancellationToken)
     }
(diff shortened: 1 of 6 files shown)


breedloj (approved):

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
```diff
@@ -55,17 +67,106 @@ export async function fetchSupplementalContextForSrc(
...
+): Promise<CodeWhispererSupplementalContextItem[]> {
+    const inputChunk: Chunk = getInputChunk(document, position, crossFileContextConfig.numberOfLinesEachChunk)
+    const fsPath = URI.parse(document.uri).fsPath
+    const inlineProjectContextRequest: QueryInlineProjectContextRequestV2 = {
+        query: inputChunk.content,
+        filePath: fsPath,
+        target,
     }
 
+    const controller = LocalProjectContextController.getInstance()
```
kmile, line 160: Do we ever `setInstance()`? I couldn't find any code initializing the singleton.

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
```diff
@@ -55,17 +67,106 @@ export async function fetchSupplementalContextForSrc(
...
-        return shouldProceed === undefined
-            ? undefined
-            : {
-                  supplementalContextItems: [],
-                  strategy: 'Empty',
-              }
+    const supplementalContextConfig = getSupplementalContextConfig(document.languageId)
+
+    if (supplementalContextConfig === undefined) {
+        return supplementalContextConfig
```
kmile, line 73: Is this intentionally returning `undefined`?

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
```diff
@@ -55,17 +67,106 @@ export async function fetchSupplementalContextForSrc(
...
+    }
+
+    if (supplementalContextConfig === 'OpenTabs_BM25') {
```
kmile, line 76: This never happens?

kmile (approved):

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
kmile: I think I found it :) https://github.com/aws/language-servers/blob/12dc8d767be42d04d50303143e1a551fb103bdc5/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts#L54

nborges-aws: The instance should be initialized on start up, but I should defensively check the instance exists for extra safety in case of failures. I'll make this change.

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
nborges-aws: Undefined is indeed intentional here. The caller of this method returns undefined as well if there is no context result.

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
nborges-aws: Yeah, `getSupplementalContextConfig` should be enhanced to return different strategies once the others are implemented (others are currently no-ops), which is what this conditional was targeting. I'll remove as its unreachable in the short term.

kmile (approved):

PR #984, opened 2025-04-17
feat: add text based tool updates for agentic-chat

Hweinstock:
## Problem
There is no way to tell what tools the agent is running and what their results are. 

Additionally, streaming content to the chat is currently done through `lsp.sendProgress` which does not provide meaningful interface to build up a response. For example, each call overwrites all previous data, and doesn't provide any guardrails to prevent this from happening accidentally. 

## Solution
- Add an abstraction that wraps `lsp.sendProgress` for streaming and writing result "blocks" back to the user. 
- Add a simple text based ui to stream these results as they are computed. 
- add emojis and markdown to be replaced by proper mynah UI components. 
- Add the moment we just dump all results to the chat. We could add hard-coded custom formatting options for known tools, but that raises the question what we should do for unknown tools (MCP). 

## Demo
- executeBash is enabled for the demo. 

https://github.com/user-attachments/assets/0b750dde-f20b-47c7-80b3-bbc16a06e9d2

## Known Issues
When the request to Q endpoint fails, the chat hangs. 

## License


kmile:
This looks amazing! 🕺

jpinkney-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -318,14 +331,17 @@ export class AgenticChatController implements ChatHandlers {
...
+        chatResultStream: AgenticChatResultStream
+    ): Promise<ToolResult[]> {
         const results: ToolResult[] = []
 
         for (const toolUse of toolUses) {
             if (!toolUse.name || !toolUse.toolUseId) continue
 
             try {
-                this.#debug(`Running tool ${toolUse.name} with input:`, JSON.stringify(toolUse.input))
+                await chatResultStream.writeResultBlock({ body: `${executeToolMessage(toolUse)}` })
```
kmile, line 344: Is it also possible to update a previous block? I imagine the UX being 'pending' .. 'in progress' ... 'done'.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -122,10 +124,21 @@ export class AgenticChatController implements ChatHandlers {
         return this.#tabBarController.onConversationClick(params)
     }
 
+    async #sendProgressToClient(chunk: ChatResult | string, partialResultToken?: string | number) {
+        if (!isNullish(partialResultToken)) {
+            await this.#features.lsp.sendProgress(chatRequestType, partialResultToken, chunk)
+        }
+    }
```
kmile, line 131: I love how you've been able to keep updating the client a separate concern from the agent loop and from the tool execution itself. Really flexible approach!

kmile (approved):
Awesome PR!

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: Once a block is added, its committed. This was an intentional choice to minimize interacting with state, and treat the block store as persistent. 

However, you could use the stream writer to implement a progress block that renders 'in progress', 'pending', etc. until the actual resulting block is committed. The streaming writer doesn't commit any blocks to the store until it is closed (in this example, the result is resolved).

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: Thanks, that way my goal!

PR #986, opened 2025-04-17
fix: avoid unnecessary vector library re-initialization on config update

breedloj:
## Problem
Vector library can be unnecessarily re-initialized on configuration changes even if the enablement setting does not change.

## Solution
Cache the current value and only enable/disable the library on an appropriate change.

## License


kmile (approved):

breedloj:
This change has been made redundant by #967. Closing.

PR #987, opened 2025-04-17
fix: add workspace folders to model context.

Hweinstock:
TODO: update failing tests. 
- The bug is still present after https://github.com/aws/language-servers/pull/948, but this PR needs to be refactored to work with these changes. 

## Problem
Q doesn't know what the workspace folder path is. When I ask it "what is the path to my workspace" it tries to use a few tools, then fails, then tries to find from the root directory which is where the problem originates.

This works on the VSC implementation. This is affecting how tools are used. 

## Solution
- Ensure it is passed as context similar to how agentic-chat feature branch does it: [https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/codewhispererChat/controllers/chat/chatRequest/converter.ts ](https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/codewhispererChat/controllers/chat/chatRequest/converter.ts) which is missing from the equivalent in Flare here: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts

## Verification and Testing
- added a unit test
- can be verified end-to-end with the following prompt:
image

- Before you'd get something like:
image

## License


Hweinstock:
Moved to https://github.com/aws/language-servers/pull/995

PR #988, opened 2025-04-18
feat(chat-client): support profile banner changes

vidkhath:
## Problem
Regional dev profile support for chat client.

## Solution
image

image

## License

### DIFF
--- a/chat-client/src/client/chat.test.ts
+++ b/chat-client/src/client/chat.test.ts
@@ -18,6 +18,7 @@ import { createChat } from './chat'
 import sinon = require('sinon')
 import { TELEMETRY } from '../contracts/serverContracts'
 import {
+    ENTER_FOCUS,
     ERROR_MESSAGE_TELEMETRY_EVENT,
     SEND_TO_PROMPT_TELEMETRY_EVENT,
     TAB_ADD_TELEMETRY_EVENT,
@@ -65,18 +66,33 @@ describe('Chat', () => {
         global.ResizeObserver = undefined
     })
 
-    it('publishes ready event and initial tab add event, when initialized', () => {
-        assert.callCount(clientApi.postMessage, 4)
+    it('publishes ready event when initialized', () => {
+        assert.calledWithExactly(clientApi.postMessage.firstCall, { command: READY_NOTIFICATION_METHOD })
+    })
 
-        assert.calledWithExactly(clientApi.postMessage.firstCall, {
+    it('creates initial tab when chat options are provided', () => {
+        const bannerText = 'This is a test banner message'
+        const eventParams = {
+            command: CHAT_OPTIONS,
+            params: {
+                chatNotifications: {
+                    bannerText: bannerText,
+                },
+            },
+        }
+        const sendToPromptEvent = createInboundEvent(eventParams)
+        window.dispatchEvent(sendToPromptEvent)
+
+        assert.calledWithExactly(clientApi.postMessage.firstCall, { command: READY_NOTIFICATION_METHOD })
+
+        assert.calledWithExactly(clientApi.postMessage.secondCall, {
             command: TELEMETRY,
-            params: { name: 'enterFocus' },
+            params: { name: ENTER_FOCUS },
         })
-        assert.calledWithExactly(clientApi.postMessage.secondCall, { command: READY_NOTIFICATION_METHOD })
 
         assert.calledWithExactly(clientApi.postMessage.thirdCall, {
             command: TAB_ADD_NOTIFICATION_METHOD,
-            params: { tabId: initialTabId },
+            params: { tabId: sinon.match.string },
         })
 
         assert.calledWithExactly(clientApi.postMessage.lastCall, {
@@ -84,7 +100,7 @@ describe('Chat', () => {
             params: {
                 triggerType: 'click',
                 name: TAB_ADD_TELEMETRY_EVENT,
-                tabId: initialTabId,
+                tabId: sinon.match.string,
             },
         })
     })
@@ -313,9 +329,9 @@ describe('Chat', () => {
             assert.called(TabFactory.prototype.enableHistory)
             // @ts-ignore
             assert.called(TabFactory.prototype.enableExport)
-        })
+        }).timeout(20000)
 
-        it('does not enable history and export features support if flags are falsy', () => {
+        it('does not enable history and export features support if flags are falsy', async () => {
             const chatOptionsRequest = createInboundEvent({
                 command: CHAT_OPTIONS,
                 params: {
@@ -329,7 +345,7 @@ describe('Chat', () => {
             assert.notCalled(TabFactory.prototype.enableHistory)
             // @ts-ignore
             assert.notCalled(TabFactory.prototype.enableExport)
-        })
+        }).timeout(20000)
     })
 
     describe('onGetSerializedChat', () => {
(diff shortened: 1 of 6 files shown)


chat-client/src/client/tabs/tabFactory.ts
```diff
@@ -80,6 +87,25 @@ export class TabFactory {
         return tabData
     }
 
+    public setProfileBanner(messages: ChatMessage[] | undefined) {
```
viktorsaws, line 90: nit: `setProfileBanner` is too specific name, and does not represent that several messages are supported. I'd make it not related to profile

chat-client/src/client/tabs/tabFactory.ts
```diff
@@ -80,6 +87,25 @@ export class TabFactory {
...
+            this.bannerMessage = messages[0]
+        }
+    }
+
+    public getProfileBanner(): ChatItem | undefined {
+        if (this.bannerMessage) {
+            return {
+                type: ChatItemType.ANSWER,
+                status: 'info',
+                messageId: 'regionProfile',
```
viktorsaws, line 102: same here, can we use something not related to profiles?

chat-client/src/client/chat.ts
```diff
@@ -175,10 +185,8 @@ export const createChat = (
                     tabFactory.enableExport()
                 }
 
-                const allExistingTabs: MynahUITabStoreModel = mynahUi.getAllTabs()
-                for (const tabId in allExistingTabs) {
-                    mynahUi.updateStore(tabId, tabFactory.getDefaultTabData())
-                }
+                const initialTabId = mynahApi.createTabId(true)
```
viktorsaws, line 188: Could you make a recording of how this looks like? Would be nice to see, as we're delaying rendering of first tab until after handshake with server is done

chat-client/src/client/tabs/tabFactory.ts
vidkhath: I think we need to hardcode this to regionProfile thats how MynahUI makes it in a banner on top of the chat

chat-client/src/client/chat.ts
vidkhath: sent on slack

chat-client/src/client/chat.test.ts
```diff
@@ -65,26 +66,41 @@ describe('Chat', () => {
...
+            params: {
+                chatNotifications: {
+                    bannerText: bannerText,
+                },
+            },
+        }
+        const sendToPromptEvent = createInboundEvent(eventParams)
+        window.dispatchEvent(sendToPromptEvent)
+
+        assert.calledWithExactly(clientApi.postMessage.firstCall, { command: READY_NOTIFICATION_METHOD })
```
volodkevych, line 86: Do we send `onReady` event again, the 2nd time after the `CHAT_OPTIONS`? If no, we should remove this check from here, because it's confusing.

chat-client/src/client/mynahUi.ts
```diff
@@ -223,6 +224,12 @@ export const createMynahUi = (
                 ],
                 ...(disclaimerCardActive ? { promptInputStickyCard: disclaimerCard } : {}),
             }
+
+            const tabStore = mynahUi.getTabData(tabId).getStore()
+
+            if (!tabStore?.tabMetadata || !tabStore.tabMetadata.openTabKey) {
```
volodkevych, line 230: What are we doing here? Why? Can you please add description in comment in the code, because this code intent is not super clear.

chat-client/src/client/mynahUi.ts
```diff
@@ -408,7 +415,7 @@ export const createMynahUi = (
                 // Update the tab defaults to hide the programmer mode card for new tabs
                 mynahUi.updateTabDefaults({
                     store: {
-                        chatItems: tabFactory.createTab(true, disclaimerCardActive, false).chatItems,
```
volodkevych, line 411: Were we creating 2 initial tabs before? Otherwise, why do we stop creating this tab here?

chat-client/src/client/mynahUi.ts
```diff
@@ -450,11 +452,11 @@ export const createMynahUi = (
         return tabId ? mynahUi.getAllTabs()[tabId]?.store : undefined
     }
 
-    const createTabId = (needWelcomeMessages: boolean = false, chatMessages?: ChatMessage[]) => {
```
volodkevych, line 453: Why did we remove those 2 params from here?

volodkevych: Ok, I see what happened to `chatMessages` now, but the question still remains for `needWelcomeMessages`.

chat-client/src/client/tabs/tabFactory.ts
```diff
@@ -80,6 +86,25 @@ export class TabFactory {
...
     }
 
+    public setBannerMessage(messages: ChatMessage[] | undefined) {
+        if (messages?.length) {
+            // For now this messages array is only populated with banner data hence we use the first item
+            this.bannerMessage = messages[0]
+        }
+    }
+
+    public getBannerMessage(): ChatItem | undefined {
```
volodkevych, line 96: nit: private

chat-client/src/client/mynahUi.ts
vidkhath: we are not stopping createTab, we are just grabbing the chatItems. `tabFactory.createTab` just returns `MynahUIDataModel`. Now we have a new function that returns chat items. So I just replaced these chatItems to the new function. Functionality wise nothing changes.

chat-client/src/client/tabs/tabFactory.ts
viktorsaws: Actually we can skip setting messageId in client, and let server set it when sending `ChatOptionsUpdateParams`. messageId: 'regionProfile' does not have impact on rendering of chat items  https://github.com/aws/language-server-runtimes/blob/main/types/chat.ts#L359-L361

chat-client/src/client/mynahUi.ts
```diff
@@ -419,14 +426,9 @@ export const createMynahUi = (
...
-        tabs: {
-            [initialTabId]: {
-                isSelected: true,
-                store: tabFactory.createTab(true, disclaimerCardActive, programmingModeCardActive),
-            },
-        },
+        tabs: {},
         defaults: {
-            store: tabFactory.createTab(true, false, programmingModeCardActive),
+            store: tabFactory.createTab(false),
```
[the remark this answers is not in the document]
vidkhath: initial implementation also set it to false

chat-client/src/client/chat.ts
```diff
@@ -181,8 +183,16 @@ export const createChat = (
             case GET_SERIALIZED_CHAT_REQUEST_METHOD:
                 mynahApi.getSerializedChat(message.requestId, message.params as GetSerializedChatParams)
                 break
+            case CHAT_OPTIONS_UPDATE_NOTIFICATION_METHOD:
```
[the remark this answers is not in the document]
vidkhath: We set it only once because banner message is a variable in the tab factory class. So if we set it once for this instance of tab factory then everytime we try to `getChatItems` for any number of tabs we still get the banner message

viktorsaws (approved):

chat-client/src/client/mynahUi.ts
```diff
@@ -83,6 +84,8 @@ const getTabPairProgrammingMode = (mynahUi: MynahUI, tabId: string) => {
     return promptInputOptions.find(item => item.id === 'pair-programmer-mode')?.value === 'true'
 }
 
+const openTabKey = 'openTab'
```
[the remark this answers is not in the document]
vidkhath: Yes, we pass this key with boolean true or false when tab is created

PR #998, opened 2025-04-18
feat: improve symlink handling

taldekar:
## Problem
When `resolvePaths` in `withSymlinks({ resolvePaths })` is set to true, `fdir` uses the absolute path to the symlinked directory. However, ignore, unless `allowRelativePaths` is toggled on, [throws on paths outside the base workspace folder directory](https://www.npmjs.com/package/ignore#:~:text=options.allowRelativePaths%3F%3A%20boolean%20since%205.2.0), so path's beginning with `..` will throw.

## Solution
Exclude directories that are outside the current workspace folder. If `includeSymlinks` is set to true, then use the relative path for the directory (i.e. with prefix `absolute/path/to/symlink/in/workspace/**`), which would arguably provide better context to Q.

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
@@ -113,7 +113,7 @@ export class LocalProjectContextController {
                 LocalProjectContextController.instance = this
 
                 const contextItems = await this.getContextCommandItems()
-                await this.contextCommandsProvider?.processContextCommandUpdate(contextItems)
+                await this.contextCommandsProvider.processContextCommandUpdate(contextItems)
                 void this.maybeUpdateCodeSymbols()
             } else {
                 this.log.warn(`Vector library could not be imported from: ${libraryPath}`)
@@ -128,7 +128,7 @@ export class LocalProjectContextController {
             await this._vecLib?.clear?.()
             this._vecLib = undefined
         }
-        this.contextCommandsProvider?.dispose()
+        this.contextCommandsProvider.dispose()
     }
 
     public async updateIndex(filePaths: string[], operation: UpdateMode): Promise<void> {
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -332,11 +332,14 @@ export class LocalProjectContextController {
...
-                        return filter.ignores(path.relative(absolutePath, dirPath))
+                        const relativePath = path.relative(absolutePath, dirPath)
+                        return relativePath.startsWith('..') || filter.ignores(relativePath)
                     })
                     .glob(...(fileExtensions?.map(ext => `**/*${ext}`) ?? []), '**/.gitignore')
                     .filter((filePath: string, isDirectory: boolean) => {
-                        if (isDirectory || filter.ignores(path.relative(absolutePath, filePath))) {
+                        const relativePath = path.relative(absolutePath, filePath)
+
+                        if (isDirectory || relativePath.startsWith('..') || filter.ignores(relativePath)) {
```
ege0zcan, line 342: I haven't used fdir before so might be an obvious question, but this looks like we use the same logic in both exclude and filter, what is the difference here

kmile: `exclude` stops crawling the folder altogether, while `filter` filters the crawled results. At least that's my interpretation: https://github.com/thecodrr/fdir/blob/master/documentation.md#filterfunction

kmile (approved):

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
taldekar: Yup, that's correct.

ege0zcan: I see that makes sense. What confused me was the `relativePath.startsWith('..') || filter.ignores(relativePath)` check happening in both places but I wasn't aware that `exclude` was only applied to directories so the same check was also needed in filter as well so that files were also correctly filtered

PR #999, opened 2025-04-18
fix(amazonq): move context command provider to agentic chat controller

zixlin7:
## Problem
- moving contextComandProvider into agenticChatController instead
## License


breedloj:
Changes LGTM - this breaks a direct coupling between the low level context controller and the agentic server that was causing build issues resulting in inline changes needing to be reverted: #991

breedloj (approved):

ege0zcan (approved):

PR #1001, opened 2025-04-19
refactor(amazonq): adding error handling for security scan and inline suggestions.

LokeshDogga13:
## Problem
With the chat functionality, Flare LSP has logic to intercept and detect when auth errors are thrown by the Q API for eg. when token has expired, it parses the error to determine it and crafts the auth follow up type that sends a response to the toolkit which indicates to the CHAT UI to show the Re-authenticate button.
We want to see if a similar error parsing can be supported with inline suggestions and security scans.

## Solution
Updated the error handling in security scan and inline suggestions. So, when the token expires we will throw a custom AmazonQServiceAuthenticationExpiredError and handle it appropriately on the client side.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
@@ -22,13 +22,23 @@ import { truncateOverlapWithRightContext } from './mergeRightUtils'
 import { CodeWhispererSession, SessionManager } from './session/sessionManager'
 import { CodePercentageTracker } from './codePercentage'
 import { CodeWhispererPerceivedLatencyEvent, CodeWhispererServiceInvocationEvent } from '../../shared/telemetry/types'
-import { getCompletionType, getEndPositionForAcceptedSuggestion, isAwsError, safeGet } from '../../shared/utils'
+import {
+    getCompletionType,
+    getEndPositionForAcceptedSuggestion,
+    getErrorMessage,
+    isAwsError,
+    safeGet,
+} from '../../shared/utils'
 import { makeUserContextObject } from '../../shared/telemetryUtils'
 import { fetchSupplementalContext } from '../../shared/supplementalContextUtil/supplementalContextUtil'
 import { textUtils } from '@aws/lsp-core'
 import { TelemetryService } from '../../shared/telemetry/telemetryService'
 import { AcceptedSuggestionEntry, CodeDiffTracker } from './codeDiffTracker'
-import { AmazonQError, AmazonQServiceInitializationError } from '../../shared/amazonQServiceManager/errors'
+import {
+    AmazonQError,
+    AmazonQServiceAuthenticationExpiredError,
+    AmazonQServiceInitializationError,
+} from '../../shared/amazonQServiceManager/errors'
 import {
     AmazonQBaseServiceManager,
     QServiceManagerFeatures,
@@ -36,6 +46,7 @@ import {
 import { initBaseTokenServiceManager } from '../../shared/amazonQServiceManager/AmazonQTokenServiceManager'
 import { AmazonQWorkspaceConfig } from '../../shared/amazonQServiceManager/configurationUtils'
 import { initBaseIAMServiceManager } from '../../shared/amazonQServiceManager/AmazonQIAMServiceManager'
+import { getAuthFollowUpType } from '../chat/utils'
 
 const EMPTY_RESULT = { sessionId: '', items: [] }
 export const CONTEXT_CHARACTERS_LIMIT = 10240
@@ -480,6 +491,10 @@ export const CodewhispererServerFactory =
                                 throw error
                             }
 
+                            const authFollowType = getAuthFollowUpType(error)
+                            if (authFollowType == 're-auth') {
+                                throw new AmazonQServiceAuthenticationExpiredError(getErrorMessage(error))
+                            }
                             return EMPTY_RESULT
                         })
                 })
--- a/server/aws-lsp-codewhisperer/src/language-server/securityScan/codeWhispererSecurityScanServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/securityScan/codeWhispererSecurityScanServer.ts
@@ -2,6 +2,8 @@ import {
     CancellationToken,
     ExecuteCommandParams,
     InitializeParams,
+    LSPErrorCodes,
+    ResponseError,
     Server,
 } from '@aws/language-server-runtimes/server-interface'
 import { performance } from 'perf_hooks'
@@ -16,6 +18,7 @@ import { SecurityScanEvent } from '../../shared/telemetry/types'
 import { getErrorMessage, parseJson } from '../../shared/utils'
 import { v4 as uuidv4 } from 'uuid'
 import { AmazonQTokenServiceManager } from '../../shared/amazonQServiceManager/AmazonQTokenServiceManager'
+import { getAuthFollowUpType } from '../chat/utils'
 
 const RunSecurityScanCommand = 'aws/codewhisperer/runSecurityScan'
 const CancelSecurityScanCommand = 'aws/codewhisperer/cancelSecurityScan'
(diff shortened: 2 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/securityScan/codeWhispererSecurityScanServer.ts
```diff
@@ -192,6 +195,12 @@ export const SecurityScanServerToken =
                 logging.log(`Security scan failed. ${error}`)
                 securityScanTelemetryEntry.result = 'Failed'
                 const err = getErrorMessage(error)
+                const authFollowType = getAuthFollowUpType(error)
+                if (authFollowType == 're-auth') {
+                    throw new ResponseError(LSPErrorCodes.RequestFailed, err, {
```
shruti0085, line 200: this would throw an error back to the toolkit and not hit L205-207 which we want to send back incase of failures.
I would recommend changing the return payload to include the responseError object you created.
```
var exception = getException(error) // this sends back the response error in case or re-auth or the base `error` as is
  return {
                    status: 'Failed',
                    error: errorrMessage,
                    exception: exception
                } as SecurityScanResponse
```

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/errors.ts
```diff
@@ -63,3 +63,10 @@ export class AmazonQServiceNoProfileSupportError extends AmazonQError {
         this.name = 'AmazonQServiceNoProfileSupportError'
     }
 }
+
+export class AmazonQServiceAuthenticationExpiredError extends AmazonQError {
+    constructor(message: string = 'Current authentication token is expired.') {
```
shruti0085, line 68: What do you think about `E_AMAZON_Q_INVALID_CONNECTION`? to match L27
A similar change for the class name as well `AmazonQServiceInvalidConnectionError`

server/aws-lsp-codewhisperer/src/language-server/securityScan/codeWhispererSecurityScanServer.ts
shruti0085: Also any reason we cannot create the AmazonQServiceAuthenticationExpired exception here instead of a new Response Error? This will allow us to possibly create a reusable utility call that creates this exception and use it across various features i.e in this class and security scan server

server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
```diff
@@ -480,6 +491,10 @@ export const CodewhispererServerFactory =
                                 throw error
                             }
 
+                            const authFollowType = getAuthFollowUpType(error)
+                            if (authFollowType == 're-auth') {
+                                throw new AmazonQServiceAuthenticationExpiredError(getErrorMessage(error))
+                            }
```
shruti0085, line 497: As mentioned in the previous comment, let's extract this logic into a  conditional check
```
if (hasConnectionExpired(error))
{
   throw new AmazonQServiceAuthenticationExpiredError(getErrorMessage(error))
}
```

LokeshDogga13: Updated

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/errors.ts
```diff
@@ -65,8 +65,10 @@ export class AmazonQServiceNoProfileSupportError extends AmazonQError {
 }
 
 export class AmazonQServiceInvalidConnectionError extends AmazonQError {
+    override message: string
     constructor(message: string = 'Current authentication token is expired.') {
         super(message, 'E_AMAZON_Q_INVALID_CONNECTION')
         this.name = 'AmazonQServiceInvalidConnectionError'
+        this.message = 'Current authentication token is expired.'
```
shruti0085, line 72: why is this override required? The current one's super already passes it in isn't it?

LokeshDogga13: When we receive this object on our end, the value of property message is null. So, I had to override this to get the value.

server/aws-lsp-codewhisperer/src/language-server/securityScan/codeWhispererSecurityScanServer.ts
```diff
@@ -195,15 +196,14 @@ export const SecurityScanServerToken =
                 logging.log(`Security scan failed. ${error}`)
                 securityScanTelemetryEntry.result = 'Failed'
                 const err = getErrorMessage(error)
-                const authFollowType = getAuthFollowUpType(error)
-                if (authFollowType == 're-auth') {
-                    throw new ResponseError(LSPErrorCodes.RequestFailed, err, {
-                        awsErrorCode: 'E_AMAZON_Q_AUTHENTICATION_EXPIRED',
-                    })
-                }
+                const exception = hasConnectionExpired(error as Error)
```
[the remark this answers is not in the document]
LokeshDogga13: I had to do this cast because hasConnectionExpired expects an Error, and this type of error here is unknown

server/aws-lsp-codewhisperer/src/language-server/securityScan/codeWhispererSecurityScanServer.ts
```diff
@@ -195,15 +196,14 @@ export const SecurityScanServerToken =
...
+                    ? new AmazonQServiceInvalidConnectionError(err)
+                    : null
```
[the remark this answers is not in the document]
LokeshDogga13: Updated

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/errors.ts
LokeshDogga13: Removed this variable, as we are already getting it from error message.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/errors.ts
```diff
@@ -1,10 +1,12 @@
 // Base error class for Amazon Q
 export class AmazonQError extends Error {
     public code: string
+    override message: string
```
shruti0085: can this be made public instead of an override?

server/aws-lsp-codewhisperer/src/language-server/securityScan/types.ts
```diff
@@ -60,7 +60,7 @@ export interface SecurityScanRequestParams extends ExecuteCommandParams {
 export interface SecurityScanResponse {
     status: SecurityScanStatus
     findings?: SecurityScanFindings
-    error?: string
+    errorMessage?: string
```
shruti0085, line 63: Please ensure to make a corresponding change client side as well

shruti0085 (approved):

server/aws-lsp-codewhisperer/src/language-server/securityScan/types.ts
LokeshDogga13: Yes, updated there as well - https://github.com/aws/aws-toolkit-visual-studio-staging/pull/2452/files#:~:text=public%20string-,ErrorMessage,-%7B%20get%3B

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/errors.ts
LokeshDogga13: no, this conflicts with base class message property.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/errors.ts
```diff
@@ -63,3 +63,10 @@ export class AmazonQServiceNoProfileSupportError extends AmazonQError {
         this.name = 'AmazonQServiceNoProfileSupportError'
     }
 }
+
+export class AmazonQServiceInvalidConnectionError extends AmazonQError {
+    constructor(message: string = 'Current authentication token is expired.') {
+        super(message, 'E_AMAZON_Q_INVALID_CONNECTION')
```
viktorsaws, line 69: nit: I'd call this "E_AMAZON_Q_CONNECTION_EXPIRED", "invalid" connection can be too generic

LokeshDogga13: @viktorsaws Modified it to "E_AMAZON_Q_CONNECTION_EXPIRED"

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/errors.ts
LokeshDogga13: Yes, added. Thanks !

viktorsaws (approved):

PR #1010, opened 2025-04-19
fix: history not persisted for agentic chat via IdC signin

volodkevych:
## Problem

Currently, history is not persisted if I use agentic chat and sign in via IdC, but work if I sign in via BuilderId.
Apparently, Q API does not return `conversationId` anymore in response (reason to be clarified).

## Solution

Given a lot of chat functionality, including history, depends on `conversationId` present, we set it to random UUID. We set it for agentic chat only, base chat based on `SendMessage` continues working as before.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -192,13 +192,18 @@ export class AgenticChatController implements ChatHandlers {
 
         const triggerContext = await this.#getTriggerContext(params, metric)
         const isNewConversation = !session.conversationId
+        if (isNewConversation) {
+            // agentic chat does not support conversationId in API response,
+            // so we set it to random UUID per session, as other chat functionality
+            // depends on it
+            session.conversationId = uuid()
+        }
 
         token.onCancellationRequested(() => {
             this.#log('cancellation requested')
             session.abortRequest()
         })
 
-        const conversationIdentifier = session?.conversationId ?? 'New conversation'
         const chatResultStream = this.#getChatResultStream(params.partialResultToken)
         try {
             const additionalContext = await this.#additionalContextProvider.getAdditionalContext(
@@ -223,7 +228,7 @@ export class AgenticChatController implements ChatHandlers {
                 session,
                 metric,
                 chatResultStream,
-                conversationIdentifier,
+                session.conversationId,
                 token,
                 triggerContext.documentReference
             )


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -192,13 +192,18 @@ export class AgenticChatController implements ChatHandlers {
 
         const triggerContext = await this.#getTriggerContext(params, metric)
         const isNewConversation = !session.conversationId
+        if (isNewConversation) {
```
aditya169, line 195: Thanks so much for the quick fix! When you have a moment, could you please add a few notes in the PR about how you tested this? It would be super helpful for future reference. Thanks again!

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -192,13 +192,18 @@ export class AgenticChatController implements ChatHandlers {
...
+            // agentic chat does not support conversationId in API response,
+            // so we set it to random UUID per session, as other chat functionality
+            // depends on it
+            session.conversationId = uuid()
```
aditya169, line 199: Thanks so much for the quick fix! When you have a moment, could you please add a few notes in the PR about how you tested this? It would be super helpful for future reference. Thanks again!

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -223,7 +228,7 @@ export class AgenticChatController implements ChatHandlers {
                 session,
                 metric,
                 chatResultStream,
-                conversationIdentifier,
+                session.conversationId,
```
kmile, line 231: If there already is history in the `#chatHistoryDb`, does it matter that the `conversationId`s don't align? I don't _think_ it does, but it seems like we would end up with tabs that have messages from multiple conversations in them.

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
volodkevych: With current implementation, there will be `conversationId` per tab as session is per tab and I create new conversation id only id it's not been set before.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
volodkevych: I reproduced and tested it with [Flare vscode test client](https://github.com/aws/language-servers/tree/main/client/vscode). [CONTRIBUTING.MD](https://github.com/aws/language-servers/blob/main/CONTRIBUTING.md#running--debugging) contains guide about how to start.
The client currently has issue integrating with Q developer profiles (something to be fixed properly as a followup), but I needed IdC signin, so I set `this.enableDeveloperProfileSupport = false` in my codebase to disable that feature and it worked.

PR #1011, opened 2025-04-19
fix: add history to request on each chat prompt

kmile:
## Problem
Chat history was not included on every request unless the history was just rehydrated before the first prompt.

## Solution
Add the chat history from the chat database on the initial request for each prompt.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -265,14 +265,10 @@ export class AgenticChatController implements ChatHandlers {
             ChatTriggerType.MANUAL,
             this.#customizationArn,
             profileArn,
+            this.#chatHistoryDb.getMessages(params.tabId, 10),
             this.#features.agent.getTools({ format: 'bedrock' }),
             additionalContext
         )
-
-        if (!session.localHistoryHydrated && requestInput.conversationState) {
-            requestInput.conversationState.history = this.#chatHistoryDb.getMessages(params.tabId, 10)
-            session.localHistoryHydrated = true
-        }
 
         return requestInput
     }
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -265,15 +265,11 @@ export class AgenticChatController implements ChatHandlers {
             ChatTriggerType.MANUAL,
             this.#customizationArn,
             profileArn,
+            this.#chatHistoryDb.getMessages(params.tabId, 10),
             this.#features.agent.getTools({ format: 'bedrock' }),
             additionalContext
         )
 
-        if (!session.localHistoryHydrated && requestInput.conversationState) {
```
jpinkney-aws, line 272: qq: what does hydration do and why do we need to remove it?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -321,11 +317,6 @@ export class AgenticChatController implements ChatHandlers {
                 documentReference
             )
 
-            // Store the conversation ID from the first response
-            if (iterationCount === 1 && result.data?.conversationId) {
```
jpinkney-aws, line 325: another qq: why did this become irrelevant?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: This was only used to check whether we finished loading history the when we initialized the session. Since we need the history to be included on every prompt, we shouldn't only do this the first time.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: ConversationId is no longer provided by the backend. It is created by the client when we initialize the session.

jpinkney-aws (approved):

PR #1012, opened 2025-04-20
chore: update sample vscode extension with error handling for chat errors

ege0zcan:
## Problem

In the current setup, chat-client needs to receive a chat message that does not have the `isPartialResult` field in order to understand that an answer stream has reached completion. If an answer stream that gets started does not get the final message, the input prompt will be stuck in disabled state, rendering the tab unusable

### Case 1
When max number of tool iterations (currently 10) was reached and a response had not yet been finalized because there were still more tool uses waiting to happen, the controller would return the chat response as an empty object with success field set to false in `runAgentLoop` and when this result was passed to `handleFinalResult` it would see `success` is false and return a ResponseError. This resulted in the chat being stuck. On top of it, the message wouldn't be included in the chat history because history logic is part of `handleFinalResult` function that gets executed after the ResponseError is returned

To fix this behaviour, added logic to the `runAgentLoop` function that now checks if the max iterations has been completed. If yes, the partial result up until that point is marked as the final result and an extra warning message is added to the message that gets shown to the user. No more ResponseError is being thrown to the client so this case is now fully handled in the server 

### Testing

This can be easily tested with a prompt like the following which is guarenteed to trigger the read tool more than 10 times:

`Please perform a read operation on the following paths, they are very critical for the task that we will work on today: 

<list of 15 paths>
`

See sample screenshot:

### Case 2
When there is an API error (e.g something goes wrong with the backend), the `handleRequestError` gets triggered and if a message stream was already started, the message stream should be closed. The sample vscode extension that is used in this repo didn't handle response errors and this meant that the final result would never get passed to the chat-client, leaving chat in the stuck state explained above. Added error handling to the extension by wrapping the 
`languageClient.sendRequest(chatRequestType` function in a try catch block. Added some extra logic to the client so that if an error happens but if there was some partial progress that had already been streamed, the extension will preserve the existing message and add an extra line of errror warning to the message shown to the end user

Note that messages which have faced some sort of an error are not included in the chat history database, the implications of this should be considered.

### Testing
To make testing simpler I hardcoded `handleFinalResult` to return a ResponseError and verified that the client would keep the existing messages and add an extra line of warning to the message. See the sample screenshot:

## License

### DIFF
--- a/client/vscode/src/chatActivation.ts
+++ b/client/vscode/src/chatActivation.ts
@@ -41,7 +41,6 @@ import {
 } from 'vscode-languageclient/node'
 import * as jose from 'jose'
 import * as vscode from 'vscode'
-import * as fs from 'fs'
 
 export function registerChat(languageClient: LanguageClient, extensionUri: Uri, encryptionKey?: Buffer) {
     const webviewInitialized: Promise<Webview> = new Promise(resolveWebview => {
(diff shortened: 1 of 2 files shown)


client/vscode/src/chatActivation.ts
```diff
@@ -80,42 +79,73 @@ export function registerChat(languageClient: LanguageClient, extensionUri: Uri,
...
+                                            message.params.tabId
+                                        )
+                                    }
+                                )
+
+                                const editor =
+                                    window.activeTextEditor ||
+                                    window.visibleTextEditors.find(editor => editor.document.languageId != 'Log')
+                                if (editor) {
+                                    message.params.cursorPosition = [editor.selection.active]
```
nkomonen-amazon, line 112: Does `cursorPosition` exist, should it be `cursorState`? I think this is of the type `ChatParams` based on the current implementation in aws-toolkit-vscode

ege0zcan: This part is not necessary at all actually, LLM managed to sneak this in. I will remove this part

ege0zcan:
> After [porting your changes](https://github.com/aws/aws-toolkit-vscode/pull/7121) + testing in VSC I got to around 60 ToolUseEvents and finally got the error: `Error occurred during chat request: Error: Improperly formed request.`
> 
> Is this expected, does `Improperly formed request` imply throttling of some sort? If the service does not return an appropriate error message when it is throttled, maybe we need to wrap this?
> 
> For context IIUC we bumped the agentic loop limit to 100, so testing for 10 is not sufficient anymore: #1022

I believe this error is happening due to some other reasons like the input message reaching some size limit. We need to apply extra handling to the input message on the controller side so that we truncate the message if messages start growing over the limits. https://github.com/aws/language-servers/pull/1050 should fix this on the controller side

PR #1013, opened 2025-04-20
feat(amazonq): add pair programming toggle

jpinkney-aws:
## Problem
Missing pair programming toggle

## Solution
Add pair programming toggle on the bottom left of chat

Depends on https://github.com/aws/language-server-runtimes/pull/449

Screenshot 2025-04-19 at 8 30 04 PM

## License


kmile (approved):

Hweinstock (approved):
Did we confirm the heuristic used by the backend for checking if pair programming is on? I remember @kmile was able to trick it depending on which tools were included, which seems like a very fragile check.

kmile:
The pair programming mode actually only (mostly?) influences which tools we run. The model doesn't know or care if the toggle is on or off, we don't send it to the backend.

PR #1015, opened 2025-04-20
fix(amazonq): fix context transparency doesn't show for file & file c…

zixlin7:
…lick

## Problem
- fileClick pass in absolute path and handle prompt file outside of workspace
- fix an issue when no file open the context transparency list in chat response doesn't show because the workspace is not in triggerContext, we should utilize the workspaceUtil to get the current workspace
- fix typing for context command

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -3,6 +3,7 @@
  * Will be deleted or merged.
  */
 
+import * as path from 'path'
 import {
     ChatTriggerType,
     GenerateAssistantResponseCommandInput,
@@ -82,9 +83,10 @@ import {
     TriggerContext,
 } from './context/agenticChatTriggerContext'
 import { AdditionalContextProvider } from './context/addtionalContextProvider'
-import { getNewPromptFilePath } from './context/contextUtils'
+import { getNewPromptFilePath, getUserPromptsDirectory, promptFileExtension } from './context/contextUtils'
 import { ContextCommandsProvider } from './context/contextCommandsProvider'
 import { LocalProjectContextController } from '../../shared/localProjectContextController'
+import { workspaceUtils } from '@aws/lsp-core'
 
 type ChatHandlers = Omit<
     LspHandlers<Chat>,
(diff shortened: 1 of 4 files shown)


jpinkney-aws (commented):
I see a lot of these are just using the first workspace folder, are there any negative implications with that?

zixlin7:
> I see a lot of these are just using the first workspace folder, are there any negative implications with that?

This is the same implementation on VS Code side to use the first workspace, but I will followup and do some more testing see how this plays with multi-folder workspace

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -721,7 +723,16 @@ export class AgenticChatController implements ChatHandlers {
 
     async onFileClicked(params: FileClickParams) {
         // TODO: also pass in selection and handle on client side
-        await this.#features.lsp.window.showDocument({ uri: params.filePath })
+        const workspaceRoot = workspaceUtils.getWorkspaceFolderPaths(this.#features.lsp)[0]
```
kmile, line 726: There technically could be more than one workspace folder.

kmile (approved):

jpinkney-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
volodkevych: +1, we pass both absolute and relative path to MynahUI when we define context commands, the fact that MynahUI exposes only relative path [here](https://github.com/aws/language-servers/blob/main/chat-client/src/client/mynahUi.ts#L196) needs to be fixed @dogusata @Jurredr

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/addtionalContextProvider.ts
```diff
@@ -135,24 +140,16 @@ export class AdditionalContextProvider {
...
-            let itemType: ContextCommandItemType | undefined
-            if (item.icon === 'file') {
-                itemType = 'file'
-            } else if (item.icon === 'folder') {
-                itemType = 'folder'
-            } else if (item.icon === 'code-block') {
-                itemType = 'code'
-            }
-            if (itemType) {
+            if (item.route && item.route.length === 2) {
```
volodkevych, line 146: Is there any reason we skip items that do not have route? What about supporting @sage and @workspace?

zixlin7: yes this is intentional, for @workspace we don't need to call the local context library to get additional prompt, we should be able to just store the relevant document when we query them and map that directly to the context list

PR #1017, opened 2025-04-20
fix: fallback to fs if document context fails to sync

Hweinstock:
## Problem
If the document can't be found in the active documents, then it fails to populate it in the request to the server. This results in confusing behavior where some files may fail to fetch here: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts#L84. 

## Solution
- add a fallback mechanism such that if we can't find the document through the LSP, we use the underlying file system to read the file instead. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
@@ -120,21 +120,29 @@ export class AgenticChatTriggerContext {
         }
         const textDocument = await this.getTextDocument(textDocumentIdentifier.uri)
 
-        return this.#documentContextExtractor.extractDocumentContext(
-            textDocument,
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
```diff
@@ -126,6 +130,28 @@ export class AgenticChatTriggerContext {
...
     }
 
+    /**
+     * Fetch the current textDocument such that:
+     * 1. If the document is synced with LSP, return the synced textDocument
+     * 2. If the document is not synced with LSP, read the file from the file system
+     * 3. If the file cannot be read, return undefined
+     * @param uri
+     * @returns
+     */
```
kmile, line 140: This is great, we'd get a more up-to-date representation of open files!

kmile (approved):

PR #1020, opened 2025-04-20
feat: add header and buttons to chat response

ctlai95:
## Problem

New fields added need to be passed to mynah-ui

## Solution

- bump runtimes version
- add utils to convert runtime types to mynah types
- pass `header` and `buttons` to mynah-ui

## License

### DIFF
--- a/chat-client/package.json
+++ b/chat-client/package.json
@@ -23,7 +23,7 @@
     },
     "dependencies": {
         "@aws/chat-client-ui-types": "^0.1.16",
-        "@aws/language-server-runtimes-types": "^0.1.14",
+        "@aws/language-server-runtimes-types": "^0.1.19",
         "@aws/mynah-ui": "file:./lib/aws-mynah-ui-4.31.0-beta.6.tgz"
     },
     "bundleDependencies": [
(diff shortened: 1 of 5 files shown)


kmile (approved):

chat-client/src/client/mynahUi.ts
```diff
@@ -460,7 +461,7 @@ export const createMynahUi = (
 
         if (isPartialResult) {
             // type for MynahUI differs from ChatResult types so we ignore it
-            mynahUi.updateLastChatAnswer(tabId, { ...chatResultWithoutType, header: header })
+            mynahUi.updateLastChatAnswer(tabId, { ...chatResultWithoutType, header: header, buttons: buttons })
```
zixlin7, line 464: Do we still need this if chatMessage already has header and buttons?

chat-client/package.json
```diff
@@ -23,7 +23,7 @@
     },
     "dependencies": {
         "@aws/chat-client-ui-types": "^0.1.16",
-        "@aws/language-server-runtimes-types": "^0.1.14",
+        "@aws/language-server-runtimes-types": "^0.1.19",
```
zixlin7, line 26: Should we bump "@aws/language-server-runtimes" too?

chat-client/src/client/mynahUi.ts
ctlai95: yeah the types are slightly different, I pulled them out and converted them above

zixlin7 (approved):

PR #1024, opened 2025-04-20
feat(amazonq): initial implementation of read/list chat result

zixlin7:
## Problem
right now we are outputing fsRead/listDir as json to chat stream
## Solution
- Initial implementation of processing fsRead listDir to chat result stream as contextList

Next steps are:
- handle the usecase when listDir/fsRead requires user confirmation
- update to support multiple messages

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -432,6 +445,50 @@ export class AgenticChatController implements ChatHandlers {
         return results
     }
 
+    #processReadOrList(toolUse: ToolUse, chatResultStream: AgenticChatResultStream): ChatResult | undefined {
+        // return initial message about fsRead or listDir
+        const toolUseId = toolUse.toolUseId!
+        const currentPath = (toolUse.input as any)?.path
```
kmile, line 451: I wonder if we can get the actual type from the tool definition here. We might be able to add it to the `Tool` classes so we don't have to `any` this.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -432,6 +445,50 @@ export class AgenticChatController implements ChatHandlers {
...
+            }
+        }
+
+        const contextList: FileList = {
+            rootFolderTitle: title,
+            filePaths: currentFileList.map(item => item.relativeFilePath),
+            details: fileDetails,
+        }
+
+        // TODO: handle confirmation use case
```
kmile, line 486: Not necessary for now.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
```diff
@@ -10,11 +10,16 @@ interface ResultStreamWriter {
...
+
+interface FileDetailsWithPath extends FileDetails {
+    relativeFilePath: string
+}
 export class AgenticChatResultStream {
     static readonly resultDelimiter = '\n\n'
     #state = {
         chatResultBlocks: [] as ChatResult[],
         isLocked: false,
+        fileList: {} as Record<string, FileDetailsWithPath[]>,
```
kmile, line 22: Are we creating 1 big list of all files and folders read? Or do we need to keep track of them separately?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -394,9 +395,17 @@ export class AgenticChatController implements ChatHandlers {
 
         for (const toolUse of toolUses) {
             if (!toolUse.name || !toolUse.toolUseId) continue
+            if (toolUse.name === 'fsRead' || toolUse.name === 'listDirectory') {
```
ctlai95, line 398: can we make this an if/else with line 406?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
```diff
@@ -26,11 +31,33 @@ export class AgenticChatResultStream {
...
+
+        // TODO: the combining of contextList is temporary just so we can see them
+        // they should be different messages once we support multiple messages
         return chatResults.reduceRight((acc, c) => ({
             ...acc,
             body: c.body + AgenticChatResultStream.resultDelimiter + acc.body,
+            ...((c.contextList || acc.contextList) && {
+                contextList: {
+                    filePaths: [...(acc.contextList?.filePaths ?? []), ...(c.contextList?.filePaths ?? [])],
+                    rootFolderTitle: c.contextList?.rootFolderTitle
```
kmile, line 56: This would flip from files/folders or vice versa if we have both in the same response.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
ctlai95, line 451: I'm running into this too, should we try to cast the known param type?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
kmile: Maybe this should be `contextList` since we are going to have a `fileList` for `fsWrite`.

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
zixlin7: I was thinking this would be a list for each toolUseId, open to suggestions

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
kmile: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts#L8

Is this one of the input types?

https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolServer.ts#L16 is supposed to break if the input type and the invoke parameter type don't match, so it should have type safety with the JSON provided by the model (assuming we're run the tool already)

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
zixlin7: yes they should not be combined, each tooluseID should have its own message?

ctlai95 (approved):

kmile (approved):

PR #1026, opened 2025-04-21
feat: initial fsWrite chat message

ctlai95:
## Problem

Initial FsWrite tool message implementation

image

## Solution
- Send header params from controller
- Update mynahUi to handle tool specific messages. Things like fullWidth and padding do not belong in the flare library, so we need custom logic to add them in the UI.

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -461,7 +461,24 @@ export const createMynahUi = (
 
         if (isPartialResult) {
             // type for MynahUI differs from ChatResult types so we ignore it
-            mynahUi.updateLastChatAnswer(tabId, { ...chatResultWithoutType, header: header, buttons: buttons })
+            mynahUi.updateLastChatAnswer(tabId, {
+                ...chatResultWithoutType,
+                header: header,
+                buttons: buttons,
+                // TODO: this won't be needed once we have multiple message support
+                ...(type === 'tool'
+                    ? {
+                          header: {
+                              ...header,
+                              // TODO: this is specific to fsWrite tool, modify for other tools as needed
+                              fileList: { ...header?.fileList, fileTreeTitle: '', hideFileCount: true },
+                              buttons: header?.buttons?.map(button => ({ ...button, status: 'clear' })),
+                          },
+                          fullWidth: true,
+                          padding: false,
+                      }
+                    : {}),
+            })
             return
         }
 
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -94,6 +94,7 @@ import { LocalProjectContextController } from '../../shared/localProjectContextC
 import { workspaceUtils } from '@aws/lsp-core'
 import { FsReadParams } from './tools/fsRead'
 import { ListDirectoryParams } from './tools/listDirectory'
+import { FsWrite, FsWriteParams } from './tools/fsWrite'
 
 type ChatHandlers = Omit<
     LspHandlers<Chat>,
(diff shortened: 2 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -425,9 +426,18 @@ export class AgenticChatController implements ChatHandlers {
                     content: [toolResultContent],
                 })
 
-                // no need to write tool result for listDir and fsRead into chat stream
-                if (toolUse.name !== 'fsRead' && toolUse.name !== 'listDirectory') {
-                    await chatResultStream.writeResultBlock({ body: toolResultMessage(toolUse, result) })
+                switch (toolUse.name) {
+                    case 'fsRead':
+                    case 'listDirectory':
+                        // no need to write tool result for listDir and fsRead into chat stream
```
laileni-aws, line 432: nit: Can I know the reason, Why ?

laileni-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
ctlai95: we are sending it before running the tool above, but actually it might make more sense to move it here

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -446,6 +456,36 @@ export class AgenticChatController implements ChatHandlers {
         return results
     }
 
+    async #getFsWriteChatResult(toolUse: ToolUse): Promise<ChatResult> {
+        const input = toolUse.input as unknown as FsWriteParams
+        const fileName = path.basename(input.path)
+        // TODO: right now diff changes is coupled with fsWrite class, we should move it to shared utils
+        const fsWrite = new FsWrite(this.#features)
+        const diffChanges = await fsWrite.getDiffChanges(input)
```
kmile, line 464: This should be a static, but even better would be in a shared util like you mentioned so the representation is not coupled with the tool itself.

chat-client/src/client/mynahUi.ts
```diff
@@ -461,7 +461,24 @@ export const createMynahUi = (
...
+                ...chatResultWithoutType,
+                header: header,
+                buttons: buttons,
+                // TODO: this won't be needed once we have multiple message support
+                ...(type === 'tool'
+                    ? {
+                          header: {
+                              ...header,
+                              // TODO: this is specific to fsWrite tool, modify for other tools as needed
+                              fileList: { ...header?.fileList, fileTreeTitle: '', hideFileCount: true },
```
jpinkney-aws, line 474: is only this line going to be modified for other tools or all of it? If so, just wondering if it makes sense to pull this out to a different file so that its easy to understand which tools have which interface

Hweinstock: +1 to separating this logic out. Any tool specific logic should be isolated from the core logic.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -425,9 +426,18 @@ export class AgenticChatController implements ChatHandlers {
                     content: [toolResultContent],
                 })
 
-                // no need to write tool result for listDir and fsRead into chat stream
-                if (toolUse.name !== 'fsRead' && toolUse.name !== 'listDirectory') {
-                    await chatResultStream.writeResultBlock({ body: toolResultMessage(toolUse, result) })
+                switch (toolUse.name) {
```
Hweinstock, line 429: For the sake of future MCP support, I think this tool specific branching and logic should be isolated from the core logic.

kmile (approved):
I agree with the suggestions to separating out the tool specific logic from the main controller, but that can be done as a follow-up once we have all the tools in there.

PR #1029, opened 2025-04-21
feat: add @workspace context in agentic chat

volodkevych:
## Problem

`@workspace` is not supported in agentic chat context commands now.

## Solution

The support for `@workspace` was added. It relies on local workspace indexing working.

TODO in followup PRs: prompt customer to enable indexing if it's disabled.

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
```diff
@@ -125,11 +129,25 @@ export class ContextCommandsProvider implements Disposable {
             description: 'Add a saved prompt to context',
             icon: 'magic',
         }
-        const allCommands = [
+        const commands = [
+            ...(localProjectContextEnabled
```
volodkevych, line 133: I'll probably remove this check with the following PRs, prompting customer to enable the indexing lib. For now, I think it's safer to keep the check here.

PR #1035, opened 2025-04-22
feat: implement stop response

jpinkney-aws:
## Problem
agentic chat has stop response disabled

## Solution
enable it and make it so that when its clicked it returns back an empty answer, which completes the request stream and unlocks the UI

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -264,6 +264,17 @@ export class AgenticChatController implements ChatHandlers {
                 chatResultStream
             )
         } catch (err) {
+            if (token?.isCancellationRequested) {
+                /**
+                 * when the session is aborted it generates an error.
+                 * we need to resolve this error with an answer so the
+                 * stream stops
+                 */
+                return {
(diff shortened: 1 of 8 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -264,6 +264,17 @@ export class AgenticChatController implements ChatHandlers {
...
         } catch (err) {
+            if (token?.isCancellationRequested) {
+                /**
+                 * when the session is aborted it generates an error.
+                 * we need to resolve this error with an answer so the
+                 * stream stops
+                 */
+                return {
+                    type: 'answer',
+                    body: '',
```
kmile, line 275: Should we render what we have? Show a "this request was canceled" text?

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
jpinkney-aws: whatever's been rendered will already be sent to chat

> this request was canceled

I was thinking that as well, but kept it with what we currently do in the agentic chat branch. I think its a pretty quick two way door

PR #1036, opened 2025-04-22
fix: remove thumbs up/down button

Randall-Jiang:
## Problem
remove the thumb up/down button on the answer card
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
@@ -59,7 +59,7 @@ describe('ChatController', () => {
     const expectedCompleteChatResult: ChatResult = {
         messageId: mockMessageId,
         body: 'Hello World!',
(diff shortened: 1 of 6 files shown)


jpinkney-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatController.test.ts
```diff
@@ -59,7 +59,7 @@ describe('ChatController', () => {
     const expectedCompleteChatResult: ChatResult = {
         messageId: mockMessageId,
         body: 'Hello World!',
-        canBeVoted: true,
+        canBeVoted: false,
```
jpinkney-aws, line 62: actually you probably don't need to change the normal chat controller, you only need to change the agentic chat controller ones

volodkevych: yes, the change should not affect the base chat

PR #1039, opened 2025-04-22
fix: improve chat rendering if there are additional chat messages

kmile:
## License


chat-client/src/client/mynahUi.ts
```diff
@@ -569,25 +576,7 @@ export const createMynahUi = (
...
-            canBeVoted: undefined,
-            codeReference: undefined,
-            fileList: undefined,
-        })
-
-        mynahUi.endMessageStream(tabId, chatResult.messageId ?? '')
-
-        mynahUi.updateLastChatAnswer(tabId, {
-            type: ChatItemType.ANSWER,
+        mynahUi.endMessageStream(tabId, chatResult.messageId ?? '', {
```
jpinkney-aws, line 579: does this have the potential to break history if we aren't finishing with an answer card?

jpinkney-aws (approved):

chat-client/src/client/mynahUi.ts
kmile: This doesn't have to be the last card, just the one specified.

And in the backend, we only put `assistantResponse` items in the `chatResult`, the tools go in the `additionalMessages`.

kmile: History is done on the server and yes, it currently breaks if we have lingering `toolUses` until we port over https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/shared/db/chatDb/chatDb.ts#L311

viktorsaws (approved):

viktorsaws (approved):

PR #1045, opened 2025-04-22
feat: added support for injecting additional context commands

anasstahr:
## Problem
Currently we don't have a way to support additional context commands from the client side.

## Solution
Adding support for this

## License

### DIFF
--- a/chat-client/src/client/chat.ts
+++ b/chat-client/src/client/chat.ts
@@ -18,6 +18,7 @@ import {
     CopyCodeToClipboardParams,
     ERROR_MESSAGE,
     ErrorMessage,
+    FeatureContext,
     GENERIC_COMMAND,
     GenericCommandMessage,
     INSERT_TO_CURSOR_POSITION,
@@ -99,7 +100,8 @@ type ChatClientConfig = Pick<MynahUIDataModel, 'quickActionCommands'> & {
 export const createChat = (
     clientApi: { postMessage: (msg: UiMessage | UiResultMessage | ServerMessage) => void },
     config?: ChatClientConfig,
-    chatClientAdapter?: ChatClientAdapter
+    chatClientAdapter?: ChatClientAdapter,
+    featureConfigSerialized?: string
 ) => {
     let mynahApi: InboundChatApi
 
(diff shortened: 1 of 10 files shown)


client/vscode/src/chatActivation.ts
```diff
@@ -404,7 +404,11 @@ function generateJS(webView: Webview, extensionUri: Uri): string {
     <script type="text/javascript" src="${entrypoint.toString()}" defer onload="init()"></script>
     <script type="text/javascript">
         const init = () => {
-            amazonQChat.createChat(acquireVsCodeApi(), {disclaimerAcknowledged: false});
+            amazonQChat.createChat(acquireVsCodeApi(), 
+                {disclaimerAcknowledged: false}, 
+                undefined,
+                undefined,
```
volodkevych, line 410: can we have example to show `@sage` injected here please?

chat-client/src/client/mynahUi.ts
```diff
@@ -200,10 +201,11 @@ export const createMynahUi = (
         },
         onTabAdd: (tabId: string) => {
             const defaultTabBarData = tabFactory.getDefaultTabData()
+            const contextCommands = [...(contextCommandGroups || []), ...(featureConfig?.get('contextCommands') || [])]
```
volodkevych, line 204: Why are we doing here some thing different in comparison to what was done above? Can we share the logic? Why is 'contextCommands' key used here, but not in the logic above?

chat-client/src/client/chat.ts
```diff
@@ -183,8 +195,25 @@ export const createChat = (
                 }
 
                 const allExistingTabs: MynahUITabStoreModel = mynahUi.getAllTabs()
+
+                const additionalContextCommands: any[] = []
+                featureConfig.forEach((featureContext: FeatureContext, _feature: string) => {
+                    additionalContextCommands.push({
```
volodkevych, line 201: I do not think it's safe to use all the items from featureConfig as additional context commands. Can we just pull that 'highlightCommand' here, which is used for @sage?

chat-client/src/client/mynahUi.ts
```diff
@@ -692,9 +707,12 @@ ${params.message}`,
             commands: toContextCommands(group.commands),
         }))
 
+        console.log('Context commands feature ', featureConfig?.get('highlightCommands'))
+        const contextCommands = [...(contextCommandGroups || []), ...(featureConfig?.get('highlightCommands') || [])]
+
         Object.keys(mynahUi.getAllTabs()).forEach(tabId => {
             mynahUi.updateStore(tabId, {
-                contextCommands: contextCommandGroups,
+                contextCommands: contextCommands,
```
volodkevych, line 715: This will not work: `contextCommands` do not have `variation` and `value.string..` properties.

volodkevych (approved):

volodkevych (approved):

PR #1052, opened 2025-04-22
feat(amazonq): implement shared q service server

patrickvbeurden:
## Problem

The `AmazonQServiceManager` has evolved into a service / server hybrid, both vending Q functionality to all dependent servers and handling LSP events. Currently, all servers try to initialize the service manager and have to setup duplicate lsp listeners because there are no guarantees on which servers end up in the final bundle or actually initializes the service manager, which is not an issue today but does not scale well.

## Solution

Introduce a dedicated server for the initialization of the service manager, which is always bundled as a "hard dependency" for all other servers. This way, the service manager also has its own LSP object to attach handlers to, which prevents accidental future overrides. 

If multiple servers try to initialize the service manager, an error is thrown at runtime. If a dependent server tries to use the q functionality while the service manager is not initialized yet, an error is thrown at runtime.

### check list

- [x] E2E basic "smoke test" (validated no errors, expect initialization orders and chat works for token)
- [x] Additional unit tests
- [x] More elaborate E2E testing (IAM and token servers spinning up, for the latter both profile / no profile flows stil work)
- [x] Updates to documentation regarding bundle creation in `app/aws-lsp-codewhisperer-runtimes`

## License


server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -178,4 +150,22 @@ export class ServerConfigurationProvider {
         this.logging.error(`${message}: ${error}`)
         return new ResponseError(LSPErrorCodes.RequestFailed, message)
     }
+
+    private get serviceManager(): AmazonQTokenServiceManager {
+        if (!this.cachedServiceManager) {
+            this.cachedServiceManager = AmazonQTokenServiceManager.getInstance()
```
viktorsaws, line 156: Can't we use `AmazonQServiceAPI` here instead of maintaining cached instance here? I think we should align uses of `AmazonQTokenServiceManager` and `AmazonQServiceAPI` across the codebase, and use either one of them, to avoid confusion.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/BaseAmazonQServiceManager.ts
```diff
@@ -155,13 +193,51 @@ export abstract class BaseAmazonQServiceManager<
...
+    getStreamingClient(): S
+    getConfiguration(): Readonly<AmazonQWorkspaceConfig>
+    addDidChangeConfigurationListener(listener: DidChangeConfigurationListener): Promise<void>
+    removeDidChangeConfigurationListener(listener: DidChangeConfigurationListener): void
+}
 
-        this.logging.debug('BaseAmazonQServiceManager functionality initialized')
+export type AmazonQServiceBase = AmazonQServiceAPI<CodeWhispererServiceBase, StreamingClientServiceBase>
+
+export class AmazonQServiceAPI<C extends CodeWhispererServiceBase, S extends StreamingClientServiceBase>
```
viktorsaws, line 210: I tried to wrap my head around why would we really need this proxy object to access methods of `ServiceManager`, and I feel like it is not really necessary, unless I'm missing very obvious reason.

If the idea is to prevent access to not initialized `ServiceManager` behind this class, I don't see how it really protects against mis-usage in consuming code. Since in all consuming code we need to do something like `const amazonQService = new AmazonQServiceAPI(() => AmazonQTokenServiceManager.getInstance())`, that code could start making calls to `amazonQService` immediately after. 

Assuming that all factory `serviceManagerFactory` implementations passed inside constructor here always call `ServiceManager.getInstance()`, why don't we make `.getInstance()` call a part of every public method of `BaseAmazonQServiceManager` itself? Then we don't need this indirection with passing factories, and consumes can keep using `ServiceManager` implementations directly as before.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/BaseAmazonQServiceManager.ts
```diff
@@ -145,6 +159,30 @@ export abstract class BaseAmazonQServiceManager<
...
+
+        this.features.lsp.didChangeConfiguration(async () => {
+            this.logging.debug('Received didChangeconfiguration event')
+            await this.handleDidChangeConfiguration()
+        })
+
+        this.logging.debug('Attached onInitialized and didChangeConfiguration lsp listeners.')
+    }
+
+    public setupConfigurableLspHandlers(): void {
```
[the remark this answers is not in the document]
patrickvbeurden: Discussed offline, all handlers are now explicitly set in the `AmazonQServiceServerFactory` for improved clarity.

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
patrickvbeurden: Discussed offline, largely reverted the changes and removed `AmazonQServiceAPI` (see comment below)

viktorsaws (approved):

PR #1056, opened 2025-04-22
fix: reject should terminate agentic loop

Hweinstock:
## Problem
The reject button causes the chat to hang since its caught in the agentic loop. 

## Solution
- Create a new `ToolApprovalException` that we bubble up to escape the loop. 
- For other errors, we still feed them back to the model. 
- This is separate from cancellation via the stop button, so I avoided re-using the same language. 

## Verification

https://github.com/user-attachments/assets/3e4fdd94-26ae-49b6-8b19-3a7640900dfb

- 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -516,7 +516,7 @@ export class AgenticChatController implements ChatHandlers {
                 if (needsConfirmation) {
                     const deferred = this.#createDeferred()
                     session.setDeferredToolExecution(toolUse.toolUseId, deferred.resolve, deferred.reject)
-                    this.#log(`Promting for tool approval for tool: ${toolUse.name}`)
+                    this.#log(`Prompting for tool approval for tool: ${toolUse.name}`)
                     await deferred.promise
                 }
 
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -515,13 +516,8 @@ export class AgenticChatController implements ChatHandlers {
...
                     const deferred = this.#createDeferred()
                     session.setDeferredToolExecution(toolUse.toolUseId, deferred.resolve, deferred.reject)
-
-                    // the below line was commented out for now because
-                    // the partial result block from above is not streamed to chat window yet at this point
-                    // so the buttons are not in the window for the promise to be rejected/resolved
-                    // this can to be brought back once intermediate messages are shown
-
-                    // await deferred.promise
+                    this.#log(`Promting for tool approval for tool: ${toolUse.name}`)
```
jpinkney-aws, line 519:
```suggestion
                    this.#log(`Prompting for tool approval for tool: ${toolUse.name}`)
```

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -308,7 +308,8 @@ export class AgenticChatController implements ChatHandlers {
                 chatResultStream
             )
         } catch (err) {
-            if (token?.isCancellationRequested) {
+            // TODO: On ToolValidationException, we want to show custom mynah-ui components making it clear it was cancelled.
+            if (token?.isCancellationRequested || err instanceof ToolApprovalException) {
```
jpinkney-aws, line 312: what UI would this eventually be?

jpinkney-aws (approved):

jpinkney-aws:
Very similar with what i'm doing for https://github.com/aws/language-servers/pull/1057 which is nice. I think bubbling everything back up to that main loop will make things the easiest to handle

zixlin7 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: This is what it looks like on agentic-chat: 
image. 

Just some red text mostly.

jpinkney-aws (approved):

PR #1057, opened 2025-04-22
feat: add cancellation handling to tools

jpinkney-aws:
## Problem
tools can't be canceled

## Solution
- plumb the cancellation token through to the tools
- tools can optionally handle the cancellation token if it makes sense. e.g. execute bash/list directories makes sense, write file and read file doesn't
- when a token is cancelled, it bubbles up the main agentic loop, sending an answer back to the chat-client and unlocking the UI 

## License

### DIFF
--- a/core/aws-lsp-core/src/util/workspaceUtils.ts
+++ b/core/aws-lsp-core/src/util/workspaceUtils.ts
@@ -35,7 +35,7 @@ export async function readDirectoryRecursively(
 
     while (queue.length > 0) {
         if (token?.isCancellationRequested) {
-            features.logging.info('cancelled listing directory')
+            features.logging.info('cancelled readDirectoryRecursively')
             throw new CancellationError('user')
         }
 
--- a/core/aws-lsp-core/src/util/workspaceUtils.ts
+++ b/core/aws-lsp-core/src/util/workspaceUtils.ts
@@ -1,6 +1,8 @@
 import * as path from 'path'
 import { URI } from 'vscode-uri'
 import { Features } from '@aws/language-server-runtimes/server-interface/server'
+import { CancellationToken } from '@aws/language-server-runtimes/server-interface'
+import { CancellationError } from './awsError'
 
 type ElementType<T> = T extends (infer U)[] ? U : never
 type Dirent = ElementType<Awaited<ReturnType<Features['workspace']['fs']['readdir']>>>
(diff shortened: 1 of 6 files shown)


core/aws-lsp-core/src/util/workspaceUtils.ts
```diff
@@ -31,6 +34,11 @@ export async function readDirectoryRecursively(
     const formatter = options?.customFormatCallback ?? formatListing
 
     while (queue.length > 0) {
+        if (token?.isCancellationRequested) {
+            features.logging.info('cancelled listing directory')
```
Hweinstock, line 38: this is meant to be a general utility to use elsewhere (currently not), so may be worth changing `listing directory` here to `readDirectoryRecursively` even though right now they are the same.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -554,6 +554,10 @@ export class AgenticChatController implements ChatHandlers {
                         break
                 }
             } catch (err) {
+                if (CancellationError.isUserCancelled(err)) {
```
Hweinstock, line 557: I handle tool validation rejections very similarly [here](https://github.com/aws/language-servers/pull/1056). Do you think their logic should be combined more? 

I think they should be separate since the stop button and the reject button are different, and could warrant different behavior in the future, but open to changing it.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -315,7 +314,7 @@ export class ExecuteBash {
                 onStderr: async (chunk: string) => {
                     if (cancellationToken?.isCancellationRequested) {
                         this.logging.debug('Bash command execution cancelled during stderr processing')
-                        return
+                        throw new CancellationError('user')
```
Hweinstock, line 317: do we need to kill the process or is this sufficient?

Hweinstock (approved):
Nice!

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
jpinkney-aws: I think we should separate them since their intentions are different. I think they'd just have a minor overlap

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
jpinkney-aws: Just looked into this, it's already cancelled by https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts#L334

Hweinstock (approved):

PR #1062, opened 2025-04-22
feat: update confirm header after button click WIP

zixlin7:
## Problem
- fix issue that json output is still shown due to missed `break` in prev rebase
- update the header buttons for user confirmation 
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -521,6 +521,7 @@ export class AgenticChatController implements ChatHandlers {
                     session.setDeferredToolExecution(toolUse.toolUseId, deferred.resolve, deferred.reject)
                     this.#log(`Prompting for tool approval for tool: ${toolUse.name}`)
                     await deferred.promise
+                    await chatResultStream.writeResultBlock(this.#getUpdateBashConfirmResult(toolUse, true))
                 }
 
                 const result = await this.#features.agent.runTool(toolUse.name, toolUse.input)
@@ -552,6 +553,7 @@ export class AgenticChatController implements ChatHandlers {
                     case 'executeBash':
                         const bashToolResult = this.#getBashExecutionChatResult(toolUse, result)
                         await chatResultStream.writeResultBlock(bashToolResult)
+                        break
                     default:
                         await chatResultStream.writeResultBlock({
                             type: 'tool',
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -561,6 +563,11 @@ export class AgenticChatController implements ChatHandlers {
                         break
                 }
             } catch (err) {
+                // If we did not approve a tool to be used, bubble this up to interrupt agentic loop
+                if (err instanceof ToolApprovalException) {
+                    await chatResultStream.writeResultBlock(this.#getUpdateBashConfirmResult(toolUse, false))
+                }
+
                 // If we did not approve a tool to be used or the user stopped the response, bubble this up to interrupt agentic loop
                 if (CancellationError.isUserCancelled(err) || err instanceof ToolApprovalException) {
```
jpinkney-aws, line 572: I guess we can remove this `err instanceof ToolApprovalException` now from here

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -521,6 +521,7 @@ export class AgenticChatController implements ChatHandlers {
                     session.setDeferredToolExecution(toolUse.toolUseId, deferred.resolve, deferred.reject)
                     this.#log(`Prompting for tool approval for tool: ${toolUse.name}`)
                     await deferred.promise
+                    await chatResultStream.writeResultBlock(this.#getUpdateBashConfirmResult(toolUse, true))
```
jpinkney-aws, line 524: is bash the only tool that needs confirmation right now and thats why it's safe to add that here?

Hweinstock (approved):
There is still a slight UI-bug but this fixes most of the behavior so I'm approving for now with the hopes it can get in the bug bash. 

We've identified the root-cause of desyncing between the final output and the intermediate results, just need more time to prepare a fix.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: Yeah, I plan to work on fsWrite approval next, and I can refactor this. We will need different mynah ids for that one I believe.

Hweinstock (approved):

PR #1068, opened 2025-04-22
feat: send chat update on undo

ctlai95:
## Problem

Card should be updated after undo is clicked

## Solution

Implement the chat update mechanism on button click

Refer to https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/amazonq/webview/ui/apps/cwChatConnector.ts#L394-L404 for previous implementation.

image

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -171,6 +171,26 @@ export class AgenticChatController implements ChatHandlers {
             const toolUseId = params.messageId
             try {
                 await this.#undoFileChange(toolUseId)
+                const cachedToolUse = this.#triggerContext.getToolUseLookup().get(toolUseId)
+                if (cachedToolUse) {
+                    this.#features.chat.sendChatUpdate({
+                        tabId: params.tabId,
+                        data: {
+                            messages: [
+                                {
+                                    ...cachedToolUse.chatResult,
+                                    header: {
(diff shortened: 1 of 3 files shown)


ctlai95:
ci should pass once https://github.com/aws/language-servers/pull/1069 is merged

zixlin7 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -171,6 +171,26 @@ export class AgenticChatController implements ChatHandlers {
             const toolUseId = params.messageId
             try {
                 await this.#undoFileChange(toolUseId)
+                const cachedToolUse = this.#triggerContext.getToolUseLookup().get(toolUseId)
+                if (cachedToolUse) {
+                    this.#features.chat.sendChatUpdate({
```
Hweinstock, line 176: What exactly does sendChatUpdate do? Is it force pushing this chat message to the client? Is there a reason we can't do this through the chatResultStream interface?

ctlai95: My understanding is this is used for updating an existing message with messageId. I guess since this isn't a progress message we shouldn't use chatResultStream? Not sure why one is under `this.#features.chat` and the other is under `this.#features.lsp` though

Hweinstock (approved):
Approving to unblock but worth checking with @kmile if using chat directly is right approach going forward

kmile:
@ctlai95  @Hweinstock Yup looks good to me!

PR #1070, opened 2025-04-22
fix: ux polishing for read tool messages.

laileni-aws:
## Problem
- Existing plugin shows read files multiple times and emitted per file.

image

## Solution
- Use a `messageIdToUpdate` state to store the messageId and update this for next read tool messages.
- This code change is equivalent to this PR: https://github.com/aws/aws-toolkit-vscode/pull/7006 in `feature/agentic-chat` branch.

image

## TODO:
- I will send out a follow up PR supporting List directory UX changes and mode refactored code.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
@@ -15,14 +15,22 @@ interface ResultStreamWriter {
 interface FileDetailsWithPath extends FileDetails {
     relativeFilePath: string
 }
+
+type OperationType = 'read' | 'write' | 'listDir'
+
+interface FileOperation {
+    type: OperationType
+    filePaths: FileDetailsWithPath[]
+}
 export class AgenticChatResultStream {
     static readonly resultDelimiter = '\n\n'
     #state = {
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
```diff
@@ -33,16 +41,31 @@ export class AgenticChatResultStream {
     getResult(only?: string): ChatResult {
         return this.#joinResults(this.#state.chatResultBlocks, only)
     }
+    getMessageIdToUpdate(): string | undefined {
+        return this.#state.messageIdToUpdate
+    }
```
kmile, line 46: Nit: if we add more message types, is this actually the 'read/query/search' message to update, and there might be others in the future?

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
laileni-aws: Yeah I agree
I am trying to use this same `messageIdToUpdate` for all the messages from `read`, `list` and `grep`(future). So we will be having only one messageId and can update all the messages contextList using this messageId.
Thats the plan if there are no challenges!

PR #1072, opened 2025-04-22
fix: ui polish for execute confirmation

zixlin7:
## Problem
- using status in header after button click
- keep the command in body of the message
## Solution
Screenshot 2025-04-22 at 4 16 19 PM

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -468,19 +468,18 @@ export class AgenticChatController implements ChatHandlers {
      * @param session
      */
     async waitForToolApproval(
-        toolUseId: string,
-        toolUseName: string,
+        toolUse: ToolUse,
         resultStream: AgenticChatResultStream,
         promptBlockId: number,
         session: ChatSessionService
     ) {
         const deferred = this.#createDeferred()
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -468,19 +468,18 @@ export class AgenticChatController implements ChatHandlers {
...
-        toolUseName: string,
+        toolUse: ToolUse,
         resultStream: AgenticChatResultStream,
         promptBlockId: number,
         session: ChatSessionService
     ) {
         const deferred = this.#createDeferred()
-        session.setDeferredToolExecution(toolUseId, deferred.resolve, deferred.reject)
-        this.#log(`Prompting for tool approval for tool: ${toolUseName}`)
+        session.setDeferredToolExecution(toolUse.toolUseId!, deferred.resolve, deferred.reject)
```
Hweinstock, line 477: is there anyway we can avoid the type assertion here? Perhaps by requiring it in the params?

Hweinstock (approved):

ctlai95 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
zixlin7: will followup

PR #1074, opened 2025-04-22
feat: add file search tool

grl-wang:
## Problem
required a file search tool to support search a directory and its subdirectories using regex patterns.
## Solution
adding a file search tool to support search a directory and its subdirectories using regex patterns.

Screenshot 2025-04-22 at 5 16 19 PM

TODO:
  the file search have same UX as the List directory. need to send out a follow up PR once List directory UX change is ready. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -454,7 +454,7 @@ export class AgenticChatController implements ChatHandlers {
             )
 
             //  Add the current assistantResponse message to the history DB
-            if (result.data?.chatResult.body) {
+            if (result.data?.chatResult.body !== undefined) {
                 this.#chatHistoryDb.addMessage(tabId, 'cwc', conversationIdentifier ?? '', {
                     body: result.data?.chatResult.body,
                     type: 'answer' as any,


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fileSearch.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    public async requiresAcceptance(path: string): Promise<CommandValidation> {
+        return { requiresAcceptance: !workspaceUtils.isInWorkspace(getWorkspaceFolderPaths(this.lsp), path) }
+    }
+
+    public async invoke(params: FileSearchParams): Promise<InvokeOutput> {
+        const path = sanitize(params.path)
+        try {
+            // Get all files and directories
+            const listing = await workspaceUtils.readDirectoryRecursively(
```
Hweinstock, line 79: This looks like a direct copy-paste of listDirectory. Is there any reason we can't extend listDirectory to accept a pattern?

grl-wang: this tool support filter the list directory result by regex patterns. 
Similar as extract the list directory tool as separate tool from fsRead. 

Based on the testing, extract this as a separate tool will increase the LLM to directly call the file system tool to search a directory instead of LLM executing the bash `find` command to search a tool.

Hweinstock: responded on internal thread.

PR #1075, opened 2025-04-22
fix: ux polish for list directory tool messages.

laileni-aws:
## Problem
- Existing UX shows multiple context list chatItem cards for list directory tool messages.
## Solution
- Use a `messageIdToUpdateListDir` state to store the messageId and update this for next list directory tool messages.
image

## TODO:
- Need to combine both read and list directory tool messages into one context list.
- This code change is equivalent to this PR: https://github.com/aws/aws-toolkit-vscode/pull/7006 in `feature/agentic-chat` branch.

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
```diff
@@ -30,6 +30,7 @@ export class AgenticChatResultStream {
         uuid: randomUUID(),
         messageId: undefined as string | undefined,
         messageIdToUpdate: undefined as string | undefined,
+        messageIdToUpdateListDir: undefined as string | undefined, //TODO: Combine both messageIdToUpdate and messageIdToUpdateListDir, Tried doing this but facing issue "errorMessage":"TypeError: Cannot read properties of undefined (reading 'push')"
```
zixlin7, line 33: can we combine this into a map with toolName as key and messageId as value?

laileni-aws: I tried this approach and I am facing issue `"errorMessage":"TypeError: Cannot read properties of undefined (reading 'push')"`
I am investigating on this! and also added a TODO for this PR.

laileni-aws: Right now we have a map between 
messageId as key and  toolName as value
kind of vice versa

zixlin7 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -691,34 +691,33 @@ export class AgenticChatController implements ChatHandlers {
...
-        if (toolUse.name !== 'fsRead') {
-            //TODO: Implement list directory UX in next PR.
-            return {}
-        }
-        let messageId = toolUse.toolUseId || ''
-        if (chatResultStream.getMessageIdToUpdate()) {
-            messageId = chatResultStream.getMessageIdToUpdate()!
-        } else if (messageId) {
-            chatResultStream.setMessageIdToUpdate(messageId)
+        let messageIdToUpdate = toolUse.toolUseId!
```
jpinkney-aws, line 694: is this always going to be defined? what happens if toolUseId isn't?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -691,34 +691,33 @@ export class AgenticChatController implements ChatHandlers {
...
+        const currentId = chatResultStream.getMessageIdToUpdateForTool(toolUse.name!)
```
jpinkney-aws, line 695: will toolUse name always be defined? what happens if name isn't?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
```diff
@@ -41,22 +41,23 @@ export class AgenticChatResultStream {
...
     /**
      * Adds a file operation for a specific message
      * @param messageId The ID of the message
-     * @param type The type of operation ('read' or 'listDir' or 'write')
+     * @param type The type of operation ('fsRead' or 'listDirectory' or 'fsWrite')
      * @param filePaths Array of FileDetailsWithPath involved in the operation
      */
-    addMessageOperation(messageId: string, type: OperationType, filePaths: FileDetailsWithPath[]) {
-        this.#state.messageOperations.set(messageId, { type, filePaths })
+    addMessageOperation(messageId: string, type: string, filePaths: FileDetailsWithPath[]) {
```
jpinkney-aws, line 59: can we make a type here instead of converting to string? then you wouldn't need to cast to an operation type on the next line

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
laileni-aws: We come to this part of code `processReadOrList` once toolName and toolId is available!

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
laileni-aws: Replied [here](https://github.com/aws/language-servers/pull/1075#discussion_r2056311759)

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
laileni-aws: I tried this approach,
OperationType can not be imported to `agenticChatController.ts` so explicitly mentioning here!

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
jpinkney-aws: Ah so really the ToolUse type should really be something like `Required<ToolUse>` as an input to this function since the optional fields are actually defined? Not a blocker, but worth a consideration

jpinkney-aws (approved):

PR #1077, opened 2025-04-23
feat: cancel transformation polling when region is changed

saketgo:
## Problem
When region is changed in Visual Studio IDE, cancel the transformation plan status polling in LSP.
## Solution
Created a new transform command: cancel polling which sets "CancelPollingEnabled" variable. When region is changed, this command is invoked from IDE and variable is set to true. If true, cancel the polling.
## Tests
Started a job in IDE using IAD, switched the region to FRA, polling is stopped. Previously, LSP was polling the job even if region is changed.
This change is backward compatible because the if condition will be true only when the command is called from IDE which will be the case for GA release.

## License


server/aws-lsp-codewhisperer/src/language-server/netTransform/netTransformServer.ts
```diff
@@ -128,6 +133,11 @@ export const QNetTransformServerToken =
                         )
                         return response
                     }
+                    case CancelPollingCommand: {
+                        const request = params as CancelPollingRequest
```
pranav-firake, line 137: do we need to pass boolean `CancelPollingEnabled` to this command?
if there is no case we can call this with `CancelPollingEnabled` as false, we can probably remove it

server/aws-lsp-codewhisperer/src/language-server/netTransform/netTransformServer.ts
```diff
@@ -43,8 +46,10 @@ const PollTransformForPlanCommand = 'aws/qNetTransform/pollTransformForPlan'
 const GetTransformPlanCommand = 'aws/qNetTransform/getTransformPlan'
 const CancelTransformCommand = 'aws/qNetTransform/cancelTransform'
 const DownloadArtifactsCommand = 'aws/qNetTransform/downloadArtifacts'
+const CancelPollingCommand = 'aws/qNetTransform/cancelPolling'
 import { SDKInitializator } from '@aws/language-server-runtimes/server-interface'
 import { AmazonQTokenServiceManager } from '../../shared/amazonQServiceManager/AmazonQTokenServiceManager'
+import { Console } from 'console'
```
ege0zcan, line 52: extra import here should be removed

server/aws-lsp-codewhisperer/src/language-server/netTransform/netTransformServer.ts
saketgo: Didn't understand your point

server/aws-lsp-codewhisperer/src/language-server/netTransform/netTransformServer.ts
saketgo: Removed

pranav-firake (approved):

huawenm (approved):

volodkevych (approved):

volodkevych (approved):

PR #1078, opened 2025-04-23
feat(agentic chat): add permission check for all tools and add UI permission cards

ashishrp-aws:
## Problem
User permissions check for reading files/list directory outside of workspace is not setup.

## Solution
- add permission check for fsRead and listDirectory tools
- Adds permission cards for allowing access

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -96,8 +96,8 @@ import { getNewPromptFilePath, getUserPromptsDirectory, promptFileExtension } fr
 import { ContextCommandsProvider } from './context/contextCommandsProvider'
 import { LocalProjectContextController } from '../../shared/localProjectContextController'
 import { CancellationError, workspaceUtils } from '@aws/lsp-core'
-import { FsReadParams } from './tools/fsRead'
-import { ListDirectoryParams } from './tools/listDirectory'
+import { FsRead, FsReadParams } from './tools/fsRead'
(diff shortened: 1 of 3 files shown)


kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -542,6 +543,19 @@ export class AgenticChatController implements ChatHandlers {
...
+                    const initialListDirResult = this.#processReadOrList(toolUse, chatResultStream)
+                    if (initialListDirResult) {
+                        await chatResultStream.writeResultBlock(initialListDirResult)
+                    }
+                } else if (toolUse.name === 'fsWrite') {
+                    const input = toolUse.input as unknown as FsWriteParams
+                    const document = await this.#triggerContext.getTextDocument(input.path)
+                    this.#triggerContext
+                        .getToolUseLookup()
+                        .set(toolUse.toolUseId, { ...toolUse, oldContent: document?.getText() })
```
kmile, line 556: Are we taking a snapshot before showing the "allow writing outside workspace" message? Could this snapshot then be outdated by the time the file is actually modified?

kmile (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
ashishrp-aws: This was existing context used by wiriting code diffs. But yeah it is a possibility, incase of text changes during this period, tool use fails and model retries with updated file information

PR #1081, opened 2025-04-23
feat(chat-client): open use input prompt for agentic chat and new prompt should st…

ashishrp-aws:
…op current response

## Problem
- User prompt is closed while agentic chat loop runs, the requirement is to keep it open and stop current running loop incase of new prompt.

## Solution
- promptInputDisabledState is set to false
- Each new prompt triggers stop response, so we can cancel the current loop incase of new prompt.

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -90,6 +90,7 @@ export const handleChatPrompt = (
     _eventId?: string
 ) => {
     let userPrompt = prompt.escapedPrompt
+    messager.onStopChatResponse(tabId)
     if (prompt.command) {
         // Temporary solution to handle clear quick actions on the client side
         if (prompt.command === '/clear') {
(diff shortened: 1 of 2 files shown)


ctlai95:
tests are failing

kmile (approved):

Hweinstock (approved):
Nice! This will be a great UI improvement. After fixing the tests this looks good to me.

chat-client/src/client/mynahUi.ts
```diff
@@ -90,6 +90,7 @@ export const handleChatPrompt = (
     _eventId?: string
 ) => {
     let userPrompt = prompt.escapedPrompt
+    messager.onStopChatResponse(tabId)
```
zixlin7, line 93: does this affect regular chat too? maybe that's ok

Hweinstock (approved):

PR #1085, opened 2025-04-23
fix: remove undefined header from followup authenticate button

volodkevych:
## Problem

We were seeing empty yellow header on top of followup buttons, like "Authenticate"

## Solution

Do not set header with empty properties.

Also fixed: TypeError: object is not iterable

## License


aditya169 (approved):

velatha:
curious to see, how this empty header got introduced ?it was not there till yesterday

viktorsaws (approved):

viktorsaws (approved):

PR #1088, opened 2025-04-23
fix: add result attribute when emitting telemetry event

opieter-aws:
## Problem
Telemetry events sent to VSCode are missing a `result` attribute, causing a warning in the VSCode logs:

```
[warning] telemetry: invalid Metric: "amazonq_exitFocusChat" emitted without the `result` property, which is always required. Consider using `.run()` instead of `.emit()`, which will set these properties automatically. See https://github.com/aws/aws-toolkit-vscode/blob/master/docs/telemetry.md#guidelines
```

## Solution
Add `result: 'Succeeded'` to the data when telemetry metrics are emitted

## Testing
Manually verified E2E that the warning is fixed

## License

### DIFF
--- a/chat-client/src/client/mynahUi.test.ts
+++ b/chat-client/src/client/mynahUi.test.ts
@@ -98,7 +98,7 @@ describe('MynahUI', () => {
             assert.notCalled(onQuickActionSpy)
             assert.calledWith(onChatPromptSpy, { prompt, tabId, context: undefined })
             assert.calledWith(addChatItemSpy, tabId, { type: ChatItemType.PROMPT, body: prompt.escapedPrompt })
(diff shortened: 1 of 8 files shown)


justinmk3 (approved):

Hweinstock:
do we need to add a `result: failed` somewhere or is this added implicitly?
I know on the VSC side its added implicitly inside our spans, but I don't see the same functionality here.

jpinkney-aws (approved):

jpinkney-aws:
I don't think any of these are modeled as succeed/failure metrics. Flare doesn't have the concept of spans

opieter-aws:
Indeed. Flare just emits based on a hook, no evaluation of success or failure for this is implemented.

Hweinstock (approved):

PR #1089, opened 2025-04-23
fix(chat-client): workspace checks for permission cards

ashishrp-aws:
## Problem
Permission check cards are appearing for all the files inside workspace. This is happening due to regresion in workspace.getTextDocument failure which works only for open documents on editor.

## Solution
- Fixing workspace check for all tools with lsp.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
@@ -2,6 +2,8 @@ import { sanitize } from '@aws/lsp-core/out/util/path'
 import { URI } from 'vscode-uri'
 import { CommandValidation, InvokeOutput, validatePath } from './toolShared'
 import { Features } from '@aws/language-server-runtimes/server-interface/server'
+import { workspaceUtils } from '@aws/lsp-core'
+import { getWorkspaceFolderPaths } from '@aws/lsp-core/out/util/workspaceUtils'
 
 // Port of https://github.com/aws/aws-toolkit-vscode/blob/5a0404eb0e2c637ca3bd119714f5c7a24634f746/packages/core/src/codewhispererChat/tools/fsRead.ts#L17
 
@@ -14,10 +16,12 @@ export class FsRead {
     static maxResponseSize = 200_000
     private readonly logging: Features['logging']
     private readonly workspace: Features['workspace']
+    private readonly lsp: Features['lsp']
 
-    constructor(features: Pick<Features, 'workspace' | 'logging'> & Partial<Features>) {
+    constructor(features: Pick<Features, 'lsp' | 'workspace' | 'logging'> & Partial<Features>) {
         this.logging = features.logging
         this.workspace = features.workspace
+        this.lsp = features.lsp
     }
 
     public async validate(params: FsReadParams): Promise<void> {
@@ -52,7 +56,14 @@ export class FsRead {
 
     public async requiresAcceptance(params: FsReadParams): Promise<CommandValidation> {
         // true when the file is not resolvable within our workspace. i.e. is outside of our workspace.
-        return { requiresAcceptance: !(await this.workspace.getTextDocument(URI.file(params.path).toString())) }
+        try {
+            const isInWorkspace = workspaceUtils.isInWorkspace(getWorkspaceFolderPaths(this.lsp), params.path)
+            return { requiresAcceptance: !isInWorkspace }
+        } catch (error) {
+            console.error('Error checking file acceptance:', error)
+            // In case of error, safer to require acceptance
+            return { requiresAcceptance: true }
+        }
     }
 
     public async invoke(params: FsReadParams): Promise<InvokeOutput> {
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
@@ -1,6 +1,6 @@
 import { sanitize } from '@aws/lsp-core/out/util/path'
 import { URI } from 'vscode-uri'
-import { CommandValidation, InvokeOutput, validatePath } from './toolShared'
+import { CommandValidation, InvokeOutput, requiresPathAcceptance, validatePath } from './toolShared'
 import { Features } from '@aws/language-server-runtimes/server-interface/server'
 import { workspaceUtils } from '@aws/lsp-core'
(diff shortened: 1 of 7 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
```diff
@@ -52,7 +56,14 @@ export class FsRead {
 
     public async requiresAcceptance(params: FsReadParams): Promise<CommandValidation> {
         // true when the file is not resolvable within our workspace. i.e. is outside of our workspace.
-        return { requiresAcceptance: !(await this.workspace.getTextDocument(URI.file(params.path).toString())) }
+        try {
+            const isInWorkspace = workspaceUtils.isInWorkspace(getWorkspaceFolderPaths(this.lsp), params.path)
+            return { requiresAcceptance: !isInWorkspace }
+        } catch (error) {
+            console.error('Error checking file acceptance:', error)
+            // In case of error, safer to require acceptance
```
Hweinstock, line 64: agreed, nice decision here.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsRead.ts
```diff
@@ -52,7 +56,14 @@ export class FsRead {
 
     public async requiresAcceptance(params: FsReadParams): Promise<CommandValidation> {
         // true when the file is not resolvable within our workspace. i.e. is outside of our workspace.
-        return { requiresAcceptance: !(await this.workspace.getTextDocument(URI.file(params.path).toString())) }
+        try {
+            const isInWorkspace = workspaceUtils.isInWorkspace(getWorkspaceFolderPaths(this.lsp), params.path)
+            return { requiresAcceptance: !isInWorkspace }
+        } catch (error) {
+            console.error('Error checking file acceptance:', error)
```
Hweinstock, line 63: I think we want to use the injected logger (`this.logging`) instead of console unless this message really must go to console.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
```diff
@@ -125,7 +129,15 @@ export class FsWrite {
     }
 
     public async requiresAcceptance(params: FsWriteParams): Promise<CommandValidation> {
-        return { requiresAcceptance: !(await this.workspace.getTextDocument(URI.file(params.path).toString())) }
+        // true when the file is not resolvable within our workspace. i.e. is outside of our workspace.
+        try {
+            const isInWorkspace = workspaceUtils.isInWorkspace(getWorkspaceFolderPaths(this.lsp), params.path)
```
Hweinstock, line 134: this looks like the same logic as that in `fsRead`, can we centralize it? Perhaps in toolShared?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/listDirectory.ts
```diff
@@ -53,7 +53,14 @@ export class ListDirectory {
     }
 
     public async requiresAcceptance(params: ListDirectoryParams): Promise<CommandValidation> {
-        return { requiresAcceptance: !workspaceUtils.isInWorkspace(getWorkspaceFolderPaths(this.lsp), params.path) }
+        try {
+            const isInWorkspace = workspaceUtils.isInWorkspace(getWorkspaceFolderPaths(this.lsp), params.path)
```
Hweinstock, line 57: same comment as above. I imagine we will use this check a lot to validate a path.

Hweinstock:
This test might also need to be updated. I believe it was relying on the old utility which is no longer used. 
```
should not require acceptance if fsPath is inside the workspace:

      AssertionError [ERR_ASSERTION]: Expected requiresAcceptance to be false for a path inside the workspace
      + expected - actual

      -true
      +false
      
      at Context.<anonymous> (src/language-server/agenticChat/tools/fsRead.test.ts:135:16)
```

ashishrp-aws:
fixing test and corrected it for comments

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolShared.ts
```diff
@@ -41,3 +45,36 @@ export enum OutputKind {
...
+            if (logging) {
+                logging.debug('No workspace folders found when checking file acceptance')
+            }
+            return { requiresAcceptance: true }
+        }
+        const isInWorkspace = workspaceUtils.isInWorkspace(workspaceFolders, path)
+        return { requiresAcceptance: !isInWorkspace }
+    } catch (error) {
+        if (logging) {
+            logging.debug(`Error checking file acceptance: ${error}`)
```
Hweinstock, line 75: (not a blocker) but should this be error level?

ashishrp-aws: fixed for it

Hweinstock:
LGTM, will merge once CI passes.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsWrite.ts
```diff
@@ -41,9 +43,11 @@ export interface FsWriteBackup {
 
 export class FsWrite {
     private readonly workspace: Features['workspace']
+    private readonly lsp: Features['lsp']
 
-    constructor(features: Pick<Features, 'workspace'> & Partial<Features>) {
+    constructor(features: Pick<Features, 'workspace' | 'lsp'> & Partial<Features>) {
```
pras0131, line 48: Can we also add logging to the Pick list here?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolShared.ts
```diff
@@ -41,3 +45,36 @@ export enum OutputKind {
...
+ *
+ * @param path The file path to check
+ * @param lsp The LSP feature to get workspace folders
+ * @param logging Optional logging feature for better error reporting
+ * @returns CommandValidation object with requiresAcceptance flag
+ */
+export async function requiresPathAcceptance(
+    path: string,
+    lsp: Features['lsp'],
+    logging?: Features['logging']
```
pras0131, line 61: nit: can make it non-optional instead by making the change above to accept logging inside fsWrite constructor and we won't need to check for if(logging) check to log stuff.

Hweinstock (approved):

PR #1098, opened 2025-04-23
refactor: remove use of eval from vec lib import

nborges-aws:
## Problem
Eval was added to the vector library import as quick solution to unblock build failures caused by webpack's static analysis resolving said import. We would like to avoid using eval but maintain webpack ignoring our import 
## Solution
Creating a dynamic import function here bypasses webpack's import resolution which enables the import to be properly imported at runtime, while avoiding the use of eval.

## License


server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -27,6 +27,7 @@ const LIBRARY_DIR = (() => {
     }
     return path.join(__dirname, 'indexing')
 })()
+const dynamicVecLibImport = new Function('path', 'return import(path)')
```
jpinkney-aws, line 30: are there any docs on why this works? Might be useful to include in the description + as an inline comment

volodkevych: Is this much different from `eval`?

nborges-aws: Good shout Josh, I've updated the description and added an inline comment for clarity. In terms of difference from `eval`, they are similar given that they're both evaluated at runtime. Aside from code cleanliness, the main upside here is that `new Function` creates an isolated scope to run in.

PR #1100, opened 2025-04-23
Add methods for consumers to to wait on initial project context index build

nborges-aws:
## Problem
Currently there is no way for consumers that rely on the project index to guarantee it fully completes building before use. 
## Solution
Add dedicated methods that allows consumers to check if indexing is in progress and await index being fully built

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
@@ -61,6 +61,7 @@ export class LocalProjectContextController {
     private maxIndexSizeMB?: number
     private respectUserGitIgnores?: boolean
     private indexCacheDirPath: string = path.join(homedir(), '.aws', 'amazonq', 'cache')
+    private _isIndexing = false
     private indexingComplete = false
 
     private readonly fileExtensions: string[] = Object.keys(languageByExtension)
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -182,9 +194,33 @@ export class LocalProjectContextController {
...
+                this.indexingComplete = true
             }
         } catch (error) {
             this.log.error(`Error building index: ${error}`)
+        } finally {
+            this._isIndexing = false
+        }
+    }
+
+    public async waitForIndexBuilt(timeoutMs: number = 30000): Promise<boolean> {
```
jpinkney-aws, line 206: do we have any data on how long indexes usually take? 

I also wonder if consumers of this api would prefer having it time out or instead have an event listener that notifies them when indexing is complete? What are your thoughts?

nborges-aws: I don't have any formalized data; only anecdotal through testing during my development. Building the index for ~400 files for me has consistently been under a second. 

In terms of the api, I think there could be value in exposing both options for consumers to use as needed. If a consumer makes a call that requires indexing completion, they can await completion with this method call. Otherwise they have the option of passive notification whenever it does complete

PR #1101, opened 2025-04-23
feat: support for project context in Q Chat

nborges-aws:
## Problem
Chat requests currently are not set up to accept and include context in chat requests 
## Solution
Support hydrating any chat request containing @workspace command with local project context.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
@@ -1,13 +1,17 @@
 import { TriggerType } from '@aws/chat-client-ui-types'
-import { ChatTriggerType, UserIntent, Tool, ToolResult } from '@amzn/codewhisperer-streaming'
+import { ChatTriggerType, UserIntent, Tool, ToolResult, RelevantTextDocument } from '@amzn/codewhisperer-streaming'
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
```diff
@@ -1,32 +1,46 @@
...
         this.#documentContextExtractor = new DocumentContextExtractor({ logger, workspace })
+        this.#logger = logger
     }
 
     async getNewTriggerContext(params: ChatParams | InlineChatParams): Promise<TriggerContext> {
         const documentContext: DocumentContext | undefined = await this.extractDocumentContext(params)
 
+        const useRelevantDocuments =
+            'context' in params
+                ? params.context?.some(context => typeof context !== 'string' && context.command === '@workspace')
```
ege0zcan, line 35: Why do we need `typeof context !== 'string' ` this. Isn't it guaranteed that context is going to be an `QuickActionCommand` object

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
nborges-aws: This catch, I don't see any explicit reason why this check is needed. I'll check in with Jonathan on Monday to see if there's a defensive necessity for this that I'm not seeing. cc @breedloj

PR #1104, opened 2025-04-24
feat: add ide category to auto-trigger threshold computation

andrewyuq:
JB and VSC need different IDE cofficients for auto trigger threshold

## Problem

JB and VSC need different IDE cofficients for auto trigger threshold

## Solution

Pass IDE category as the parameter and get different coefficient based on this info

## License


server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
```diff
@@ -289,11 +289,16 @@ export const CodewhispererServerFactory =
...
+                    const initializeParams = lsp.getClientInitializeParams()
+                    if (initializeParams !== undefined) {
+                        ideCategory = makeUserContextObject(initializeParams, runtime.platform, 'INLINE')?.ideCategory
+                    }
                     const autoTriggerResult = autoTrigger({
                         fileContext, // The left/right file context and programming language
                         lineNum: params.position.line, // the line number of the invocation, this is the line of the cursor
                         char: triggerCharacter, // Add the character just inserted, if any, before the invication position
-                        ide: '', // TODO: Fetch the IDE in a platform-agnostic way (from the initialize request?)
+                        ide: ideCategory ?? "VSCODE",
```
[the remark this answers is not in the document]
andrewyuq: so when this feature was developed we were only supporting JB and VSC, and for all other IDEs, there isn't a coefficient for them, I think the current logic for other IDEs is that they will use VSC coefficient and this change is remaining this behavior.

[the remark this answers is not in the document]
andrewyuq: acutally for VSC we will have a negative shift, for all other IDEs they will be 0.

jpinkney-aws (approved):

server/aws-lsp-codewhisperer/src/shared/telemetryUtils.ts
```diff
@@ -47,8 +47,9 @@ export const getUserAgent = (initializeParams: InitializeParams, serverInfo?: Se
 }
 
 const IDE_CATEGORY_MAP: { [key: string]: IdeCategory } = {
+    // TODO: VSCode key needs to change for getting the correct coefficient value for inline
     'AmazonQ-For-VSCode': 'VSCODE',
-    'Amazon-Q-For-JetBrains': 'JETBRAINS',
+    'Amazon Q For JetBrains': 'JETBRAINS',
```
volodkevych, line 52: What is this inconsistency caused by? Cannot we pass extension name in `clientInfo` following the same format as for other IDEs?

server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
```diff
@@ -289,11 +289,16 @@ export const CodewhispererServerFactory =
                     const codewhispererAutoTriggerType = triggerType(fileContext)
                     const previousDecision =
                         sessionManager.getPreviousSession()?.getAggregatedUserTriggerDecision() ?? ''
+                    let ideCategory: string | undefined = ''
+                    const initializeParams = lsp.getClientInitializeParams()
+                    if (initializeParams !== undefined) {
+                        ideCategory = makeUserContextObject(initializeParams, runtime.platform, 'INLINE')?.ideCategory
```
volodkevych, line 295: Why don't we call `getIdeCategory()` here if we need just `ideCategory`?

server/aws-lsp-codewhisperer/src/shared/telemetryUtils.ts
andrewyuq: The extension name for JetBrains is "Amazon Q For JetBrains", with spaces, not sure why the version with hyphens are added there, this is tested with Jb talking to flare, 
For Eclipse and VS, they should be already talking to flare so assuming the names are correct, for VSC I'm not sure, it will be determined when we decide to move VSC inline to flare.

volodkevych (approved):

PR #1111, opened 2025-04-24
fix: validate tool output content size

ctlai95:
## Problem

Service is sometimes throwing ValidationException with reason `CONTENT_LENGTH_EXCEEDS_THRESHOLD` if content becomes too large.

## Solution

Validate the tool output to ensure the results are not too large

Previous implementation: https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/codewhispererChat/tools/toolUtils.ts#L75-L90

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -628,6 +628,7 @@ export class AgenticChatController implements ChatHandlers {
                 } else if (typeof result === 'object') {
                     toolResultContent = { json: result }
                 } else toolResultContent = { text: JSON.stringify(result) }
+                this.#validateToolResult(toolUse, toolResultContent)
 
                 results.push({
                     toolUseId: toolUse.toolUseId,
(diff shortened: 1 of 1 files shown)


tsmithsz (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -698,6 +699,27 @@ export class AgenticChatController implements ChatHandlers {
         return results
     }
 
+    #validateToolResult(toolUse: ToolUse, result: ToolResultContentBlock) {
+        let maxToolResponseSize
+        switch (toolUse.name) {
+            case 'fsRead':
+                maxToolResponseSize = 200_000
```
tsmithsz, line 706: nit: Should we store these as constants somewhere?

ctlai95: I think we should but not sure where, maybe in runtimes?

tsmithsz: Possibly. I haven't really looked into that package to be honest. Maybe we can just add it to the constants file here or create one for agenticChat: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/chat/constants.ts

ege0zcan:
Does it makes sense to use partial results of a long output by truncating the long output?  In the current case we fail the validation and the tool result doesn't get added to the message.

PR #1115, opened 2025-04-24
fix: implement proper error handling.

Hweinstock:
## Problem
Error handling for backend requests is non-existent and the existing error handling elsewhere relies on flaky checks. 

Testing these changes with @jpinkney-aws, we identified 3 problems fixed by these changes. 
- backend errors cause chat to hang. 
- the stop button doesn't always stop the action depending on the time it was pressed. 
- the authenticate button doesn't pop-up, its blocked by "thinking...". 

## Solution
- forward backend error message to the client to be displayed in chat. 
- on error, send a progress update to the client to kill the thinking bubble. 
- centralize error handling to minimize flaky checks. 

## Testing
- In addition to unit tests, these changes were tested e2e. See demo below for a demonstration where I have the model always fail on the second iteration of the agentic loop. 

https://github.com/user-attachments/assets/944dd737-a2e9-4b82-9976-a3addc676b42

## Notes
there is a hack implemented to ensure the thinking bubble is killed. This highlights potential bugs in the chat-client and is explained with a comment in the code. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -1144,7 +1144,7 @@ export class AgenticChatController implements ChatHandlers {
             this.#telemetryController.emitMessageResponseError(tabId, metric.metric, err.requestId, err.message)
         }
 
-        // return non-model errors back to the client.
+        // return non-model errors back to the client as errors
         if (!(err instanceof ModelServiceException)) {
             this.#log(`unknown error ${err instanceof Error ? JSON.stringify(err) : 'unknown'}`)
             this.#debug(`stack ${err instanceof Error ? JSON.stringify(err.stack) : 'unknown'}`)
@@ -1177,7 +1177,7 @@ export class AgenticChatController implements ChatHandlers {
         }
 
         const backendError = err.cause
-        // Send the backend error message directly to the client in chat.
+        // Send the backend error message directly to the client to be displayed in chat.
         return {
             type: 'answer',
             body: backendError.message,
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -101,6 +101,7 @@ import { ListDirectory, ListDirectoryParams } from './tools/listDirectory'
 import { FsWrite, FsWriteParams, getDiffChanges } from './tools/fsWrite'
 import { ExecuteBash, ExecuteBashOutput, ExecuteBashParams } from './tools/executeBash'
 import { ExplanatoryParams, InvokeOutput, ToolApprovalException } from './tools/toolShared'
+import { ModelServiceException } from './errors'
 
 type ChatHandlers = Omit<
     LspHandlers<Chat>,
@@ -334,7 +335,20 @@ export class AgenticChatController implements ChatHandlers {
                 chatResultStream
             )
         } catch (err) {
-            if (this.isUserAction(err)) {
+            // HACK: the chat-client needs to have a partial event with the associated messageId sent before it can accept the final result.
+            // Without this, the `thinking` indicator never goes away.
+            // Note: buttons being explicitly empty is required for this hack to work.
+            const errorMessageId = `error-message-id-${uuid()}`
+            await this.#sendProgressToClient(
+                {
+                    type: 'answer',
+                    body: '',
+                    messageId: errorMessageId,
+                    buttons: [],
+                },
+                params.partialResultToken
+            )
+            if (this.isUserAction(err, token)) {
                 /**
                  * when the session is aborted it generates an error.
                  * we need to resolve this error with an answer so the
@@ -343,9 +357,11 @@ export class AgenticChatController implements ChatHandlers {
                 return {
                     type: 'answer',
                     body: '',
+                    messageId: errorMessageId,
+                    buttons: [],
                 }
             }
-            return this.#handleRequestError(err, params.tabId, metric)
+            return this.#handleRequestError(err, errorMessageId, params.tabId, metric)
         }
     }
 
@@ -436,10 +452,9 @@ export class AgenticChatController implements ChatHandlers {
             }
 
             // Phase 3: Request Execution
-            this.#debug(`Request Input: ${JSON.stringify(currentRequestInput)}`)
-
-            const response = await session.generateAssistantResponse(currentRequestInput)
-            this.#debug(`Response received for iteration ${iterationCount}:`, JSON.stringify(response.$metadata))
+            const response = await this.fetchModelResponse(currentRequestInput, i =>
+                session.generateAssistantResponse(i)
+            )
 
             // Phase 4: Response Processing
             const result = await this.#processGenerateAssistantResponseResponse(
@@ -669,7 +684,7 @@ export class AgenticChatController implements ChatHandlers {
                     this.#features.chat.sendChatUpdate({ tabId, state: { inProgress: false } })
                 }
 
-                if (this.isUserAction(err)) {
+                if (this.isUserAction(err, token)) {
                     if (err instanceof ToolApprovalException && toolUse.name === 'executeBash') {
                         if (buttonBlockId) {
                             await chatResultStream.overwriteResultBlock(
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -439,10 +454,9 @@ export class AgenticChatController implements ChatHandlers {
             }
 
             // Phase 3: Request Execution
-            this.#debug(`Request Input: ${JSON.stringify(currentRequestInput)}`)
```
jpinkney-aws, line 442: would this still be useful to keep since we can see request ids?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1104,12 +1144,23 @@ export class AgenticChatController implements ChatHandlers {
             this.#telemetryController.emitMessageResponseError(tabId, metric.metric, err.requestId, err.message)
         }
 
-        if (err instanceof AmazonQServicePendingSigninError) {
+        // return non-model errors back to the client as errors
+        if (!(err instanceof ModelServiceException)) {
```
jpinkney-aws, line 1148: would an auth error end up going through this flow? Since it's not a model service exception

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1125,13 +1176,14 @@ export class AgenticChatController implements ChatHandlers {
...
-        this.#debug(`Q api request error cause ${err instanceof Error ? JSON.stringify(err.cause) : 'unknown'}`)
-        return new ResponseError<ChatResult>(
-            LSPErrorCodes.RequestFailed,
-            err instanceof Error ? err.message : 'Unknown request error'
-        )
+        const backendError = err.cause
+        // Send the backend error message directly to the client to be displayed in chat.
+        return {
+            type: 'answer',
+            body: backendError.message,
```
jpinkney-aws, line 1183: should we attach the request id to this since we know it's coming from the backend?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: I do this within `fetchModelResponse` now. Agreed this should be kept.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: auth errors come from the SDK request, so they are thrown by `fetchModelRequest` and thus wrapped as well.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: I believe this is already logged in `fetchModelResponse`. Is there any advantage of displaying this information in the chat?

jpinkney-aws: Just to make it easier for us to debug user issues 😆. As long as its logged it should be fine

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1317,7 +1369,7 @@ export class AgenticChatController implements ChatHandlers {
         } else if (toolUse?.name === 'fsRead') {
             await this.#features.lsp.window.showDocument({ uri: params.filePath })
         } else {
-            const absolutePath = params.fullPath ?? (await this.#resolveAbsolutePath(params.filePath))
+            const absolutePath = params.filePath ?? (await this.#resolveAbsolutePath(params.filePath))
```
jpinkney-aws, line 1372: what is this fix for?

Hweinstock: This was giving an error to a dependency update. Removed this change now that I've clean installed and the error is gone.

jpinkney-aws (approved):

jpinkney-aws (approved):
Re-approving after the last change

zixlin7 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1140,13 +1191,14 @@ export class AgenticChatController implements ChatHandlers {
...
-        this.#log(`Q api request error ${err instanceof Error ? JSON.stringify(err) : 'unknown'}`)
-        this.#debug(`Q api request error stack ${err instanceof Error ? JSON.stringify(err.stack) : 'unknown'}`)
-        this.#debug(`Q api request error cause ${err instanceof Error ? JSON.stringify(err.cause) : 'unknown'}`)
-        return new ResponseError<ChatResult>(
-            LSPErrorCodes.RequestFailed,
-            err instanceof Error ? err.message : 'Unknown request error'
-        )
+        const backendError = err.cause
+        // Send the backend error message directly to the client to be displayed in chat.
+        return {
```
volodkevych, line 1196: Another, more LSP, way to handle this would be to return `ResponseError` in result of LSP request. Then extension would need to translate that to error on the UI by calling "errorMessage" command with params through postMessage.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -729,6 +740,34 @@ export class AgenticChatController implements ChatHandlers {
...
+        makeRequest: (requestInput: RequestType) => Promise<ResponseType>
+    ): Promise<ResponseType> {
+        this.#debug(`Q Backend Request: ${JSON.stringify(requestInput)}`)
+        try {
+            const response = await makeRequest(requestInput)
+            this.#debug(`Q Backend Response: ${JSON.stringify(response)}`)
+            return response
+        } catch (e) {
+            this.#features.logging.error(`Error in call: ${JSON.stringify(e)}`)
+            throw new ModelServiceException(e as Error)
```
volodkevych, line 767: Technically, we also throw `throw new Error('amazonQServiceManager is not initialized')` from within `generateAssistantResponse` (aka `makeRequest`) function. Which is something that should not be show to user.

Additionally, is exception thrown from Q API user-friendly enough to show it to user?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: Yeah, that makes sense. My thinking was that if we wanted any custom formatting on this message in the future it could be done once on the LSP side rather than reimplemented by each client.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: We were asking the same question. We reached out the backend team and they said this should be okay for now, but we'll likely want something better in the long run. 

I was hoping to find a better solution here, but wasn't able to given the time constraints.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
volodkevych: `ResponseError` has generic `data` param, so that might be used. Custom formatting could also be on `chat-client` side ([here](https://github.com/aws/language-servers/blob/6e813fdc97c1494337cd9ffc724ecd25331625b8/chat-client/src/client/mynahUi.ts#L740)), depending on what exactly you mean.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1119,12 +1159,23 @@ export class AgenticChatController implements ChatHandlers {
...
 
-        if (err instanceof AmazonQServicePendingSigninError) {
+        // return non-model errors back to the client as errors
+        if (!(err instanceof ModelServiceException)) {
+            this.#log(`unknown error ${err instanceof Error ? JSON.stringify(err) : 'unknown'}`)
+            this.#debug(`stack ${err instanceof Error ? JSON.stringify(err.stack) : 'unknown'}`)
+            this.#debug(`cause ${err instanceof Error ? JSON.stringify(err.cause) : 'unknown'}`)
+            return new ResponseError<ChatResult>(
+                LSPErrorCodes.RequestFailed,
+                err instanceof Error ? err.message : 'Unknown request error'
```
justinmk3, line 1169: Are there any plans to add something like `innerCode` which holds the `code` (sdk v2) or `name` (sdk v3) from the original error?

Hweinstock: it sounds like their isn't yet a reliable way to distinguish errors produced by the backend yet ([source](https://github.com/aws/language-servers/pull/1167#discussion_r2062732356)). So rather than rely on their errors, I refactored this to wrap their error and include a `code` [field](https://github.com/aws/language-servers/blob/daca805d06f731d2a5dd3f86172b86580cf0e69e/server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts#L3) , but only on the language server side, and is not sent back to the client. 

My intention with this error handling is to do as much on the language server side as possible, such that the client will only have to render the `ChatResult` from the `data` field of the `ResponseError`. My thinking is that this will reduce the amount of custom error handling each IDE will have to implement.

justinmk3: That makes. But for unforeseen cases, we might want to give clients an escape-hatch? Else they will need to do their own checks, or check the message text, for errors that require client-side handling (such as https://github.com/aws/language-servers/pull/1197 )

PR #1124, opened 2025-04-24
refactor: propagate model errors via LSP response errors. 

Hweinstock:
## Problem
- Some follow up work from https://github.com/aws/language-servers/pull/1115. 

## Solution
- Follow the LSP paradigm to return response error to the client, and handle it there. 
- This is in combination with client side changes [here](https://github.com/aws/aws-toolkit-vscode/pull/7161). 
- These together produce the following experience on model api failures. 
image

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1115,14 +1115,12 @@ export class AgenticChatController implements ChatHandlers {
             return createAuthFollowUpResult(authFollowType)
         }
 
-        const backendError = err.cause
-        // Send the backend error message directly to the client to be displayed in chat.
-        return {
+        return new ResponseError<ChatResult>(LSPErrorCodes.RequestFailed, err.cause.message, {
```
jpinkney-aws, line 1118: will the error cause message be logged and then the other body be shown to the user?

volodkevych (approved):

volodkevych:
Thanks for the followup.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
Hweinstock: Yeah, see the screenshot I added above. I was debating try to parse the requestId and include it in the chat message (done on the agentic-chat vsc branch), but I think it makes more sense to be the logs since its a technical detail and only useful for internal users. 

I think ideally we add a button to this chat message that opens the logs (like we do in most VSC error messages). However, this is not a p0 to me since it requires a decent amount of work to allow Flare to open IDE logs.

Hweinstock:
Should be safe to assume this is merged with [client side changes](https://github.com/aws/aws-toolkit-vscode/pull/7161) with the version fixing setup, but want to double-check in case they get out of sync since this is fairly critical. Will open PR once I've verified.

jpinkney-aws (approved):

PR #1127, opened 2025-04-25
fix: bubble up throttling error for ListAvailableProfiles API to client

shruti0085:
## Problem
When a throttling exception is thrown with the ListAvailableProfiles API, the error is swallowed and client is informed of a generic failure without information on the cause of the failure. Since clients do not have this information, a vague message is shown to the user similar to `Unable to retrieve profiles`. We want this to be bubbled upstream to the client
## Solution
Whenever a throttling exception is thrown with the ListAvailableProfiles API, a ResponseError with a specific awsErrorCode `E_AMAZON_Q_PROFILE_THROTTLING` is now sent back for the client to interpret and use for showing an appropriate message to the user.
Note: When atleast one of the calls fail for ListProfiles invoked for different regions but no profiles are found, a throttling error will be thrown back to the client.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/qDeveloperProfiles.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/qDeveloperProfiles.ts
@@ -8,6 +8,7 @@ import {
 import { isBool, isObject, SsoConnectionType } from '../utils'
 import { AWS_Q_ENDPOINTS } from '../../shared/constants'
 import { CodeWhispererServiceToken } from '../codeWhispererService'
+import { AmazonQServiceProfileThrottlingError } from './errors'
 
 export interface AmazonQDeveloperProfile {
     arn: string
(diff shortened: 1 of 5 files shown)


server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/qDeveloperProfiles.ts
```diff
@@ -60,13 +61,28 @@ export const getListAllAvailableProfilesHandler =
         }
 
         const fulfilledResults = result.filter(settledResult => settledResult.status === 'fulfilled')
+        const hasThrottlingError = result.some(
```
viktorsaws, line 64: Could you also add unit test in `server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/qDeveloperProfiles.test.ts` for `getListAllAvailableProfilesHandler` function? It is used in more places that qConfigurationServer.

viktorsaws (approved):

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/qDeveloperProfiles.ts
shruti0085: Added test

viktorsaws (approved):

PR #1128, opened 2025-04-25
fix: update pp mode in tab store

yueny2020:
## Problem
- update pp mode in tab store

## Solution
- The pp mode is per tab and every time the pp mode change needs to update the tab store. Then use the pp mode in tab store for ui display.

##### Before change: the pp mode is cross tab and doesn't align with the pp mode toggle button.
 - In Tab1, the pp mode = off. Then click add new tab button, the Tab2 was created with pp mode = on, but the code block still showed `insert-to-cursor`.

Tab1:
image
Tab2:
image

##### After change: the pp mode is per tab and align with the pp mode toggle button.
- In Tab1, the pp mode = off. Then click add new tab button, the Tab2 was created with pp mode = on, but the code block did NOT showed `insert-to-cursor`.

Tab1:
image
Tab2
image

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -39,6 +39,7 @@ import {
     NotificationType,
     MynahUIProps,
     QuickActionCommand,
+    ChatItemFormItem,
 } from '@aws/mynah-ui'
 import { VoteParams } from '../contracts/telemetry'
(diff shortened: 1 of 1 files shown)


laileni-aws (approved):
LGTM.
For context pp = pair programming mode.

ctlai95 (approved):

PR #1131, opened 2025-04-25
chore: emit telemetry for inline chat result

pras0131:
## Problem

The PR contains the changes to emit telemetry regarding an action taken against inline chat result response by the user.

## License


server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
```diff
@@ -18,6 +18,14 @@ import {
...
     ChatAddMessageEvent,
     UserIntent,
+    InlineChatEvent,
+    UUID,
+    Timestamp,
+    PrimitiveInteger,
+    Boolean,
+    InlineChatUserDecision,
+    Double,
+    ProgrammingLanguage,
```
ege0zcan, line 28: Do we need all these new types imported? I don't see them being used

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
pras0131: Looks like not all of them are required, let me remove the ones not in use.

ege0zcan (approved):

PR #1141, opened 2025-04-26
fix: invalid json aborts the loop

ctlai95:
## Problem

Sometimes the server can respond with an invalid JSON as input and causing the entire process to stop/hang.

## Solution

1. Handle the error gracefully so that it doesn't stop the agentic loop partway through
2. Send the error back to the server as a valid ToolResult with error

Previous implementation: https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/codewhispererChat/controllers/chat/controller.ts#L754-L758

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -14,6 +14,7 @@ import {
     SendMessageCommandOutput,
     ToolResult,
     ToolResultContentBlock,
+    ToolResultStatus,
     ToolUse,
 } from '@amzn/codewhisperer-streaming'
 import {
@@ -448,6 +449,11 @@ export class AgenticChatController implements ChatHandlers {
                 ? []
                 : this.#chatHistoryDb.getMessages(tabId)
 
+            // Phase 3: Request Execution
+            const response = await this.fetchModelResponse(currentRequestInput, i =>
+                session.generateAssistantResponse(i)
+            )
+
             //  Add the current user message to the history DB
             if (currentMessage && conversationIdentifier) {
                 this.#chatHistoryDb.addMessage(tabId, 'cwc', conversationIdentifier, {
@@ -458,11 +464,6 @@ export class AgenticChatController implements ChatHandlers {
                     userInputMessageContext: currentMessage.userInputMessage?.userInputMessageContext,
                 })
             }
-
-            // Phase 3: Request Execution
-            const response = await this.fetchModelResponse(currentRequestInput, i =>
-                session.generateAssistantResponse(i)
-            )
 
             // remove the temp loading message when we have response
             if (loadingMessageId) {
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -521,8 +522,22 @@ export class AgenticChatController implements ChatHandlers {
...
+                    toolUseId: toolUse.toolUseId,
+                    status: ToolResultStatus.ERROR,
+                    content: [{ text: result.error }],
+                }))
+                if (result.error.startsWith('ToolUse input is invalid JSON:')) {
+                    currentRequestInput.conversationState!.currentMessage!.userInputMessage!.content =
+                        "Your toolUse input isn't valid. Please check the syntax and make sure the input is complete. If the input is large, break it down into multiple tool uses with smaller input."
+                }
+            }
             currentRequestInput = this.#updateRequestInputWithToolResults(currentRequestInput, toolResults)
```
jguoamz, line 541: Asking out of ignorance, would this method overwrite the currentRequestInput?

ctlai95: it's already been used at this point for the current request, so this is really updating for the next request

jguoamz (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -448,6 +449,11 @@ export class AgenticChatController implements ChatHandlers {
                 ? []
                 : this.#chatHistoryDb.getMessages(tabId)
 
+            // Phase 3: Request Execution
```
jpinkney-aws, line 452: what was the rational about moving these before the current user message was added to history?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -521,8 +522,22 @@ export class AgenticChatController implements ChatHandlers {
...
+                // Process tool uses and update the request input for the next iteration
+                toolResults = await this.#processToolUses(pendingToolUses, chatResultStream, session, tabId, token)
+            } else {
+                // Send an error card to UI?
+                toolResults = pendingToolUses.map(toolUse => ({
+                    toolUseId: toolUse.toolUseId,
+                    status: ToolResultStatus.ERROR,
+                    content: [{ text: result.error }],
+                }))
+                if (result.error.startsWith('ToolUse input is invalid JSON:')) {
```
jpinkney-aws, line 536: is this just a hardcoded response somewhere?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatEventParser.ts
```diff
@@ -130,10 +130,19 @@ export class AgenticChatEventParser implements ChatResult {
                 }
 
                 if (toolUseEvent.stop) {
-                    const parsedInput =
-                        typeof this.toolUses[toolUseId].input === 'string'
-                            ? JSON.parse(this.toolUses[toolUseId].input === '' ? '{}' : this.toolUses[toolUseId].input)
-                            : this.toolUses[toolUseId].input
+                    let parsedInput
+                    try {
+                        parsedInput =
```
jpinkney-aws, line 135: theres a lot of conditional logic in this, can we unwind it so it makes it a little easier to understand

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
ctlai95: We were thinking that it's possible for the request to fail which means we shouldn't add it to history. So I think it's better to add it after the request is complete

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
ctlai95: Actually I realized I don't need this, will remove it

jguoamz (approved):

PR #1156, opened 2025-04-27
fix: remove guessIntentFromPrompt functionality while preserving user Intent property

mr-lee:
## Problem

User intents are no longer relevant for chat.

## Solution

Remove the guessing capability.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
@@ -38,7 +38,7 @@ export class QChatTriggerContext {
 
         return {
             ...documentContext,
-            userIntent: undefined,
+            userIntent: this.#guessIntentFromPrompt(params.prompt.prompt),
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
```diff
@@ -38,7 +38,7 @@ export class QChatTriggerContext {
 
         return {
             ...documentContext,
-            userIntent: this.#guessIntentFromPrompt(params.prompt.prompt),
```
jguoamz, line 41: I think technically we don't need to change this file, the agentic flow goes through agenticChatTriggerContext.ts

mr-lee: ope, good point let me put it back.

tsmithsz (approved):

jguoamz (approved):

ctlai95 (approved):

PR #1167, opened 2025-04-27
feat: allow backend errors to be handled seperately

Hweinstock:
## Problem
Backend error messages are expected to be shown directly to the customer in chat. However, there are other error messages we don't want to show, and should show a generic message. 

## Solution
- Implement an error wrapper with a `code` field, that allows each type of error to have its own handling. 
- Expand reach of `try` block such that any internal errors surfaced are caught and returned as `ResponseError`. This also makes this easier to test. 

## Testing and Verification
- updated a unit test, and added another for generic behavior. 
- Backend error (replacing backend call with an error)

https://github.com/user-attachments/assets/12b44857-6a52-4d8f-8e7b-55fcac374429

- generic error (throwing an error elsewhere in the agentic loop)

https://github.com/user-attachments/assets/8ded36f1-018d-401e-8a7f-0a90ee818f59

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -339,24 +339,25 @@ export class AgenticChatController implements ChatHandlers {
             cwsprChatConversationType: 'AgenticChat',
         })
 
-        const triggerContext = await this.#getTriggerContext(params, metric)
-        const isNewConversation = !session.conversationId
-        if (isNewConversation) {
-            // agentic chat does not support conversationId in API response,
-            // so we set it to random UUID per session, as other chat functionality
-            // depends on it
-            session.conversationId = uuid()
-        }
-
-        token.onCancellationRequested(() => {
-            this.#log('cancellation requested')
-            this.#telemetryController.emitInteractWithAgenticChat('StopChat', params.tabId)
-            session.abortRequest()
-            session.rejectAllDeferredToolExecutions(new CancellationError('user'))
-        })
-
-        const chatResultStream = this.#getChatResultStream(params.partialResultToken)
         try {
+            const triggerContext = await this.#getTriggerContext(params, metric)
+            const isNewConversation = !session.conversationId
+            if (isNewConversation) {
+                // agentic chat does not support conversationId in API response,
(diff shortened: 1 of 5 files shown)


velatha (changes requested):
Is it possible to add a bit more context than just "'This is an error from the backend'"

Hweinstock:
That was a demo message I threw in there to demonstrate, an actual error from the backend would display its message there. I can try to produce an ISE to show what this would look like.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1426,9 +1401,20 @@ export class AgenticChatController implements ChatHandlers {
             return createAuthFollowUpResult(authFollowType)
         }
 
-        return new ResponseError<ChatResult>(LSPErrorCodes.RequestFailed, err.cause.message, {
+        // Show backend error messages to the customer.
+        if (err.code === 'QModelResponse') {
+            this.#features.logging.error(`QModelResponse Error: ${JSON.stringify(err)}`)
+            return new ResponseError<ChatResult>(LSPErrorCodes.RequestFailed, err.message, {
```
jguoamz, line 1407: can we add a special check here to show this message when statusCode is 500?
https://github.com/aws/aws-toolkit-vscode/pull/6984/files

Hweinstock: Is it safe to assume 500 is throttling? Isn't this also ISE?

Hweinstock:
@velatha Internal failure error demo: 

https://github.com/user-attachments/assets/ad4b259c-d028-41aa-afe5-227b35b3125f

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
jguoamz: you are right, right now this is a gap in our system where we couldn't differentiate bedrock throttling and real ISE. Given the fact that we just got a lot of capacity and we fallback to 3.5, the chance of real throttling happening is low so maybe it's fine to keep this as it is for now

jguoamz (approved):

ctlai95 (approved):

velatha (approved):

PR #1169, opened 2025-04-27
feat: show agentLoop error in chat

Hweinstock:
## Problem
When the agent exceeds our threshold (currently 100), the frontend doesn't render anything. This is because we return a ResponseError without a valid ChatResult body. 

## Solution
- Centralize all error handling to the `handleRequestError` method. I.e. do not return response errors from anywhere else. This allows us to ensure we are returning valid ChatResults in the data field. 
- Show a specialized message for agent loop exceeding. 

## Verification

https://github.com/user-attachments/assets/ef4b2ddc-5355-4658-bb85-c77984ebf330

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -971,7 +971,7 @@ describe('AgenticChatController', () => {
             assert.deepStrictEqual(chatResult, utils.createAuthFollowUpResult('full-auth'))
         })
 
-        it.only('returns a ResponseError if response streams returns an error event', async () => {
+        it('returns a ResponseError if response streams returns an error event', async () => {
             generateAssistantResponseStub.callsFake(() => {
                 return Promise.resolve({
                     $metadata: {
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -25,6 +25,7 @@ import {
     InlineChatResult,
     CancellationToken,
     CancellationTokenSource,
+    ErrorCodes,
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1312,9 +1312,9 @@ export class AgenticChatController implements ChatHandlers {
         triggerContext: TriggerContext,
         isNewConversation: boolean,
         chatResultStream: AgenticChatResultStream
-    ): Promise<ChatResult | ResponseError<ChatResult>> {
+    ): Promise<ChatResult> {
         if (!result.success) {
-            return new ResponseError<ChatResult>(LSPErrorCodes.RequestFailed, result.error)
+            throw new AgenticChatError(result.error, 'FailedResult')
```
ctlai95, line 1317: I think we need to pass in a chatResult here for the message to automatically show up

ctlai95 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
```diff
@@ -970,7 +971,7 @@ describe('AgenticChatController', () => {
             assert.deepStrictEqual(chatResult, utils.createAuthFollowUpResult('full-auth'))
         })
 
-        it('returns a ResponseError if response streams return an error event', async () => {
+        it.only('returns a ResponseError if response streams returns an error event', async () => {
```
ctlai95, line 974: I assume `.only` was committed by accident?

Hweinstock: yes, good catch. I'm used to the VSC repo that fails the build on these.

ctlai95 (approved):

zixlin7 (approved):

laileni-aws (approved):

volodkevych (approved):

PR #1172, opened 2025-04-28
feat(chat-client): add feature flag to toggle agentic mode

viktorsaws:
## Problem
We introduced few breaking changes to Chat Client, which change interaction pattern and maybe not compatible with non-agentic chat server or extension.

## Solution
Put changed behavior behind `config.agenticMode` flag, set by Extension and passed as argument to `createChat` function at initializing Chat Client in Webview code. Old releases of extensions are not impacted - agenticMode is disabled by default.

In code, all functionality that's related to new agentic chat UX was moved behind this flag. All functionality that's added and should only be used in agentic chat must be set behind this flag as well.

## License


chat-client/src/client/mynahUi.ts
```diff
@@ -439,7 +454,9 @@ export const createMynahUi = (
             throw new Error(`Unhandled tab bar button id: ${buttonId}`)
         },
         onPromptInputOptionChange: (tabId, optionsValues) => {
-            handlePromptInputChange(mynahUi, tabId, optionsValues)
+            if (agenticMode) {
```
volodkevych, line 457: nit: it's probably no needed here - this button for pair programming should never be shown for non agentic mode

viktorsaws: yes, but I added it just in case something triggers it and messes up prompt with pair programming related stuff

chat-client/src/client/mynahUi.ts
```diff
@@ -649,30 +687,51 @@ export const createMynahUi = (
               }
             : {}
 
-        mynahUi.endMessageStream(tabId, chatResult.messageId ?? '', {
```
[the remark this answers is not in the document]
viktorsaws: I considered that, but there's still a lot of shared code, which we can forget to handle in one of variants. I could switch to that, I don't have preference

volodkevych (approved):

PR #1177, opened 2025-04-28
fix: fix execute command header issues

bywang56:
## Problem
1. Execute bash headers flickers during command execution
2. Shell header incorrectly added to other messages
## Solution
Fixed both

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -925,27 +925,52 @@ export class AgenticChatController implements ChatHandlers {
         if (toolUse.name !== 'executeBash') {
             return
         }
+
+        const toolMsgId = toolUse.toolUseId!
+        const chatMsgId = chatResultStream.getResult().messageId
+        let headerEmitted = false
+
+        const initialHeader: ChatMessage['header'] = {
+            body: 'shell',
+            status: { status: 'success', icon: 'ok', text: '' },
+            buttons: [{ id: 'stop-shell-command', text: 'Stop', icon: 'stop' }],
+        }
+
         const completedHeader: ChatMessage['header'] = {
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -925,28 +925,53 @@ export class AgenticChatController implements ChatHandlers {
...
                     body: '```',
-                    messageId: toolUse.toolUseId,
                     header: completedHeader,
                 })
+
+                await chatResultStream.writeResultBlock({
+                    type: 'answer',
+                    messageId: chatMsgId,
+                    body: '',
+                    header: undefined,
```
Hweinstock, line 973: Out of curiosity, why do we need to pass this as explicitly undefined?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
```diff
@@ -131,7 +132,7 @@ export class AgenticChatResultStream {
                                         },
                                     },
                                 }),
-                            header: c.header ? { ...c.header } : { ...am.header },
+                            header: Object.prototype.hasOwnProperty.call(c, 'header') ? c.header : am.header,
```
Hweinstock, line 135: in what cases would these two checks be different?

Hweinstock (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
bywang56: What we need here is pass a specific undefined value to header so
Old (c.header ? …) treats header: undefined as “missing” (falsy) and falls back to the default.
New (hasOwnProperty('header')) sees the explicit header key and preserves the undefined override.

PR #1179, opened 2025-04-28
fix: loading appears too often

ctlai95:
## Problem
Issues with `Thinking...` flickering and appearing when waiting on user acceptance.

## Solution
- Moved the logic to show the message back to the controller. Now it will show the loading message right before every model request.
- Added a method to check if a toolUseEvent stream is taking too long, in that case we want to show an additional `Thinking...` message

TODO: 
- If there is already a loading message when the timeout triggers, it causes a single flicker. We probably need to check if the loading already exists or not.
- Add `answer-stream` type to runtimes package

https://github.com/user-attachments/assets/45ee648a-929b-4da5-bd38-a99d0ff8652f

## License


zixlin7 (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1813,6 +1831,41 @@ export class AgenticChatController implements ChatHandlers {
         return chatEventParser.getResult()
     }
 
+    /**
+     * This is needed to handle the case where a toolUseEvent takes a long time to resolve the stream and looks like
+     * nothing is happening.
+     */
+    async #showLoadingIfRequired(
```
Hweinstock, line 1838: I appreciate that you isolated this logic! If I understand correctly, we will hopefully be able to remove this in the future once the backend resolves tool use faster.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
```diff
@@ -183,6 +183,10 @@ export class AgenticChatResultStream {
 
         return {
             write: async (intermediateChatResult: ChatMessage) => {
+                const isLoading = intermediateChatResult.messageId?.startsWith('loading-')
```
Hweinstock, line 186: could be worth mentioning this behavior in a doc string. Not obvious to the caller that messageIds prefixed by `loading-` are handled separately.

laileni-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
ctlai95: That's right, I think if we start seeing that the timeout starts resolving itself then we can remove this

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatResultStream.ts
ctlai95: Sure, updating

zixlin7 (approved):

PR #1183, opened 2025-04-29
fix: override the MynahUi maxUserInput config

grl-wang:
## Problem
there is bug in the input prompt popup UI to display the max prompt chars size as 4000. 
image
This is cause as use the default flare chat ui config as following. 

```
  userInputLengthWarningThreshold: 3500,
  maxUserInput: 4096,
```

## Solution
Fix this issue by override the default flare chat ui config in the language server package. verified with a large prompt
image

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -39,7 +39,6 @@ import {
     NotificationType,
     MynahUIProps,
     QuickActionCommand,
-    ChatItemFormItem,
     ChatItemButton,
 } from '@aws/mynah-ui'
 import { VoteParams } from '../contracts/telemetry'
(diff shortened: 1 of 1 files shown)


chat-client/src/client/mynahUi.ts
```diff
@@ -474,6 +468,11 @@ export const createMynahUi = (
         },
         config: {
             maxTabs: 10,
+            // RTS max user input is 600k, we need to leave around 500 chars to user to type the question
+            // beside, MynahUI will automatically crop it depending on the available chars left from the prompt field itself by using a 96 chars of threshold
```
ctlai95, line 472: just curious, is this documented somewhere?

aditya169: I think its this (but please verify) https://code.amazon.com/packages/AWSVectorConsolasPlatformModel/blobs/a6045cab97739954a74b35a6e3ff061c322f4555/--/model/types/conversation_types.smithy#L77

grl-wang: the flare chat UI config and document: 

https://tiny.amazon.com/9q5acyyp/defaultchatUIconfig
https://tiny.amazon.com/hzyugvya/reduce96frommaxuserinput

RTS max user input size
https://tiny.amazon.com/16x05fxos/maxsizeuserinput

ctlai95 (approved):

chat-client/src/client/mynahUi.ts
dogusata: Yes, https://github.com/aws/mynah-ui/blob/main/docs/CONFIG.md#maxuserinput

PR #1184, opened 2025-04-29
fix: file should be grey out and unclickable after undo

Randall-Jiang:
## Problem
When user click undo/undo all, the view diff still clickable and visible.
## Solution

https://github.com/user-attachments/assets/6ca9ec24-f382-433a-b0ba-da9ca81dc2a1

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -249,19 +249,44 @@ export class AgenticChatController implements ChatHandlers {
         if (!cachedToolUse) {
             return
         }
+        const fileList = cachedToolUse.chatResult?.header?.fileList
+
+        const updatedHeader = {
+            ...cachedToolUse.chatResult?.header,
+            buttons: cachedToolUse.chatResult?.header?.buttons?.filter(button => button.id !== 'undo-changes'),
```
laileni-aws, line 256: nit:
`const buttons = cachedToolUse.chatResult?.header?.buttons?.filter(button => button.id !== 'undo-changes')`
and assign this here for Readability.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -249,19 +249,44 @@ export class AgenticChatController implements ChatHandlers {
...
                     {
                         ...cachedToolUse.chatResult,
-                        header: {
-                            ...cachedToolUse.chatResult?.header,
-                            buttons: cachedToolUse.chatResult?.header?.buttons?.filter(
-                                button => button.id !== 'undo-changes'
-                            ),
-                            status: { status: 'error', icon: 'cancel', text: 'Change discarded' },
-                        },
+                        header: updatedHeader,
```
laileni-aws, line 289: Please test if this is causing flickering UX? If not we should be good

laileni-aws:
If user clicked on `undo` and if user has diff open, IDE need to close the diff view ux.
Can you include this change?

jpinkney-aws:
Looks like the build is failing because of "Object literal may only specify known properties, and 'clickable' does not exist in type 'FileDetails'."

laileni-aws (approved):
LGTM

tsmithsz (approved):

PR #1189, opened 2025-04-29
fix: add profileArn for STE and fix timeBetweenChunks

zixlin7:
## Problem
STE requests are failing for addMessage because:
- we are not passing profileArn
- timeBetweenChunks has a limit of 100

## Solution
- add profileArn
- cap timeBetweenChunks length

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
@@ -414,6 +414,7 @@ export class TelemetryService {
         if (!params.conversationId || !params.messageId) {
             return
         }
+        const timeBetweenChunks = params.timeBetweenChunks?.slice(0, 100)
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
```diff
@@ -406,6 +414,7 @@ export class TelemetryService {
         if (!params.conversationId || !params.messageId) {
             return
         }
+        const timeBetweenChunks = params.timeBetweenChunks?.slice(0, 99)
```
jpinkney-aws, line 417: does this include 99 or 100?

zixlin7: haha I think you are right this is 99

jpinkney-aws (approved):

PR #1190, opened 2025-04-29
feat: implement new sendToPrompt params

jpinkney-aws:
## Problem
we have no way to send an event from the client side and trigger it in flare. This is needed for events like "explain issue" from VSCode's security scan, which triggers a event that gets displayed to the user, and a separate event that gets sent to the backend

## Solution
wire up an autosubmit event that when enabled takes a prompt from sendToPrompt and automatically submits it

**note**: the extended types were for testing, if this API is agreed upon i'll make the changes in the runtimes package

Related VSCode PR: https://github.com/aws/aws-toolkit-vscode/pull/7194

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -131,7 +131,11 @@ export const handleChatPrompt = (
         const context = prompt.context?.map(c => (typeof c === 'string' ? { command: c } : c))
         messager.onChatPrompt({ prompt, tabId, context }, triggerType)
     }
-    // Add user prompt to UI
+
+    initializeChatResponse(mynahUi, tabId, userPrompt)
+}
+
+const initializeChatResponse = (mynahUi: MynahUI, tabId: string, userPrompt?: string) => {
     mynahUi.addChatItem(tabId, {
         type: ChatItemType.PROMPT,
         body: userPrompt,
@@ -772,7 +776,12 @@ export const createMynahUi = (
         const tabId = getOrCreateTabId()
         if (!tabId) return
 
-        mynahUi.addToUserPrompt(tabId, params.selection, 'code')
+        if (params.autoSubmit && params.prompt) {
+            messager.onChatPrompt({ prompt: params.prompt, tabId, context: undefined }, 'contextMenu')
+            initializeChatResponse(mynahUi, tabId, params.prompt.prompt)
+        } else {
+            mynahUi.addToUserPrompt(tabId, params.selection, 'code')
+        }
         messager.onSendToPrompt(params, tabId)
     }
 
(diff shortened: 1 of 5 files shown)


chat-client/src/client/mynahUi.ts
```diff
@@ -777,11 +775,16 @@ export const createMynahUi = (
...
 
-    const sendToPrompt = (params: SendToPromptParams) => {
+    const sendToPrompt = (params: SendToPromptParams & { autoSubmit?: boolean; prompt: ChatPrompt }) => {
         const tabId = getOrCreateTabId()
         if (!tabId) return
 
-        mynahUi.addToUserPrompt(tabId, params.selection, 'code')
+        if (params.autoSubmit) {
+            messager.onChatPrompt({ prompt: params.prompt, tabId, context: undefined }, 'contextMenu')
+            initializeChatResponse(mynahUi, tabId, params.prompt.prompt)
```
jpinkney-aws, line 784: I wasn't able to figure out how to have the code selection auto submit, so I just sent mine along with the prompt

jpinkney-aws: The problem is that for explain code one prompt is sent to the user and one is sent to the backend. In the vscode implementation we're sending prompt to the user and escaped prompt to the backend

Hweinstock: whats the difference between the two prompt fields?

jpinkney-aws: escapedPrompt typically just trims the start/end of the string AFAICT

Hweinstock: is there any reason to directly use `params.prompt.prompt` then?

Hweinstock: Oh, I see this is because we are using fundamentally different values here: https://github.com/aws/aws-toolkit-vscode/pull/7194/files, I'm assuming this is the first time we want to split where we send the prompt. Yeah, I don't see a better way to handle this, but maybe worth building this support into the chat client more naturally.

jpinkney-aws: It's basically a hack, we're showing `params.prompt.prompt` to the user and then agentic chat controller sends `params.prompt.escapedPrompt` to the backend. The prompt that goes to the user is just: "Explain the issue with x,y,z" and the prompt that goes to the backend sends like 60 extra lines to the backend with a bunch more context. That's why its seperated

Hweinstock (approved):
the separate escaped prompt and prompt seems slightly hacky, but handling this naturally sounds like a large refactor of chat-client, so approving to ensure this makes it into bug bash.

leigaol:
The escaped prompt translates ' into &#39

PR #1201, opened 2025-04-29
feat: use enableLocalIndexing to control actual indexing

zixlin7:
## Problem
- for some features that doesn't depend on vector indexing, we should still initialize the library but have the setting to control if we build vector index
## Solution
- refactor to always call `init` and add check for if we need to call `buildIndex`

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
@@ -151,8 +151,6 @@ export class LocalProjectContextController {
             }
 
             // initialize vecLib and index if needed
-            this._isIndexingEnabled = enableIndexing
-
             const libraryPath = this.getVectorLibraryPath()
             const vecLib = vectorLib ?? (await eval(`import("${libraryPath}")`))
             if (vecLib) {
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -219,7 +234,7 @@ export class LocalProjectContextController {
                     merged.push(addition)
                 }
             }
-            if (this._vecLib) {
+            if (this._vecLib && this._isIndexingEnabled) {
                 await this.buildIndex()
```
breedloj, line 238: Let's make this `void` as well.

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -140,11 +141,25 @@ export class LocalProjectContextController {
...
+            if (this._vecLib) {
+                if (enableIndexing && !this._isIndexingEnabled) {
+                    void this.buildIndex()
+                }
+                this._isIndexingEnabled = enableIndexing
+                return
+            }
+
+            // initialize vecLib and index if needed
+            this._isIndexingEnabled = enableIndexing
```
breedloj, line 154: Shouldn't this assignment happen after we attempt to build the index? Otherwise a failure in the import or start calls would effectively prevent us from subsequently building the index.

zixlin7: updated

breedloj (approved):

jpinkney-aws (approved):

PR #1205, opened 2025-04-29
fix: fix project root not passed to buildIndex

zixlin7:
## Problem
`indexCacheDirPath` was passed to buildIndex incorrectly instead of project root path
this will result in index of different projects overwrite each other, @workspace might respond with a different project's code
## Solution
pass in project root

## License


leigaol (approved):
the projectRoot parameter is only used to create a hash of the indexing cache dir so that each workspace has its own indexing cache. Without this fix, we are putting all workspace index cache in exact same file, which breaks.

PR #1210, opened 2025-04-30
fix: add support for determing workspace folder with root uri/path on initialize

shruti0085:
## Description
LSP lifecycle for VS extension is controlled by Visual Studio. It currently does not set the `workspaceFolder` as part of the `Initialize Params` of the LSP initialize request, however it does send this information via `rootUri` and `rootPath` properties. The workspaceFolders is utilized for the local project context indexing and so in order to support that, this change extends logic to determine the workspaceFolder to use the root properties when workspaceFolders is not set.

## License


server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.ts
```diff
@@ -0,0 +1,23 @@
+import { InitializeParams, WorkspaceFolder } from '@aws/language-server-runtimes/protocol'
+import * as path from 'path'
+import { URI } from 'vscode-uri'
+
+export function getWorkspaceFolders(params?: InitializeParams): WorkspaceFolder[] {
+    if (!params) {
+        return []
+    }
+    const getFolderName = (uri: string) => path.basename(URI.parse(uri).path) || 'workspace'
+    if (params.workspaceFolders) {
```
awschristou, line 10: Also check if the array is empty, so that we avoid returning an empty list too early

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.ts
```diff
@@ -0,0 +1,23 @@
+import { InitializeParams, WorkspaceFolder } from '@aws/language-server-runtimes/protocol'
+import * as path from 'path'
+import { URI } from 'vscode-uri'
+
+export function getWorkspaceFolders(params?: InitializeParams): WorkspaceFolder[] {
```
awschristou, line 5: Should this method be wrapped in a try catch, returning an empty array if there is an error? In cases like VS, the inputs here would be out of the extension and user's control but we may not want the server to crash (or fail to initialize).

shruti0085: Updated

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.ts
```diff
@@ -0,0 +1,29 @@
+import { InitializeParams, WorkspaceFolder } from '@aws/language-server-runtimes/protocol'
+import { Logging } from '@aws/language-server-runtimes/server-interface'
+import * as path from 'path'
+import { URI } from 'vscode-uri'
+
+export function getWorkspaceFolders(logging: Logging, params?: InitializeParams): WorkspaceFolder[] {
```
justinmk3, line 6: why is this in localProjectContext? instead of shared/ ? seems generally applicable , not specific to project context.

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+    try {
+        if (!params) {
```
ege0zcan, line 8: nit: we can move the early return statements outside of the try catch block
```
if(!params){
   return []
}
if (params.workspaceFolders && params.workspaceFolders.length > 0) {
   return params.workspaceFolders
}

try{
...
```

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+            return []
+        }
+        const getFolderName = (uri: string) => path.basename(URI.parse(uri).fsPath) || 'workspace'
```
ege0zcan, line 11: why do we default to `workspace` here? Couldn't it lead to some unexpected side effects?

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+            return [{ name: folderName, uri: params.rootUri }]
+        }
+        if (params.rootPath) {
+            const pathUri = URI.parse(params.rootPath).toString()
+            const folderName = getFolderName(pathUri)
+            return [{ name: folderName, uri: pathUri }]
+        }
+        return []
+    } catch (error) {
+        logging.error(`Error occurred when determine workspace folders: ${error}`)
```
ege0zcan, line 26: typo: `when determine`

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.test.ts
```diff
@@ -0,0 +1,126 @@
...
+                sampleWorkspaceFolder,
+                { name: 'folder2', uri: 'file:///path/to/folder2' },
+            ]
+            const params = createParams({ workspaceFolders })
+            const result = getWorkspaceFolders(logging, params)
+
+            assert.deepStrictEqual(result, workspaceFolders)
+        })
+
+        describe('should create workspace folder from rootUri when workspaceFolders is not provided', () => {
```
ege0zcan, line 39: is this intentionally `describe` and not `it` ?

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.test.ts
```diff
@@ -0,0 +1,126 @@
...
+            const folderName = path.basename(decodedPath)
+
+            const params = createParams({ rootUri })
+            const result = getWorkspaceFolders(logging, params)
+
+            assert.deepStrictEqual(result, [{ name: folderName, uri: rootUri }])
+            assert.equal('special project', result[0].name)
+        })
+
+        describe('should return empty workspaceFolder array', () => {
```
ege0zcan, line 97: same as above, is the `describe` intentional?

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.ts
shruti0085: just as a default incase a folder name cannot be extracted. I can default it to the full path if that makes more sense

shruti0085: Updated to return the uri instead as default

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/initializeUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+        const getFolderName = (uri: string) => path.basename(URI.parse(uri).fsPath) || 'workspace'
+        if (params.workspaceFolders && params.workspaceFolders.length > 0) {
+            return params.workspaceFolders
+        }
+        if (params.rootUri) {
+            const folderName = getFolderName(params.rootUri)
+            return [{ name: folderName, uri: params.rootUri }]
+        }
+        if (params.rootPath) {
+            const pathUri = URI.parse(params.rootPath).toString()
```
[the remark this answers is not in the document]
shruti0085: Reduced the duplication and refactored it a bit

ege0zcan (approved):

PR #1212, opened 2025-04-30
fix: prefix if user reject/stop command, whole card should be dimmed

Randall-Jiang:
## Problem
when user reject/stop command, all other text in the command should also be disabled color. 
image

## Solution
image

## Notice
This is a prefix, the fix won't work until mynah release the newest version

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -621,9 +621,7 @@ export const createMynahUi = (
                             : am.type === 'directive'
                               ? ChatItemType.DIRECTIVE
                               : ChatItemType.ANSWER_STREAM,
-                    ...prepareChatItemFromMessage(am, isPairProgrammingMode, isPartialResult),
-                    // Apply muted if the message should be muted
-                    ...(shouldMute ? { muted: true } : {}),
+                    ...prepareChatItemFromMessage(am, isPairProgrammingMode, isPartialResult, shouldMute),
                 }
 
                 if (!chatItems.find(ci => ci.messageId === am.messageId)) {
@@ -676,8 +674,6 @@ export const createMynahUi = (
                 ...chatResultWithoutType, // type for MynahUI differs from ChatResult types so we ignore it
                 header: header,
                 buttons: buttons,
-                // Apply muted if the message should be muted
-                ...(shouldMute ? { muted: true } : {}),
             })
 
             // TODO, prompt should be disabled until user is authenticated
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1080,20 +1080,20 @@ export class AgenticChatController implements ChatHandlers {
...
                               status: {
-                                  status: 'error',
                                   icon: 'cancel',
                                   text: 'Rejected',
                               },
                           }),
                     buttons: isAccept ? [{ id: 'stop-shell-command', text: 'Stop', icon: 'stop' }] : [],
                 },
-            }
+            } as any
```
volodkevych, line 1096: We should not be doing it. `muted` has to be in `ChatResult` explicitly, otherwise there's no interface anymore, we go back to Mynah types.

ctlai95: +1, no `any` please. I think we need to do a conditional check in mynahUi instead of passing in a muted value here

chat-client/src/client/mynahUi.ts
```diff
@@ -618,6 +622,8 @@ export const createMynahUi = (
                               ? ChatItemType.DIRECTIVE
                               : ChatItemType.ANSWER_STREAM,
                     ...prepareChatItemFromMessage(am, isPairProgrammingMode, isPartialResult),
+                    // Apply muted if the message should be muted
+                    ...(shouldMute ? { muted: true } : {}),
```
ctlai95, line 626: Shouldn't this be inside `prepareChatItemFromMessage`?

chat-client/src/client/mynahUi.ts
```diff
@@ -822,11 +834,18 @@ export const createMynahUi = (
...
+                const shouldMute =
+                    // Check for status indicating stopped/rejected
+                    updatedMessage.header?.status?.text === 'Stopped' ||
+                    updatedMessage.header?.status?.text === 'Rejected'
+
                 const chatItem: ChatItem = {
                     type: oldMessage.type,
                     ...prepareChatItemFromMessage(updatedMessage, getTabPairProgrammingMode(mynahUi, tabId)),
+                    // Apply muted if the message should be muted
+                    ...(shouldMute ? { muted: true } : {}),
```
ctlai95, line 847: Same here, if this can be moved to `prepareChatItemFromMessage` then we can simplify this

ctlai95 (approved):

PR #1215, opened 2025-04-30
feat: restore open history tabs from client state after process crash

viktorsaws:
## REVERT THIS BEFORE MERGING:

This PR contains code wit example quick command to crash nodejs process for testing. Revert this commit before merging https://github.com/aws/language-servers/pull/1215/commits/df410911e9e2a6d189f68294de5cac2c4fc7964d

## Problem

When language server process crashes, it loses in-memory cache of mapping UI tabs to history IDs. When extension restarts language server automatically, chat client UI stays loaded, but History server does not have mapping to open tabs anymore. This results in inconsistent UI, when using chat after crash:
1. Opening history items from list will always open new UI tab, even if one was opened before crash
2. Sending messages to old open tabs are not stored in history anymore

## Solution

Potential fix: maintain mapping between history items and ui tabs in Chat client, and restore history state in language server process memory after restart. Сhat client can detect that server reconnected when it receives inbound "CHAT_OPTIONS" notification more than once: as CHAT_OPTIONS should be sent from language server once during initialization handshake.

1. Added historyId to tabId mapping in Chat client
2. Added notification to send cached state from chat client to language server when more than one CHAT_OPTIONS request is received
3. Added new `chat.onRestoreState` notification from Chat client to agenticQChat server to restore it's in-memory state from client state data. We can't use `chat.onReady` event, as it is sent once from Chat Client to server during Chat Client and MynahUI initialization. Since Chat Client remains loaded in webview, this event does not happen second time.
4. Extended `chat.openTab` request from server to client to include `historyId` string. History handling server  needs to send `historyId` so chat client can maintain it's own tab to history map.

WIP: to test this E2E, grab wip protocol changes from https://github.com/aws/language-server-runtimes/pull/491 and build runtimes package locally.

#### UX without fix

https://github.com/user-attachments/assets/65c4ce1f-f371-4222-a963-0a379623b300

#### UX with fix

https://github.com/user-attachments/assets/8c817ddf-751e-4b31-a3fb-d966166e3c78

### Alternative

Evaluate if complexity with gracefully restoring state is justified - potentially there might be edge cases that are not handled with this PR. Thorough testing and review is needed.

Alternative solution could be for extension to detect language server process crash, show informative message to customer on language server crash and reload Chat Client in webview together with Language server.

## WIP notes
* [ ] Needs thorough E2E testing
* [ ] New protocol changes (chat.onRestoreState) and in (chat.openTab) need to be added to Runtimes protocol https://github.com/aws/language-server-runtimes/pull/491
* [ ] Needs unit tests
* [ ] Revert commit that adds "/crash" quick action for triggering process exit https://github.com/aws/language-servers/pull/1215/commits/df410911e9e2a6d189f68294de5cac2c4fc7964d

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -112,6 +112,7 @@ import { CodeSearch } from './tools/codeSearch'
 import { genericErrorMsg, maxAgentLoopIterations, loadingThresholdMs } from './constants'
 import { URI } from 'vscode-uri'
 import { AgenticChatError } from './errors'
+import { HistoryTabMapping } from './qAgenticChatServer'
 
 type ChatHandlers = Omit<
     LspHandlers<Chat>,
@@ -155,7 +156,7 @@ export class AgenticChatController implements ChatHandlers {
         chatSessionManagementService: ChatSessionManagementService,
         features: Features,
         telemetryService: TelemetryService,
-        serviceManager?: AmazonQTokenServiceManager
+        serviceManager: AmazonQTokenServiceManager
     ) {
         this.#features = features
         this.#chatSessionManagementService = chatSessionManagementService
@@ -1671,8 +1672,13 @@ export class AgenticChatController implements ChatHandlers {
 
     onLinkClick() {}
 
-    async onReady() {
-        await this.#tabBarController.loadChats()
+    /**
+     * Initialize server and client state as part of initial initialization process.
+     * Can be called on initial loading or from initialized chat client after server crash.
+     */
+    async initializeServices(clientState?: any) {
+        await this.#tabBarController.loadChats(clientState)
+
         try {
             const localProjectContextController = await LocalProjectContextController.getInstance()
             const contextItems = await localProjectContextController.getContextCommandItems()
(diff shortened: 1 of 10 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1683,6 +1689,22 @@ export class AgenticChatController implements ChatHandlers {
...
+    }
+
+    /**
+     * Restore state onRestoreState event is sent from client to server.
+     * Client sends this notification when reconnection to the server is detected, ex. after server process crash.
+     * Server can gracefully restore it's own in-memory state or update client if changes were detected after restart.
+     */
+    async onRestoreState(clientState: { historyTabMapping?: HistoryTabMapping }) {
+        // Wait how history database to fully load to prevent potential race condition during language server restart after crash.
+        await this.#chatHistoryDb.waitForDbInitialization()
```
volodkevych, line 1703: Why don't we have the same issue for `onReady`?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tabBarController.ts
```diff
@@ -245,6 +270,20 @@ export class TabBarController {
...
+     * Restore TabId and HistoryId mapping in server memory from client session data.
+     * Used only if server has not loaded chats, but UI was loaded, e.g. after server crash.
+     */
+    async restoreChatsFromClientState(clientState: { historyTabMapping: HistoryTabMapping }) {
+        for (const [historyId, tabId] of Object.entries(clientState.historyTabMapping)) {
+            const selectedTab = this.#chatHistoryDb.getTab(historyId)
+            if (selectedTab) {
+                this.#chatHistoryDb.setHistoryIdMapping(tabId, historyId)
+                this.#chatHistoryDb.updateTabOpenState(tabId, true)
+            }
```
volodkevych, line 283: nit: warn if not there?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tabBarController.ts
```diff
@@ -230,11 +239,27 @@ export class TabBarController {
...
-    async loadChats() {
+    async loadChats(clientState?: any) {
         if (this.#loadedChats) {
+            this.#features.logging.log('loadChats: Chats already loaded')
             return
         }
+
         this.#loadedChats = true
+
+        if (clientState) {
```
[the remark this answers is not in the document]
viktorsaws: initially I added 2 method, but than decided to merge them. My thinking is that we need to flip `loadedChats` flag and have a check for both, so that chats are loaded only once during language server lifecycle. Otherwise, the method that restores from client state may keep overriding state, if called several times.

viktorsaws:
Decided to delegate graceful handling of chat reload to extensions

PR #1220, opened 2025-04-30
fix: fix for permission case for execute bash

ashishrp-aws:
## Problem
In default cases, Shell execution outside of workspace is not showing warning.

## Solution
- Continue in default case and not return requiresAcceptance value.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
@@ -216,8 +216,17 @@ export class ExecuteBash {
                 // For each command, validate arguments for path safety within workspace
                 for (const arg of cmdArgs) {
                     if (this.looksLikePath(arg)) {
-                        // If not absolute, resolve using workingDirectory if available.
-                        const fullPath = !isAbsolute(arg) && params.cwd ? join(params.cwd, arg) : arg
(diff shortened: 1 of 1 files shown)


laileni-aws (approved):

chungjac (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -216,8 +216,17 @@ export class ExecuteBash {
                 // For each command, validate arguments for path safety within workspace
                 for (const arg of cmdArgs) {
                     if (this.looksLikePath(arg)) {
-                        // If not absolute, resolve using workingDirectory if available.
-                        const fullPath = !isAbsolute(arg) && params.cwd ? join(params.cwd, arg) : arg
+                        // Special handling for tilde paths in Unix-like systems
+                        let fullPath: string
+                        if (!IS_WINDOWS_PLATFORM && arg.startsWith('~')) {
```
justinmk3, line 221: why is this decision not encapsulated in `isAbsolute()` ? do we have a shared module that extends the default node 'path module ?

ashishrp-aws: isAbsolute from path dependency is not catching this path and the current logic is adding to create an absolute path like `/Volumes/user/workplace/~/Desktop/test`,

PR #1227, opened 2025-05-01
feat: adding a check before invoking workspace context

suprajaven:
## Solution

* Added a check in files belonging to non agentic chat before invoking the workspace context
* Workspace context has a 60 second wait time even on cases where the context is not required, using the workspace config flag to add a check before invoking it.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
@@ -6,6 +6,7 @@ import { DocumentContext, DocumentContextExtractor } from './documentContext'
 import { SendMessageCommandInput } from '../../../shared/streamingClientService'
 import { LocalProjectContextController } from '../../../shared/localProjectContextController'
 import { convertChunksToRelevantTextDocuments } from '../tools/relevantTextDocuments'
+import { AmazonQBaseServiceManager as AmazonQServiceManager } from '../../../shared/amazonQServiceManager/BaseAmazonQServiceManager'
 
 export interface TriggerContext extends Partial<DocumentContext> {
     userIntent?: UserIntent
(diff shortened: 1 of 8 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
```diff
@@ -120,6 +125,19 @@ export class QChatTriggerContext {
     async extractProjectContext(query?: string): Promise<RelevantTextDocument[]> {
         if (query) {
             try {
+                let enableWorkspaceContext = true
+
+                if (this.amazonQServiceManager) {
+                    const config = this.amazonQServiceManager.getConfiguration()
+                    if (config.projectContext?.enableLocalIndexing === false) {
```
imykhai, line 132: It's better to have this new workspace configuration documented in readme

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/crossFileContextUtil.ts
```diff
@@ -143,6 +148,17 @@ export async function fetchProjectContext(
...
+        if (config.projectContext?.enableLocalIndexing === false) {
+            enableWorkspaceContext = false
+        }
+    }
+    if (!enableWorkspaceContext) {
+        return []
+    }
 
     try {
         controller = await LocalProjectContextController.getInstance()
```
ege0zcan, line 164: Just want to make sure again: is it possible to move this check to the `localProjectContextController.getInstance` function? 

Something like

```
public static async getInstance(): Promise<LocalProjectContextController> {
     const config = amazonQServiceManager.getConfiguration()
     if (config.projectContext?.enableLocalIndexing === false) {
          return {
             isEnabled: true,
             getContextCommandItems: () => [],
             shouldUpdateContextCommandSymbolsOnce: () => false,
         } as unknown as LocalProjectContextController
     }

     try {
            await waitUntil(async () => this.instance, {...})
        ....
        } catch (error) {
           ...
        }
    }
    
 ```

suprajaven: Since this is a static function - I cannot access the amazonQServiceManager from a static method unless it's passed as a parameter/ static property of the class. I tried the parameter way but it was very similar to this implementation and I thought I could avoid an extra hop to this method.

server/aws-lsp-codewhisperer/src/language-server/chat/contexts/triggerContext.ts
suprajaven: This config is already present, I am not introducing a new workspace config here. But ack, can add readme.

ege0zcan (approved):

breedloj (approved):

PR #1228, opened 2025-05-01
chore: add a step in the github workflow to test the packaged bundle

pras0131:
## Problem

Adding another step in the Github workflow to test the packaged bundle, to catch up issues where server crashes on starting. This will help in catching such issues earlier. 

## License

### DIFF
--- a/app/aws-lsp-codewhisperer-runtimes/scripts/test-bundles.js
+++ b/app/aws-lsp-codewhisperer-runtimes/scripts/test-bundles.js
@@ -52,7 +52,7 @@ async function testAllBundles() {
             await testBundle(path.join(buildDir, file))
             console.info(`✓ ${file} test passed`)
         } catch (error) {
-            console.info(`✗ ${file} test failed:`, error)
+            console.error(`✗ ${file} test failed:`, error)
             process.exit(1)
         }
     }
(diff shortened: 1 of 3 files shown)


app/aws-lsp-codewhisperer-runtimes/scripts/test-bundles.js
```diff
@@ -0,0 +1,64 @@
...
+                reject(new Error(`Test failed with exit code ${code}`))
+            }
+        })
+
+        serverProcess.on('spawn', () => {
+            console.info('Spawn called')
+            // Wait for 10s, then close the process in case something taking long time to process, which might cause an error
+            startupTimeout = setTimeout(() => {
+                if (process.platform === 'win32') {
+                    spawn('taskkill', ['/pid', serverProcess.pid, '/f', '/t'])
```
ege0zcan, line 36: I guess this specific windows logic is because of the following:

```
On Windows, where POSIX signals do not exist, the signal argument will be ignored except for 'SIGKILL', 'SIGTERM', 'SIGINT' and 'SIGQUIT', and the process will always be killed forcefully and abruptly (similar to 'SIGKILL'). See [Signal Events](https://nodejs.org/api/process.html#signal-events) for more details.
```
https://nodejs.org/api/child_process.html#subprocesskillsignal

Would be useful to comment it in the code so future people are not confused when they see this.

ege0zcan (approved):

app/aws-lsp-codewhisperer-runtimes/scripts/test-bundles.js
pras0131: yes, I'll add the comment as a follow up.

PR #1229, opened 2025-05-01
fix: hide stop button in legacy chat client mode

viktorsaws:
## Problem
Stop button is shown when chat client is not in agenticMode. This is not in line with legacy mode for extensions that didn't migrate to Agentic chat

## Solution
Do not show stop button in mynahUI in legacy mode

## License


chat-client/src/client/mynahUi.ts
```diff
@@ -150,6 +150,7 @@ const initializeChatResponse = (mynahUi: MynahUI, tabId: string, userPrompt?: st
     } else {
         mynahUi.updateStore(tabId, {
             loadingChat: true,
+            cancelButtonWhenLoading: false,
```
volodkevych, line 153: Shouldn't it be: `cancelButtonWhenLoading: this.agenticMode`?

PR #1236, opened 2025-05-01
fix: switch to ignore entries over patterns

Hweinstock:
## Problem
The `readDirectoryRecursively` utility was using the exclude directories as patterns. This results in the listDirectory and fileSearch tool ignoring files that are prefixed by the pattern. For example, `build-test` would be ignored, or a file named `output.txt` since it contains `"out"`. 

## Solution
- use explicit ignore entries that apply at the entry level rather than the entire path level. 
- add a logging statement for when a directory is skipped. 

## Verification
- added a unit test 
- created a demo folder to test e2e:
image

- This folder also contains a `node_modules` folder with some files in it, which are ignored. 

## License

### DIFF
--- a/core/aws-lsp-core/src/util/workspaceUtils.ts
+++ b/core/aws-lsp-core/src/util/workspaceUtils.ts
@@ -58,8 +58,7 @@ export async function readDirectoryRecursively(
         }
         for (const entry of entries) {
             const childPath = getEntryPath(entry)
-            const childPathParts = childPath.split(path.sep)
-            if (childPathParts.some(part => excludeEntries.includes(part))) {
+            if (excludeEntries.includes(entry.name)) {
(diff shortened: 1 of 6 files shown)


core/aws-lsp-core/src/util/workspaceUtils.ts
```diff
@@ -58,7 +58,9 @@ export async function readDirectoryRecursively(
         }
         for (const entry of entries) {
             const childPath = getEntryPath(entry)
-            if (options?.excludePatterns?.some(pattern => new RegExp(pattern).test(childPath))) {
+            const childPathParts = childPath.split(path.sep)
```
jguoamz, line 61: can't we just just do

`entry.name` and compare this against the pattern?

jpinkney-aws (approved):

PR #1240, opened 2025-05-02
fix: fix the extra line on executeBash and wrong warning msg for outs…

Randall-Jiang:
…ide workspace

## Problem
There was an extra line after command in executeBash and the wrong warning message for normal command outside the current workspace
image

## Solution
Screenshot 2025-05-01 at 6 12 26 PM

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -1257,7 +1257,7 @@ export class AgenticChatController implements ChatHandlers {
                     buttons,
                 }
                 const commandString = (toolUse.input as unknown as ExecuteBashParams).command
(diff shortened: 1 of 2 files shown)


Hweinstock (approved):

Hweinstock:
@dogusata As in the warning messages should be a tooltip? How exactly is that done?

dogusata:
> @dogusata As in the warning messages should be a tooltip? How exactly is that done?

Yes, as I remember, UX requirement was that way. Just showing those warning messages as a tooltip on warning icon. It is possible though. Instead of using icon, using `status` with props as follows:

```
header:{
  ...
  status:{
    status: 'warning',
    icon: 'warning',
    description: 'WARNING MESSAGE',
    position: 'left'
  },
  ...
}
...
```

Randall-Jiang:
@dogusata there was no one point out in past few days' bug bash.....

PR #1241, opened 2025-05-02
fix: show context transparency list when using @workspace

avi-alpert:
## Problem
When a user enters `@workspace`, a Context menu should appear in the response listing all the file chunks sent to Q

## Solution
Send relevant text documents to contextList

Screenshot 2025-05-01 at 11 03 52 PM

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextUtils.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextUtils.test.ts
@@ -2,9 +2,17 @@ import * as path from 'path'
 import * as sinon from 'sinon'
 import * as assert from 'assert'
 import { expect } from 'chai'
-import { getUserPromptsDirectory, getNewPromptFilePath, promptFileExtension } from './contextUtils'
+import {
+    getUserPromptsDirectory,
+    getNewPromptFilePath,
+    promptFileExtension,
+    mergeRelevantTextDocuments,
+    mergeFileLists,
+} from './contextUtils'
 import * as pathUtils from '@aws/lsp-core/out/util/path'
 import { sanitizeFilename } from '@aws/lsp-core/out/util/text'
+import { FileList } from '@aws/language-server-runtimes/server-interface'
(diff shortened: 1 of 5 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextUtils.ts
```diff
@@ -61,3 +62,118 @@ export const getNewPromptFilePath = (promptName: string): string => {
 
     return finalPath
 }
+
+export function mergeRelevantTextDocuments(documents: RelevantTextDocumentAddition[]): FileList {
```
avi-alpert, line 66: adapted from https://github.com/aws/aws-toolkit-vscode/blob/91f584f5df44b30a7ef188403ebc9a3dbd3161d5/packages/core/src/codewhispererChat/controllers/chat/controller.ts#L1239

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextUtils.test.ts
```diff
@@ -90,4 +98,237 @@ describe('contextUtils', () => {
...
+                },
+            }
+
+            const result = mergeFileLists(fileList1, fileList2)
+            expect(result.filePaths).to.deep.equal(['file.js'])
+            expect(result.details?.['file.js'].lineRanges).to.deep.equal([{ first: 1, second: 10 }])
+        })
+
+        it('should handle undefined lineRanges', () => {
+            const fileList1: FileList = {
```
jpinkney-aws, line 310: does it make sense to have a helper function that creates file lists for you?

avi-alpert: For the unit tests? i like how the file lists are explicitly declared "from scratch" in each test, it makes it clearer what the actual shape is and how the filePaths & details fields look. 

if there was duplicated test logic I would be more inclined to make a helper function, but its just the declaration of the FileLists themselves

PR #1246, opened 2025-05-02
Customizations profiles update

viktorsaws:
## Solution

* Added `CustomizationWithMetadata` interface that extends `Customization` with `region` and `profileArn` fields.
* Added feature flag `customizationsWithMetadata` to client capabilities for opt-in functionality. Client extension must set `initializationOptions.aws.awsClientCapabilities.q.customizationsWithMetadata` flag to opt-in into new functionality.

If `customizationsWithMetadata` flag is set, when `aws.q.customizations` section is requested with `onGetConfigurationFromServer`, server will:
1. fetch all Profiles available to customer.
2. fetch all Customizations across all fetched Profiles.
3. Return list of `CustomizationWithMetadata` objects.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
@@ -103,12 +103,17 @@ export const QConfigurationServerToken =
                                 amazonQServiceManager.getEnableDeveloperProfileSupport()
                             ) {
                                 logging.debug('Using enhanced customizations with metadata')
-                                ;[customizations, developerProfiles] = await Promise.all([
-                                    serverConfigurationProvider.listAllAvailableCustomizationsWithMetadata(token),
-                                    serverConfigurationProvider.listAvailableProfiles(token),
-                                ])
+
+                                // Fetch profiles first
+                                developerProfiles = await serverConfigurationProvider.listAvailableProfiles(token)
+
+                                // Then use those profiles to fetch customizations
+                                customizations =
+                                    await serverConfigurationProvider.listAllAvailableCustomizationsWithMetadata(
+                                        developerProfiles,
+                                        token
+                                    )
                             } else {
-                                // Use the original method for backward compatibility
                                 ;[customizations, developerProfiles] = await Promise.all([
                                     serverConfigurationProvider.listAvailableCustomizations(),
                                     serverConfigurationProvider.listAvailableProfiles(token),
@@ -125,11 +130,18 @@ export const QConfigurationServerToken =
                                 enableCustomizationsWithMetadata &&
                                 amazonQServiceManager.getEnableDeveloperProfileSupport()
                             ) {
-                                logging.debug('Using enhanced customizations with metadata for customizations section')
+                                logging.debug('Using enhanced customizations with metadata')
+
+                                // Fetch profiles first
+                                const profiles = await serverConfigurationProvider.listAvailableProfiles(token)
+
+                                // Then use those profiles to fetch customizations
                                 customizations =
-                                    await serverConfigurationProvider.listAllAvailableCustomizationsWithMetadata(token)
+                                    await serverConfigurationProvider.listAllAvailableCustomizationsWithMetadata(
+                                        profiles,
+                                        token
+                                    )
                             } else {
-                                // Use the original method for backward compatibility
                                 customizations = await serverConfigurationProvider.listAvailableCustomizations()
                             }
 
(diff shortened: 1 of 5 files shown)


server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -24,13 +24,43 @@ const Q_DEVELOPER_PROFILES = 'developerProfiles'
...
+    profileArn?: string
+}
+
+interface QConfigurationSections {
+    customizations: CustomizationWithMetadata[]
+    developerProfiles?: AmazonQDeveloperProfile[]
+}
+
+// Feature flag interface for client capabilities
+interface QClientCapabilities {
```
viktorsaws, line 38: nit: could be moved to shared interface location

justinmk3:
Other than adding new fields and requiring an opt-in flag, are any other client-side changes needed? The shape of the result did not change (except new fields)?

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -171,6 +238,73 @@ export class ServerConfigurationProvider {
         }
     }
 
+    async listAvailableCustomizationsForProfileAndRegion(profileArn: string, region: string): Promise<Customizations> {
+        try {
+            // Create a new service for the specific region
+            const service = this.serviceManager.getServiceFactory()(region, AWS_Q_ENDPOINTS.get(region) || '')
+            service.profileArn = profileArn
+
+            const customizations = (await service.listAvailableCustomizations({ maxResults: 100 })).customizations
```
awschristou, line 247: One of the requirements driving this change is that the clients can tell when a profile doesn't have a customization. I'd like Flare to model this, which gives the clients standardized handling.

Can we do the following:
- add a field `isDefault` to `CustomizationWithMetadata` (or Customization?)
- add a customization to _every_ profile, which marks isDefault to true
  - this customization would have a null arn (this is equivalent to clients sending "set null" meaning "use the default").
  - the name would be whatever VS Code and JetBrains currently set for their placeholder that represents "default customization". This could change in the future
  - Description is probably blank

This would be future compatible. Lets say the service serves up default markers in the future, then we would only do this for profiles that have no default customization.

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -24,13 +24,43 @@ const Q_DEVELOPER_PROFILES = 'developerProfiles'
 export const Q_CUSTOMIZATIONS_CONFIGURATION_SECTION = `${Q_CONFIGURATION_SECTION}.${Q_CUSTOMIZATIONS}`
 export const Q_DEVELOPER_PROFILES_CONFIGURATION_SECTION = `${Q_CONFIGURATION_SECTION}.${Q_DEVELOPER_PROFILES}`
 
+interface CustomizationWithMetadata extends Customization {
+    region?: string
+    profileArn?: string
```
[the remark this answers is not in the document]
pras0131: In that case, I can just pass the full profile object as I process through it rather than taking out fields like name, region etc. Something like CustomizationWithMetadata to have a field named profile.

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
pras0131: arn instead would be an empty string instead of null as it's non-null as defined right now.

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -276,10 +276,22 @@ export class ServerConfigurationProvider {
...
+                                    arn: '',
+                                    name: 'Amazon Q foundation (Default)',
+                                    description: '',
+                                    isDefault: true,
+                                    profile: profile,
+                                },
+                            ]
+                        }
+
                         return customizations.map(customization => ({
```
awschristou, line 291: Should a "Amazon Q foundation" default customization be unconditionally added to each profile? Even if the backend returns two customizations "Foo" and "Bar", users should be able to select from a list on the UI: "Amazon Q foundation", "Foo", "Bar", and in this way users are able to select the foundation customization (eg: not select one of the ones specifically in their profile).

pras0131: Wasn't it handled by the extension side even before the concept of developer profiles in place? listAvailableCustomizations returns the other customizations, not the foundational one.

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -171,6 +238,93 @@ export class ServerConfigurationProvider {
...
+                            },
+                        ] as CustomizationWithMetadata[]
+                    })
+            })
+
+            const results = await Promise.all(customizationPromises)
+
+            return results.flat()
+        } catch (error) {
+            if (error instanceof ResponseError) {
```
volodkevych, line 321: Why do we need this `if`?

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
```diff
@@ -56,27 +86,64 @@ export const QConfigurationServerToken =
...
                             return amazonQServiceManager.getEnableDeveloperProfileSupport()
                                 ? { customizations, developerProfiles }
                                 : { customizations }
                         case Q_CUSTOMIZATIONS_CONFIGURATION_SECTION:
-                            customizations = await serverConfigurationProvider.listAvailableCustomizations()
+                            if (
+                                enableCustomizationsWithMetadata &&
+                                amazonQServiceManager.getEnableDeveloperProfileSupport()
+                            ) {
+                                logging.debug('Using enhanced customizations with metadata')
```
volodkevych, line 133: nit: some copy/paste here could we refactored

volodkevych (approved):

server/aws-lsp-codewhisperer/src/language-server/configuration/qConfigurationServer.ts
pras0131: yes agreed, I'll clean this up a bit in a follow up PR.

viktorsaws:
> Other than adding new fields and requiring an opt-in flag, are any other client-side changes needed? The shape of the result did not change (except new fields)?

@justinmk3 no other client-side changes are needed, except of new desired UX to surface customizations.

PR #1252, opened 2025-05-02
fix: emit error metric without needing status code

chungjac:
## Problem
- Previously, error metrics were only emitted when the backend returned a status code. However, mid-stream exceptions, which we now log the request IDs, lack status codes, causing these error metrics to not be emitted

## Solution
- Extended error metric emission to include cases without status codes, while filtering out unactionable errors (such as user-dependent errors like PromptCharacterLimit). This ensures we capture all relevant error states for monitoring and debugging.

    

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1534,19 +1534,11 @@ export class AgenticChatController implements ChatHandlers {
...
         metric: Metric<CombinedConversationEvent>
     ): ChatResult | ResponseError<ChatResult> {
-        let errorMessage: string | undefined
-        let requestID: string | undefined
-        if (isAwsError(err) || (isObject(err) && typeof getHttpStatusCode(err) === 'number')) {
-            if (err instanceof CodeWhispererStreamingServiceException) {
-                errorMessage = err.message
-                requestID = err.$metadata.requestId
-            } else if (err?.cause?.message) {
-                errorMessage = err?.cause?.message
```
tsmithsz, line 1544: Should we add this case to the `getErrorMessage()` function?

chungjac: After testing, I actually don't think that `cause` will ever have a `message` field. Will add it to the getErrorMessage() function just in case

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1534,19 +1534,11 @@ export class AgenticChatController implements ChatHandlers {
...
-                requestID = err.cause?.$metadata.requestId
-            } else if (err instanceof Error || err?.message) {
-                errorMessage = err.message
-            }
+        const errorMessage = getErrorMessage(err)
+        const requestID = getRequestID(err)
 
+        // only emit metrics for actionable errors (skip user-dependent errors like PromptCharacterLimit)
+        if (!unactionableErrorCodes.includes(err.code)) {
```
ashishrp-aws, line 1541: Instead of not emitting can we just add another object to metric to differentiate. Let's say if there are large number of unactionable errors are occurring, we can't the ignore the pattern and solve it from our end.

ashishrp-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1548,22 +1548,22 @@ export class AgenticChatController implements ChatHandlers {
...
+        const requestID = getRequestID(err)
+        metric.setDimension('cwsprChatResponseCode', getHttpStatusCode(err) ?? 0)
+        metric.setDimension('languageServerVersion', this.#features.runtime.serverInfo.version)
 
-            metric.setDimension('cwsprChatResponseCode', getHttpStatusCode(err) ?? 0)
-            metric.setDimension('languageServerVersion', this.#features.runtime.serverInfo.version)
-            this.#telemetryController.emitMessageResponseError(tabId, metric.metric, requestID, errorMessage)
+        // use custom error message for unactionable errors (user-dependent errors like PromptCharacterLimit)
+        if (err.code && err.code in unactionableErrorCodes) {
+            const customErrMessage = unactionableErrorCodes[err.code as keyof typeof unactionableErrorCodes]
```
laileni-aws, line 1558: Can you add log message for emitting error message?

laileni-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
chungjac: Already done in these places:
https://github.com/aws/language-servers/blob/def522daee62ea37556fefe12352ef28f38523d1/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts#L1587

https://github.com/aws/language-servers/blob/def522daee62ea37556fefe12352ef28f38523d1/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts#L1595

laileni-aws (approved):

PR #1258, opened 2025-05-05
fix: use proper condition for trigger index enablement response

breedloj:
## Problem
We are conditionally reacting off `isEnabled` in the context controller for whether to send a chat response indicating the user should enable workspace indexing to complete their request. This is not working as expected since the context controller may be enabled while indexing is not.

## Solution
Moving the check to `isIndexingEnabled` which explicitly indicates that workspace indexing is enabled within the controller.

## License


jpinkney-aws (approved):

jpinkney-aws:
Can we hold off on merging this until after the release tomorrow?

PR #1260, opened 2025-05-05
fix: update listDirectory tool to output in tree-like format to reduce size

jguoamz:
## Problem
- Current listDirectory result shows all files/directories with full path which has a lot of duplicates

## Solution
- Make it return tree-like result which saves a lot of characters
  - On the same repo, output character count before the change: 18632, output character count after the change: 7348

Screenshot 2025-05-05 at 3 55 30 PM

## License

### DIFF
--- a/core/aws-lsp-core/src/util/workspaceUtils.test.ts
+++ b/core/aws-lsp-core/src/util/workspaceUtils.test.ts
@@ -3,7 +3,13 @@ import * as fs from 'fs/promises'
 import * as assert from 'assert'
 import * as path from 'path'
 import { TestFolder } from '../test/testFolder'
-import { readDirectoryRecursively, getEntryPath, isParentFolder, isInWorkspace } from './workspaceUtils'
+import {
+    readDirectoryRecursively,
+    readDirectoryWithTreeOutput,
+    getEntryPath,
+    isParentFolder,
+    isInWorkspace,
+} from './workspaceUtils'
 import { TestFeatures } from '@aws/language-server-runtimes/testing'
 
 describe('workspaceUtils', function () {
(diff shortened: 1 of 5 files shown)


core/aws-lsp-core/src/util/workspaceUtils.ts
```diff
@@ -109,3 +109,124 @@ export function isParentFolder(parentPath: string, childPath: string): boolean {
 export function isInWorkspace(workspaceFolderPaths: string[], filepath: string) {
     return workspaceFolderPaths.some(wsFolder => isParentFolder(wsFolder, filepath) || wsFolder === filepath)
 }
+
+/**
+ * Output the result in tree-like format, this will save tokens as the output contains much less characters
+ * project-root/
```
Hweinstock, line 115: nice example, that will be really helpful.

server/aws-lsp-codewhisperer/src/language-server/chat/constants.ts
```diff
@@ -31,6 +31,7 @@ export const HELP_MESSAGE = `I'm Amazon Q, a generative AI assistant. Learn more
 
 export const DEFAULT_HELP_FOLLOW_UP_PROMPT = 'How can Amazon Q help me?'
 
+// TO BE DEPRECATED, use DEFAULT_EXCLUDE_DIRS and DEFAULT_EXCLUDE_FILES instead
```
Hweinstock, line 34: AFAIK this is only used by fileSearch. Is there any reason file search can't use the tree output utility as well?

If so, we could remove the `readDirectoryRecursively` completely in favor of the tree format.

Hweinstock (commented):
This new utility and `readDirectoryRecursively` do basically the same thing with different output formats. Can we consolidate them into a single implementation? Or simply use the new utility always?

core/aws-lsp-core/src/util/workspaceUtils.test.ts
```diff
@@ -209,4 +215,160 @@ describe('workspaceUtils', function () {
...
+            await subdir2.write('file2', 'this is other content')
+
+            const subdir11 = await subdir1.nest('subdir11')
+            await subdir11.write('file3', 'this is also content')
+            const subdir12 = await subdir1.nest('subdir12')
+            await subdir12.write('file4', 'this is even more content')
+            await subdir12.write('file5', 'and this is it')
+
+            const expected =
+                '|-- subdir1/\n' +
```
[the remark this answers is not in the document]
Hweinstock: I believe the full path to the workspace root is part of the tool [parameters](https://github.com/aws/language-servers/blob/394db61133e09cfaeff2f7510ab60c571c562130/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/listDirectory.ts#L9) (given by the model), so its already aware of the full path.

Hweinstock:
> On the same repo, output character count before the change: 18632, output character count after the change: 7348

Wow, that's awesome. Do we have any telemetry (client or server side) that allows us to track this kind of stuff? Curious how we can measure the impact of this change at scale.

aditya169 (commented):
Please test on Linux and Windows as well...

core/aws-lsp-core/src/util/workspaceUtils.test.ts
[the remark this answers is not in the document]
jguoamz: the full path contains the root path, but in the test, it's a random path, so exlcuded

server/aws-lsp-codewhisperer/src/language-server/chat/constants.ts
jguoamz: fileSearch is pending major changes, so leave it as it is for now

jguoamz:
> This new utility and `readDirectoryRecursively` do basically the same thing with different output formats. Can we consolidate them into a single implementation? Or simply use the new utility always?

we should deprecate `readDirectoryRecursively` once fileSearch is updated

Hweinstock (approved):

PR #1268, opened 2025-05-06
fix: clear history for `inputTooLong` errors

tsmithsz:
## Problem
We currently try to estimate and trim the number of tokens in each request before sending to the server. However, this is an estimation and is not always accurate. As a result, we see a number of `Input too long` errors in our metrics.

We also force users to open a new conversation tab to continue.

## Solution
- Clear current tab history for `inputTooLong` errors to allow users to continue in current chat tab
- Rephrase customer-facing message

## Testing

Skip through the video if needed:

https://github.com/user-attachments/assets/7ac68c98-1169-4635-b6b1-a66da2549686

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -1081,7 +1081,7 @@ describe('AgenticChatController', () => {
             const typedChatResult = chatResult as ResponseError<ChatResult>
             assert.strictEqual(
                 typedChatResult.data?.body,
-                'Too much context loaded. Please start a new conversation or ask about specific files.'
+                'Too much context loaded. I have cleared the conversation history. Please retry your request with smaller input.'
             )
         })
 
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
```diff
@@ -1081,7 +1081,7 @@ describe('AgenticChatController', () => {
             const typedChatResult = chatResult as ResponseError<ChatResult>
             assert.strictEqual(
                 typedChatResult.data?.body,
-                'Too much context loaded. Please start a new conversation or ask about specific files.'
+                'Too much context loaded. I have cleared the conversation history. Please retry your request.'
```
jguoamz, line 1084: should we mention something along the lines of "with smaller input"?

tsmithsz: Added. We may need to modify this message entirely after Ankit's input

jguoamz (approved):

PR #1272, opened 2025-05-06
fix: regex should match workspace text in bold style and startLine can be 0

andrewyuq:
## Problem
1. @ workspace is in bold for which current regex can not capture and remove it
2. startLine || -1 will be -1 if startLine is 0, but we want 0 so use startLine ?? -1
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
@@ -119,7 +119,7 @@ export class AgenticChatTriggerContext {
         }
 
         if (hasWorkspace) {
-            promptContent = promptContent?.replace(/^@workspace\/?/, '')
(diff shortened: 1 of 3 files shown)


jpinkney-aws (commented):
Are either of these things we could catch with tests in the future?

andrewyuq:
> Are either of these things we could catch with tests in the future?

invididual unit tests could catch these. Running context E2E could also easily catch I believe.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
```diff
@@ -119,7 +119,7 @@ export class AgenticChatTriggerContext {
         }
 
         if (hasWorkspace) {
-            promptContent = promptContent?.replace(/^@workspace\/?/, '')
+            promptContent = promptContent?.replace(/\*\*@workspace\*\*/, '')
```
volodkevych, line 122: nit: it's good to replace both options - bold and not bold, otherwise we depend on a very specific UI here

volodkevych (approved):

PR #1273, opened 2025-05-07
fix: errors/cancellation not resetting undoAll state

ctlai95:
## Problem

When errors and user cancellations happen, the undo all state is not reset. This can lead to the "Undo all changes" button appearing unexpectedly.

## Solution

Call `#showUndoAllIfRequired` method in the error and cancellation handlers. Also reset the state even if there is only 1 write operation in the set.

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27107,7 +27107,7 @@
                 "@aws-sdk/client-sso-oidc": "^3.616.0",
                 "@aws-sdk/token-providers": "^3.744.0",
                 "@aws/language-server-runtimes": "^0.2.78",
-                "@aws/lsp-core": "^0.0.4",
+                "@aws/lsp-core": "^0.0.5",
                 "@smithy/node-http-handler": "^3.2.5",
                 "@smithy/shared-ini-file-loader": "^4.0.1",
                 "https-proxy-agent": "^7.0.5",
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
```diff
@@ -67,6 +67,10 @@ export class ChatSessionService {
         return this.#toolUseLookup
     }
 
+    public set toolUseLookup(toolUseLookup) {
+        this.#toolUseLookup = toolUseLookup
```
jguoamz, line 71: is this being used?

package-lock.json
```diff
@@ -27107,7 +27107,7 @@
                 "@aws-sdk/client-sso-oidc": "^3.616.0",
                 "@aws-sdk/token-providers": "^3.744.0",
                 "@aws/language-server-runtimes": "^0.2.78",
-                "@aws/lsp-core": "^0.0.4",
+                "@aws/lsp-core": "^0.0.5",
```
jpinkney-aws, line 27110: hmm, shouldn't this have been done by the release/merge to main earlier 🤔

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
jpinkney-aws: it looks like its just used for the tests, I think it should also have a type here though

ctlai95: Yeah it was for the tests, and it's actually able to infer the correct type I think because of the setter pattern

package-lock.json
ctlai95: Hmm not sure, but I see in main it's still 0.0.4

jguoamz (approved):

PR #1274, opened 2025-05-07
fix: abandon requests with invalid toolResults

tsmithsz:
## Problem

We occasionally have a race condition caused by the stop button, where toolResults for cancelled tool executions get processed after the conversation has continued, resulting in erros: 

`Got ValidationError: ToolResult with ID tooluse_ has no corresponding ToolUse`

## Solution
- Added validation to check when a message has tool results but the last message has no tool uses
- Added logic to filter valid and invalid tool results, keeping valid ones and marking invalid ones as cancelled.
- Added logic to terminate agent loop when history is invalid

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
@@ -368,21 +368,21 @@ export class ChatDatabase {
      * 5. If the last message is from the assistant and it contains tool uses, and a next user
      *    message is set without tool results, then the user message will have cancelled tool results.
      */
-    fixHistory(tabId: string, newUserMessage: ChatMessage, conversationId: string): void {
+    fixHistory(tabId: string, newUserMessage: ChatMessage, conversationId: string): boolean {
         if (!this.#initialized) {
-            return
+            return true
         }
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -611,20 +612,57 @@ export class ChatDatabase {
...
+            return true
         }
 
         const lastMsg = messages[messages.length - 1]
+        const toolResults = newUserMessage.userInputMessage?.userInputMessageContext?.toolResults
+
+        if (toolResults && toolResults.length > 0) {
+            // If last message has no tool uses but new message has tool results, this is invalid
+            if (!lastMsg.toolUses || lastMsg.toolUses.length === 0) {
+                this.#features.logging.warn('New message has tool results but last message has no tool uses')
```
jguoamz, line 632: will warn show up?

tsmithsz: From my testing it shows up: `[Warn  - 7:14:04 PM] [2025-05-07T02:14:04.645Z] testing log level`.

Since we show info and logging levels are hierarchical I think we should see them.

Issue #801, opened 2025-02-25
SSO token refresh does not retry on network errors, which can result in spurious reports of "re-login required"


justinmk3:
# Problem

Users are experiencing frequent re-login requirement. 

This may be due to transient network issues: [sso token refresh](https://github.com/aws/language-servers/blob/47e94bd09036eaee489d0774e31ef4f8654fefbb/server/aws-lsp-identity/src/sso/cache/refreshingSsoCache.ts#L163-L170) can fail because of transient network issues?

If network issues are not handled by retrying (with backoff), then temporary network issues can randomly surface to users as "you need to login again".

Examples in AWS Toolkit:

- https://github.com/aws/aws-toolkit-vscode/pull/3501
- https://github.com/aws/aws-toolkit-vscode/pull/3498
- and more (there were several followups, because the issue is tricky)

# Expected behavior

The sso lsp server should retry and be robust/adaptive under transient network issues.

To see an example of how AWS Toolkit currently handles this, look for references to `isNetworkError` in the codebase:  https://github.com/aws/aws-toolkit-vscode/blob/cc0455f112c3be8db3232acfb777341f3225c557/packages/core/src/shared/errors.ts#L829

```
packages/core/src/auth/auth.ts|16 col 33-47| import { errorCode, isAwsError, isNetworkError, ToolkitError, UnknownError } from '../shared/errors'
packages/core/src/auth/auth.ts|865 col 13-27| if (isNetworkError(e)) {
packages/core/src/auth/secondaryAuth.ts|18 col 10-24| import { isNetworkError } from '../shared/errors'
packages/core/src/auth/secondaryAuth.ts|369 col 21-35| if (isNetworkError(err) && connectionState === 'valid') {
packages/core/src/auth/sso/ssoAccessTokenProvider.ts|22 col 5-19| isNetworkError,
packages/core/src/auth/sso/ssoAccessTokenProvider.ts|201 col 25-39| } else if (!isNetworkError(err)) {
packages/core/src/codewhisperer/util/authUtil.ts|9 col 24-38| import { ToolkitError, isNetworkError, tryRun } from '../../shared/errors'
packages/core/src/codewhisperer/util/authUtil.ts|428 col 27-41| (err) => !isNetworkError(err),
packages/core/src/shared/errors.ts|829 col 17-31| export function isNetworkError(err?: unknown): err is Error & { code: string } {
packages/core/src/test/shared/errors.test.ts|15 col 5-19| isNetworkError,
[...]

PR #1283, opened 2025-05-07
fix: set streamingClient timeout config

jguoamz:
## Problem
- We are seeing errors like "socket hang up", "read ECONNRESET", "aborted", "read ETIMEDOUT" for long-lasting requests, this is caused by streaming client having a shorter timeout than the service timeout

## Solution
- set streamingClient timeout config to match service limit

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/streamingClientService.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/streamingClientService.ts
@@ -14,6 +14,7 @@ import { CredentialsProvider, SDKInitializator, Logging } from '@aws/language-se
 import { getBearerTokenFromProvider } from './utils'
 import { ConfiguredRetryStrategy } from '@aws-sdk/util-retry'
 import { CredentialProviderChain, Credentials } from 'aws-sdk'
+import { clientTimeoutMs } from '../language-server/agenticChat/constants'
 
 export type SendMessageCommandInput =
     | SendMessageCommandInputCodeWhispererStreaming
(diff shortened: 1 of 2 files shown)


ctlai95 (approved):

tsmithsz (approved):

laileni-aws (approved):

server/aws-lsp-codewhisperer/src/shared/streamingClientService.ts
```diff
@@ -74,6 +75,11 @@ export class StreamingClientServiceToken extends StreamingClientServiceBase {
             endpoint,
             token: tokenProvider,
             retryStrategy: new ConfiguredRetryStrategy(0, (attempt: number) => 500 + attempt ** 10),
+            requestHandler: {
+                keepAlive: true,
+                requestTimeout: clientTimeoutMs,
+                connectionTimeout: clientTimeoutMs,
```
justinmk3, line 81: is this increasing or decreasing the timeout? that's the key part of the "Solution" description that's missing. "set streamingClient timeout config" is not a useful Solution description.

if you are increasing the timeout to ... 3 minutes, the question is why? will that prevent the client from retrying for up to 3 minutes, even when there is no activity on the connection?

how many retries are configured currently?

justinmk3:
is this related to https://github.com/aws/language-servers/issues/801 ?

jguoamz:
> is this related to #801 ?

no, this is addressing issues where we see timeout errors for long requests

server/aws-lsp-codewhisperer/src/shared/streamingClientService.ts
jguoamz: removed `connectionTimeout` setting as it is not needed and can cause regression

justinmk3 (approved):
still missing a clear (but concise) description of the problem being solved, and how the changes solve it

PR #1293, opened 2025-05-08
Fix bugs impacting indexing disabled functionality

nborges-aws:
## Problem
Disabling local indexing in IDE settings was not reflected in the chat results. The message prompting users to "enable workspace indexing" was never emitted, and query index was still being hit with setting disabled. Vector library instance in controller is always initialized, so checking for its existence is not sufficient for disabling indexing.
## Solution
Call correct method in trigger context conditional to check for indexing enabled and notify extension if disabled. Additionally, add checks in the local context controller to ensure setting is correctly considered during querying, and cache is cleared upon disable.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
@@ -262,7 +262,7 @@ export class AgenticChatTriggerContext {
         chatResultStream?: AgenticChatResultStream
     ): Promise<RelevantTextDocumentAddition[]> {
         const localProjectContextController = await LocalProjectContextController.getInstance()
-        if (!localProjectContextController.isIndexingEnabled && chatResultStream) {
+        if (!localProjectContextController.isIndexingEnabled() && chatResultStream) {
             await chatResultStream.writeResultBlock({
                 body: `To add your workspace as context, enable local indexing in your IDE settings. After enabling, add @workspace to your question, and I'll generate a response using your workspace as context.`,
                 buttons: [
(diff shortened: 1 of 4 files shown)


breedloj (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
```diff
@@ -262,7 +262,7 @@ export class AgenticChatTriggerContext {
         chatResultStream?: AgenticChatResultStream
     ): Promise<RelevantTextDocumentAddition[]> {
         const localProjectContextController = await LocalProjectContextController.getInstance()
-        if (!localProjectContextController.isIndexingEnabled && chatResultStream) {
+        if (!localProjectContextController.isIndexingEnabled() && chatResultStream) {
```
justinmk3, line 265: was this a syntax accident? do we need a lint warning to guard against this?

nborges-aws: Yup, accidental. IIRC the variable was at one point named isIndexingEnabled, and refactors ended up changing that to the name of the method, and variable to _isIndexingEnabled. This was most likely overlooked during refactor

avi-alpert:
Please fix unit test failures before merging

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -184,7 +187,7 @@ export class LocalProjectContextController {
     }
 
     public async updateIndex(filePaths: string[], operation: UpdateMode): Promise<void> {
-        if (!this._vecLib) {
+        if (!this._vecLib || !this.isIndexingEnabled()) {
```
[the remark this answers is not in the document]
nborges-aws: good call. updated

justinmk3 (approved):

PR #1299, opened 2025-05-09
feat: merge updates for inline completions

zixlin7:
## Problem
merge in changes from feature branch:  https://github.com/aws/language-servers/tree/feature/qflare 
This includes:

- pagination
- language support
- imports

all of these are backward compatible

## Solution

Most changes have been previously reviewed

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.test.ts
@@ -21,17 +21,22 @@ import {
 import { CodeWhispererSession, SessionData, SessionManager } from './session/sessionManager'
 import {
     EMPTY_RESULT,
+    EXPECTED_NEXT_TOKEN,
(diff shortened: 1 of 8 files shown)


jpinkney-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.test.ts
```diff
@@ -73,7 +78,7 @@ const startServer = async (features: TestFeatures, server: Server): Promise<Test
     return features
 }
 
-describe('CodeWhisperer Server', () => {
+describe.only('CodeWhisperer Server', () => {
```
jpinkney-aws, line 81: wait why do we need `describe.only` here?

zixlin7: ooops I shouldn't commit that, will revert, thanks for catching!

jpinkney-aws (approved):

PR #1301, opened 2025-05-09
chore: script to create local build

pras0131:
## Problem
It contains the changes to have a packaged build ready for agent-standalone as well as build for chat-client, both with the help of single script. This can help to ease the integration with the clients.

## License


pras0131:
> Would it make sense to add some README about how to use this script directly into the repo? Also, worth integrating with package.json scripts?

Yeah, right adding this to the package.json would be more explicit. Let me add the related change and README as well.

PR #1314, opened 2025-05-12
feat(amazonq): telemetry for chat history and export

avi-alpert:
## Problem
There is no telemetry to measure metrics for chat history and export features:

- Number of tabs automatically restored from history
- Number of users exporting chat transcripts, with which format (markdown vs html)
- Number of users who are opening chat history list
- Number of users who are searching, exporting, deleting, and opening conversations from history
- How long it takes to complete history searches
- History file sizes, time to load history files

## Solution
Emit the following metrics:
- amazonq_exportTab - when a user exports current tab
- amazonq_loadHistory - when Q automatically loads history file for workspace
- amazonq_performChatHistoryAction - when user opens, searches, deletes, or exports conversation from history
- ui-click - when user clicks history tab bar button

Corresponding metrics PR: https://github.com/aws/aws-toolkit-common/pull/1023

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -115,7 +129,15 @@ export class ChatDatabase {
...
-    async databaseInitialize() {
+    /**
+     * Gets the current size of the database file in bytes.
+     * @returns Promise that resolves to the file size in bytes, or undefined if the file doesn't exist
+     */
+    getDatabaseFileSize(): number | undefined {
+        return this.#dbFileSize
+    }
+
+    async databaseInitialize(startTime: number) {
```
jngyuamz, line 140: Should we re-calculate the dbFileSize in this method? So that it automatically maintains the `#dbSize` when doing autoload

jngyuamz (approved):
> Should we re-calculate the dbFileSize in this method? So that it automatically maintains the #dbSize when doing autoload

Not a blocker since this PR is for telemetry.

tsmithsz (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
avi-alpert: I think it would make more sense to recalculate in addMessage. im not sure autoload happens after the initial loading anyway

PR #1319, opened 2025-05-13
fix: update ignore pattern of glob for sspc

wweitao:
## Problem
Source code which is created under /test/ or /tests/, they are ignored falsely, same as directory which might have hidden repository before the code package.

## Solution

Remove them from ignore patterns of glob

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
@@ -33,12 +33,8 @@ const IGNORE_PATTERNS = [
     '**/dist/**',
     '**/build/**',
     '**/out/**',
-    // Test directories
-    '**/test/**',
-    '**/tests/**',
     '**/coverage/**',
-    // Hidden directories and files
-    '**/.*/**',
+    // Hidden files
     '**/.*',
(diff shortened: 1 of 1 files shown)


ege0zcan (approved):

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
```diff
@@ -33,12 +33,8 @@ const IGNORE_PATTERNS = [
     '**/dist/**',
     '**/build/**',
     '**/out/**',
-    // Test directories
-    '**/test/**',
-    '**/tests/**',
     '**/coverage/**',
-    // Hidden directories and files
-    '**/.*/**',
```
ege0zcan, line 41: why do we want hidden directories?

wweitao: I'm thinking if customer set any thing like this in their path.

ege0zcan: Could you give an example

ege0zcan (approved):

PR #1320, opened 2025-05-13
fix: update fileSearch toolSpec and implementation

jguoamz:
## Problem
- FileSearch is returning directories in the result
- FileSearch toolSpec is outdated

## Solution
- Update fileSearch toolSpec and implementation
- This PR is not enabling the tool, we will test offline first

## License

### DIFF
--- a/core/aws-lsp-core/src/util/workspaceUtils.ts
+++ b/core/aws-lsp-core/src/util/workspaceUtils.ts
@@ -12,14 +12,16 @@ export async function readDirectoryRecursively(
     folderPath: string,
     options?: {
         maxDepth?: number
-        excludeEntries?: string[]
+        excludeDirs?: string[]
+        excludeFiles?: string[]
         customFormatCallback?: (entry: Dirent) => string
         failOnError?: boolean
     },
     token?: CancellationToken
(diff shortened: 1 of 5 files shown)


core/aws-lsp-core/src/util/workspaceUtils.ts
```diff
@@ -12,14 +12,16 @@ export async function readDirectoryRecursively(
...
+        excludeFiles?: string[]
         customFormatCallback?: (entry: Dirent) => string
         failOnError?: boolean
     },
     token?: CancellationToken
 ): Promise<string[]> {
     const dirExists = await features.workspace.fs.exists(folderPath)
-    const excludeEntries = options?.excludeEntries ?? []
+    const excludeFiles = options?.excludeFiles ?? []
+    const excludeDirs = options?.excludeDirs ?? []
```
grl-wang, line 24: is the listDirectory use the same workspaceUtils.ts, do we need to update the listDirectory?

jguoamz: it does not use the same method. It uses a new method that is already updated

grl-wang (approved):

PR #1322, opened 2025-05-13
fix: open initial tab using mynahUI defaults instead of waiting for ChatOptions

viktorsaws:
## Problem
The logic to open empty tab in Chat Client recently changed in https://github.com/aws/language-servers/pull/988/files. Before, initial tab was sent with default MynahUI options to alway open one at creating MynahUI instance. This was removed and opening new tab is delayed until `ChatOptions` inbound message is sent from Extension to Chat Client.

This introduces regression for old VS extensions, which didn't implement `ChatOptions` event. Since event is never sent into Chat client, it is not fully initialized and empty initial tab is not opened.

## Solution
Introduce old logic for opening initial tab using MynahUI defaults.

## License


volodkevych (approved):

chat-client/src/client/mynahUi.ts
```diff
@@ -483,7 +484,15 @@ export const createMynahUi = (
     }
 
     const mynahUiProps: MynahUIProps = {
-        tabs: {},
+        tabs: {
+            [tabFactory.initialTabId]: {
+                isSelected: true,
+                store: {
+                    ...tabFactory.createTab(true),
```
justinmk3, line 491: @viktorsaws should this pass `disclaimerAcknowledged` instead of `true` ?

viktorsaws: Yes, that was a miss, seems to be fixed in https://github.com/aws/language-servers/pull/1332/files

PR #1324, opened 2025-05-13
feat: expose MCP server init status

bywang56:
## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
```diff
@@ -220,6 +220,7 @@ describe('AgenticChatController', () => {
                 }))
             ),
             addTool: sinon.stub().resolves(),
+            removeTool: sinon.stub().returns(void 0),
```
ashishrp-aws, line 223: why is it not return undefined?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpTypes.ts
```diff
@@ -3,6 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+export type McpServerStatus = 'INITIALIZING' | 'READY' | 'FAILED' | 'DISABLED'
```
ashishrp-aws, line 6: nit: intsead of `READY`, can we use something else like `ENABLED`

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
bywang56: this is the same as undefined, but I can change it to align

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpTypes.ts
bywang56: make sense, so it match DISABLED

PR #1329, opened 2025-05-13
refactor(amazonq): optimization: remove validation call for selected profile

Will-ShaoHua:
## Problem
TLDR: reduce # of `listAvailableProfiles` API call in Flare due to strict service tps restriction (2 per sec).

Below Client already validate selected profile on plugin/ide startup, so the validation should be safe to remove from Flare ATM
- vscode
- jetbrains
- eclipse

Note: VS (Visual studio doesn't validate profile within plugin codepath, it's relying on Flare to perform the validation, we will follow up on this as VSCode users tend to open many windows and the validation cause service availability to drop and our priority is to recover backend service availability)

Currently Flare doesn't have global state management implemented thus we can't really cache data and share across different Flare instances. However the server side has a strict tps 2 for `listAvailableProfile` call therefore we have to cut down the # of call to mitigate the impact of getting throttled when users have multiple IDE instances opened (i.e. multiple Flare instances)

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
@@ -802,30 +802,18 @@ describe('AmazonQTokenServiceManager', () => {
                 assert.throws(() => amazonQTokenServiceManager.getCodewhispererService())
             })
 
-            it.skip('cancels on-going profile update when credentials are deleted', async () => {
+            it('cancels on-going profile update when credentials are deleted', async () => {
                 await setupServiceManagerWithProfile()
 
-                const pendingUpdate = amazonQTokenServiceManager.handleOnUpdateConfiguration(
-                    {
-                        section: 'aws.q',
-                        settings: {
-                            profileArn: 'arn:aws:testprofilearn:eu-central-1:11111111111111:profile/QQQQQQQQQQQQ',
-                        },
-                    },
-                    {} as CancellationToken
-                )
-
+                amazonQTokenServiceManager.setState('PENDING_Q_PROFILE_UPDATE')
                 assert.strictEqual(amazonQTokenServiceManager.getState(), 'PENDING_Q_PROFILE_UPDATE')
 
                 amazonQTokenServiceManager.handleOnCredentialsDeleted('bearer')
 
                 assert.strictEqual(amazonQTokenServiceManager.getState(), 'PENDING_CONNECTION')
 
-                await assert.rejects(() => pendingUpdate)
-
                 assert.strictEqual(amazonQTokenServiceManager.getState(), 'PENDING_CONNECTION')
                 assert.strictEqual(amazonQTokenServiceManager.getActiveProfileArn(), undefined)
-                sinon.assert.calledOnce(codewhispererServiceStub.abortInflightRequests)
                 assert.throws(() => amazonQTokenServiceManager.getCodewhispererService())
             })
 
--- a/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
@@ -420,7 +420,6 @@ describe('AmazonQTokenServiceManager', () => {
                 const service = amazonQTokenServiceManager.getCodewhispererService()
                 assert.strictEqual(amazonQTokenServiceManager.getState(), 'INITIALIZED')
                 assert.strictEqual(amazonQTokenServiceManager.getConnectionType(), 'identityCenter')
-                assert(codewhispererStubFactory.calledOnceWithExactly('eu-central-1', TEST_ENDPOINT_EU_CENTRAL_1))
 
                 assert.strictEqual(results[0].status, 'rejected')
                 assert(results[0].reason.message, 'Requested profile update got cancelled')
--- a/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
@@ -26,7 +26,6 @@ import * as qDeveloperProfilesFetcherModule from './qDeveloperProfiles'
 import { setCredentialsForAmazonQTokenServiceManagerFactory } from '../testUtils'
 import { StreamingClientServiceToken } from '../streamingClientService'
 import { generateSingletonInitializationTests } from './testUtils'
-import { SinonFakeTimers } from 'sinon'
 
 export const mockedProfiles: qDeveloperProfilesFetcherModule.AmazonQDeveloperProfile[] = [
     {
@@ -63,11 +62,8 @@ describe('AmazonQTokenServiceManager', () => {
 
     let amazonQTokenServiceManager: AmazonQTokenServiceManager
     let features: TestFeatures
-    let clock: SinonFakeTimers
 
     beforeEach(() => {
-        clock = sinon.useFakeTimers()
-
         // Override endpoints for testing
         AWS_Q_ENDPOINTS.set('us-east-1', TEST_ENDPOINT_US_EAST_1)
         AWS_Q_ENDPOINTS.set('eu-central-1', TEST_ENDPOINT_EU_CENTRAL_1)
@@ -107,7 +103,6 @@ describe('AmazonQTokenServiceManager', () => {
         AmazonQTokenServiceManager.resetInstance()
         features.dispose()
         sinon.restore()
-        clock.restore()
     })
 
     const setupServiceManager = (enableProfiles = false) => {
--- a/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
@@ -26,6 +26,7 @@ import * as qDeveloperProfilesFetcherModule from './qDeveloperProfiles'
 import { setCredentialsForAmazonQTokenServiceManagerFactory } from '../testUtils'
 import { StreamingClientServiceToken } from '../streamingClientService'
 import { generateSingletonInitializationTests } from './testUtils'
+import { SinonFakeTimers } from 'sinon'
 
 export const mockedProfiles: qDeveloperProfilesFetcherModule.AmazonQDeveloperProfile[] = [
     {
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/qDeveloperProfiles.ts
```diff
@@ -86,7 +86,7 @@ export const getListAllAvailableProfilesHandler =
         return allProfiles
     }
 
-async function fetchProfilesFromRegion(
+export async function fetchProfilesFromRegion(
```
Will-ShaoHua, line 89: could revert

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -318,21 +315,20 @@ export class AmazonQTokenServiceManager extends BaseAmazonQServiceManager<
...
-            endpoints: new Map([[parsedArn.region, endpoint]]),
-        })
-
-        this.handleTokenCancellationRequest(token)
-
-        const newProfile = profiles.find(el => el.arn === newProfileArn)
+        // Hack to inject a dummy profile name as it's not used by client IDE for now, if client IDE starts consuming name field then we should also pass both profile name and arn from the IDE
+        const newProfile: AmazonQDeveloperProfile = {
+            arn: newProfileArn,
+            name: 'Not being used',
```
awschristou, line 327: In case this does get surfaced somewhere unexpectedly, lets make it something less suspicious, like `Client provided profile`

Will-ShaoHua: makes sense thanks Chris

Will-ShaoHua:
fixing test failure

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
```diff
@@ -752,15 +729,7 @@ describe('AmazonQTokenServiceManager', () => {
...
-                const pendingProfileUpdate = amazonQTokenServiceManager.handleOnUpdateConfiguration(
-                    {
-                        section: 'aws.q',
-                        settings: {
-                            profileArn: 'arn:aws:testprofilearn:eu-central-1:11111111111111:profile/QQQQQQQQQQQQ',
-                        },
-                    },
-                    {} as CancellationToken
-                )
+                amazonQTokenServiceManager.setState('PENDING_Q_PROFILE_UPDATE')
```
Will-ShaoHua, line 732: it's hard to set the state to `PENDING_Q_PROFILE_UPDATE` without api call and mock https://github.com/aws/language-servers/pull/1329/files#diff-87976c8ff9a48ba3f54e1313241583da365e5d2bb65e8e2da18b05b9c4652c6eL71-L77, thus replacing with setter

Will-ShaoHua: (also it's less likely to be in the state of `PENDING_Q_PROFILE_UPDATE` as there is no network call anymore)

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
```diff
@@ -62,8 +63,11 @@ describe('AmazonQTokenServiceManager', () => {
 
     let amazonQTokenServiceManager: AmazonQTokenServiceManager
     let features: TestFeatures
+    let clock: SinonFakeTimers
 
     beforeEach(() => {
+        clock = sinon.useFakeTimers()
```
Will-ShaoHua, line 69: will revert (not being used

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
viktorsaws: You can use `await setupServiceManagerWithProfile()` method to move manager to `PENDING_Q_PROFILE_UPDATE` state. 
You can see for it starts in `PENDING_Q_PROFILE_UPDATE` state in `resets to PENDING_Q_PROFILE from PENDING_Q_PROFILE_UPDATE when receiving null profileArn` test case

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
```diff
@@ -829,31 +800,21 @@ describe('AmazonQTokenServiceManager', () => {
...
 
                 amazonQTokenServiceManager.handleOnCredentialsDeleted('bearer')
 
                 assert.strictEqual(amazonQTokenServiceManager.getState(), 'PENDING_CONNECTION')
 
-                await assert.rejects(() => pendingUpdate)
-
                 assert.strictEqual(amazonQTokenServiceManager.getState(), 'PENDING_CONNECTION')
                 assert.strictEqual(amazonQTokenServiceManager.getActiveProfileArn(), undefined)
-                sinon.assert.calledOnce(codewhispererServiceStub.abortInflightRequests)
```
viktorsaws, line 852: this assert shouldn't be removed, as it tests that `abortInflightRequests` call is actually made

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
```diff
@@ -796,19 +776,12 @@ describe('AmazonQTokenServiceManager', () => {
             it('resets to PENDING_Q_PROFILE from PENDING_Q_PROFILE_UPDATE when receiving null profileArn', async () => {
                 await setupServiceManagerWithProfile()
 
-                const pendingUpdate = amazonQTokenServiceManager.handleOnUpdateConfiguration(
```
viktorsaws, line 799: This test doesn't need to change to work

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
viktorsaws: Also, I don't understand why test would change, previous code was moving manager to `PENDING_Q_PROFILE_UPDATE` state already, and it always happened before any API calls were made https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts#L304-L309

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -89,6 +85,11 @@ export class AmazonQTokenServiceManager extends BaseAmazonQServiceManager<
         super(features)
     }
 
+    // @VisibleForTesting, please DO NOT use in production
+    setState(state: 'PENDING_CONNECTION' | 'PENDING_Q_PROFILE' | 'PENDING_Q_PROFILE_UPDATE' | 'INITIALIZED') {
```
viktorsaws, line 89: All tests were working without explicit state management before, I don't understand why that should change. The fix below is modifying only 1 codepath, which shouldn't impact tests execution in other places

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
Will-ShaoHua: nah, after removing the api call in this pr https://github.com/aws/language-servers/pull/1329/files#diff-f61068aaf71fe4a885c144504b7fbacf81cb8e8529bdc00c70474ff3e6e1a058L326, `await setupServiceManagerWithProfile()` will bring the manager to `INITIALIZED`. Another thing is `handleProfileChanged` now will finish running instantly as there is no network call

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -318,21 +320,21 @@ export class AmazonQTokenServiceManager extends BaseAmazonQServiceManager<
...
-            token: token,
-            endpoints: new Map([[parsedArn.region, endpoint]]),
-        })
-
-        this.handleTokenCancellationRequest(token)
-
-        const newProfile = profiles.find(el => el.arn === newProfileArn)
+        // Hack to inject a dummy profile name as it's not used by client IDE for now, if client IDE starts consuming name field then we should also pass both profile name and arn from the IDE
+        // When service is ready to take more tps, revert https://github.com/aws/language-servers/pull/1329 to add profile validation
+        const newProfile: AmazonQDeveloperProfile = {
```
viktorsaws, line 331: When we accept invalid profiles, we need to make sure that when service calls fail due to invalid profile, corresponding error is surfaced to extension from user-facing features. 
Otherwise, we'll show generic errors to customers without a way to recover.

Proper UX would be to return `E_AMAZON_Q_INVALID_PROFILE` LSP response error code to extension, and extension triggers profile selection UX.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
Will-ShaoHua:
```
getListAllAvailableProfilesHandlerStub = sinon
            .stub()
            .resolves(
                Promise.resolve(mockedProfiles).then(() =>
                    new Promise(resolve => setTimeout(resolve, 1)).then(() => mockedProfiles)
                )
            )
```

i think before this change (removal of api call), this stubbing will makes the manager getting into the `PENDING_Q_PROFILE_UPDATE` state

viktorsaws (approved):
Unit tests could be fixed as follow-up

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
Will-ShaoHua: no, removing API call from `handleProfileChanged` will make this line `amazonQTokenServiceManager.handleOnUpdateConfiguration(` immediately `INITIALIZED` instead of pending anymore

Will-ShaoHua: all the test cases involved in this change are due to ^^ reason

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
Will-ShaoHua: yep we aligned it's a followup to surface proper error message

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
Will-ShaoHua: i think we do involve state in the existing test cases, we have many similar snippet like
```
// setup
const pendingUpdate = amazonQTokenServiceManager.handleOnUpdateConfiguration(...)
assert.strictEqual(state, 'PENDING_Q_PROFILE_UPDATE')

// operation to test
amazonQTokenServiceManager.handleOnUpdateConfiguration(...)

// assert
assert.strictEqual(state, xxx)
```

Will-ShaoHua: there are multiple test cases testing things like
1. pending_connection -> pending_profile
2. pending_profile -> intialized
3. pending_profile_update -> initalized
4. pending_profile_update -> pending_profile
etc

Will-ShaoHua: > Also, I don't understand why test would change, previous code was moving manager to PENDING_Q_PROFILE_UPDATE state already, and it always happened before any API calls were made https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts#L304-L309

because  after execution of `const pendingUpdate = amazonQTokenServiceManager.handleOnUpdateConfiguration(`, it will become `INITIALIZED` instead of `PENDING_PROFILE_UPDATE` as before (with API call vs. without API call)

viktorsaws: ok, if that simplifies tests setup I think we could keep it. But we just need to be sure not to modify tests that are actually testing state changes

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.test.ts
Will-ShaoHua: oh yes you are right, thanks for capturing this Victor. I thought it was related to API call so removed it. Reverted

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
Will-ShaoHua: yep agree, definitely not a good pattern tho. Let's follow up on improving this code smell later.

viktorsaws (approved):

PR #1336, opened 2025-05-14
feat: migrate inline completion telemetry to Flare

yzhangok:
## Problem

Comparing to the telemetry sent by VSC plugin, there are some attributes in the telemetries emitted from LS are missing. 

## Solution

Add missing attributes, testing by adding unit test

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
@@ -24,6 +24,7 @@ import {
     ChatConversationType,
     ChatInteractionType,
     ChatTelemetryEventName,
+    CodeWhispererUserModificationEvent,
     CodeWhispererUserTriggerDecisionEvent,
     InteractWithMessageEvent,
 } from './types'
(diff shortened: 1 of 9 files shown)


server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
```diff
@@ -317,7 +326,28 @@ export class TelemetryService {
         modificationPercentage: number
         acceptedCharacterCount: number
         unmodifiedAcceptedCharacterCount: number
+        completionType: string
```
zixlin7, line 329: if these fields are only used for toolkit telemetry can they be moved to the additionalParams, similar to `emitCodeCoverageEvent` and other events?

yzhangok: 1) Moved `completionType`, `triggerType`, `credentialStartUrl` to `additionalParam` in `emitUserModificationEvent`
2) Moved `credentialStartUrl` to `additionalParam` in `emitCodeCoverageEvent`

zixlin7 (approved):

PR #1340, opened 2025-05-14
fix: add robust validation logic to fixHistory

tsmithsz:
## Problem
We are still seeing a high number of `Improperly formed request` errors caused by backend validation check failures on the history.

Main categories:
- UserInputMessage must be followed by AssistantResponseMessage
- UserInputMessage content cannot be blank if there are no ToolResults
- Got ValidationError: ToolResult with ID tooluse_ID has no corresponding ToolUse

We also see some `Improperly formed requests` caused by: `'conversationState.currentMessage.userInputMessage.userInputMessageContext.editorState.workspaceFolders' failed to satisfy constraint: Member must have length less than or equal to 100`

## Solution

Add robust validation checks on the history to match backend checks. Specifically:

- Only validate the last message in history and the incoming user message.
- Enforce user-assistant alternation by removing consecutive messages of the same type.
- Ensure tool results in the new message match tool uses in the previous assistant message.
- Prevents backend requests when validation fails by exiting the agent loop early.
- Applies validations only to new conversations, leaving existing histories unchanged.

In order to prevent race conditions, I:

- Added a unique prompt ID for each chat prompt entered by user to track which prompt is currently active
- Modified the agent loop to check both the prompt ID and cancellation token before modifying chat history. 
             - If cancellation token is true, then that means the user cancelled the request. 
             - If the prompt in the request does not match the current active prompt, means the user entered a new prompt   and we should not add the message.

- Limit workspace folders to 100 to satisfy API constraint

## Testing
- Manual Testing
- Unit Tests
      - Test cases cover backend validation checks:
     

| Test Case    | Backend Validation Check |
| -------- | ------- |
| should handle new user message with missing tool results    | AssistantResponseMessage with ToolUse must be followed by UserInputMessage with ToolResult   |
| should handle new user message with tool results after assistant message without tool uses    | Message with ToolResults must be preceded by AssistantResponseMessage with ToolUse   |
| should handle new user message with invalid tool results ID    | ToolResult with ID {tool_result.toolUseId} has no corresponding ToolUse   |
| should handle new user message with no tool results and blank content    | UserInputMessage content cannot be blank if there are no ToolResults   |

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -3,6 +3,7 @@
  * Will be deleted or merged.
  */
 
+import * as crypto from 'crypto'
 import * as path from 'path'
 import {
     ChatTriggerType,
@@ -451,6 +452,10 @@ export class AgenticChatController implements ChatHandlers {
                 chatResultStream
             )
 
+            // Generate a unique ID for this prompt
+            const promptId = crypto.randomUUID()
+            session.setCurrentPromptId(promptId)
+
             // Start the agent loop
             const finalResult = await this.#runAgentLoop(
                 initialRequestInput,
@@ -458,6 +463,7 @@ export class AgenticChatController implements ChatHandlers {
                 metric,
                 chatResultStream,
                 params.tabId,
+                promptId,
                 session.conversationId,
                 token,
                 triggerContext.documentReference
@@ -541,6 +547,7 @@ export class AgenticChatController implements ChatHandlers {
         metric: Metric<CombinedConversationEvent>,
         chatResultStream: AgenticChatResultStream,
         tabId: string,
+        promptId: string,
         conversationIdentifier?: string,
         token?: CancellationToken,
         documentReference?: FileList
@@ -556,7 +563,8 @@ export class AgenticChatController implements ChatHandlers {
             this.#debug(`Agent loop iteration ${iterationCount} for conversation id:`, conversationIdentifier || '')
 
             // Check for cancellation
-            if (token?.isCancellationRequested) {
+            if (this.#isPromptCanceled(token, session, promptId)) {
+                this.#debug('Stopping agent loop - cancelled by user')
                 throw new CancellationError('user')
             }
 
@@ -570,7 +578,7 @@ export class AgenticChatController implements ChatHandlers {
 
             //  Fix the history to maintain invariants
             if (currentMessage) {
-                const isHistoryValid = this.#chatHistoryDb.fixHistory(
+                const isHistoryValid = this.#chatHistoryDb.fixAndValidateHistory(
                     tabId,
                     currentMessage,
                     conversationIdentifier ?? ''
@@ -607,14 +615,18 @@ export class AgenticChatController implements ChatHandlers {
 
             //  Add the current user message to the history DB
             if (currentMessage && conversationIdentifier) {
-                this.#chatHistoryDb.addMessage(tabId, 'cwc', conversationIdentifier, {
-                    body: currentMessage.userInputMessage?.content ?? '',
-                    type: 'prompt' as any,
-                    userIntent: currentMessage.userInputMessage?.userIntent,
-                    origin: currentMessage.userInputMessage?.origin,
-                    userInputMessageContext: currentMessage.userInputMessage?.userInputMessageContext,
-                    shouldDisplayMessage: shouldDisplayMessage,
-                })
+                if (this.#isPromptCanceled(token, session, promptId)) {
+                    this.#debug('Skipping adding user message to history - cancelled by user')
+                } else {
+                    this.#chatHistoryDb.addMessage(tabId, 'cwc', conversationIdentifier, {
+                        body: currentMessage.userInputMessage?.content ?? '',
+                        type: 'prompt' as any,
+                        userIntent: currentMessage.userInputMessage?.userIntent,
+                        origin: currentMessage.userInputMessage?.origin,
+                        userInputMessageContext: currentMessage.userInputMessage?.userInputMessageContext,
+                        shouldDisplayMessage: shouldDisplayMessage,
+                    })
+                }
             }
             shouldDisplayMessage = true
 
@@ -634,11 +646,13 @@ export class AgenticChatController implements ChatHandlers {
             if (!result.success && result.error.startsWith(responseTimeoutPartialMsg)) {
                 const content =
                     'You took too long to respond - try to split up the work into smaller steps. Do not apologize.'
-                this.#chatHistoryDb.addMessage(tabId, 'cwc', conversationIdentifier ?? '', {
-                    body: 'Response timed out - message took too long to generate',
-                    type: 'answer',
-                    shouldDisplayMessage: false,
-                })
+                if (!this.#isPromptCanceled(token, session, promptId)) {
+                    this.#chatHistoryDb.addMessage(tabId, 'cwc', conversationIdentifier ?? '', {
+                        body: 'Response timed out - message took too long to generate',
+                        type: 'answer',
+                        shouldDisplayMessage: false,
+                    })
+                }
                 currentRequestInput = this.#updateRequestInputWithToolResults(currentRequestInput, [], content)
                 shouldDisplayMessage = false
                 continue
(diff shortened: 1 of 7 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -639,79 +642,123 @@ export class ChatDatabase {
...
-            if (newUserMessage.userInputMessage?.userInputMessageContext?.toolResults) {
-                this.#features.logging.debug('No history message found, but new user message has tool results.')
-                newUserMessage.userInputMessage.userInputMessageContext.toolResults = undefined
-                // tool results are empty, so content must not be empty
-                newUserMessage.userInputMessage.content = 'Conversation history was too large, so it was cleared.'
+            return
+        }
+
+        let i = 0
+        while (i < messages.length) {
```
jguoamz, line 667: do we need to look from the beginning?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -639,79 +642,123 @@ export class ChatDatabase {
             messages.pop()
             this.#features.logging.debug('Dropped trailing user message')
         }
+
+        //  Make sure there are alternating user and assistant messages
+        let i = 0
+        while (i < messages.length - 1) {
```
jguoamz, line 648: can we just look at current and previous?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -639,79 +642,123 @@ export class ChatDatabase {
...
 
-        if (toolResults && toolResults.length > 0) {
             // If last message has no tool uses but new message has tool results, this is invalid
-            if (!lastMsg.toolUses || lastMsg.toolUses.length === 0) {
+            if (toolResults && toolResults.length > 0 && lastMsgToolUses.length === 0) {
                 this.#features.logging.warn('New message has tool results but last message has no tool uses')
-                return false
+                newUserMessageContext.toolResults = []
+                newUserMessage.userInputMessage.content =
+                    'Content was blank and has been auto-filled for validation purposes.'
```
jguoamz, line 737: why are we doing this? This might lead to unexpected behavior from the LLM

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.test.ts
```diff
@@ -0,0 +1,507 @@
...
+            chatDb.ensureValidMessageSequence(messages)
+
+            assert.strictEqual(messages.length, 0, 'Should remove single assistant message')
+        })
+
+        it('should handle complex sequence with multiple issues', () => {
+            const messages: Message[] = [
+                { type: 'answer', body: 'Assistant first message' },
+                { type: 'prompt', body: 'User first message' },
+                { type: 'prompt', body: 'User second message' },
```
jngyuamz, line 133: When can this happen? The agent loop execution should be sequential, but this looks like some race condition. `prompt, prompt, answer, answer`

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -639,79 +642,123 @@ export class ChatDatabase {
...
+                const prevMessage = messages[i - 1]
+                const prevToolUses = prevMessage.toolUses || []
+                if (hasToolResult) {
+                    if (prevToolUses.length === 0) {
+                        this.#features.logging.warn(
+                            `User message at index ${i} has ToolResults but is not preceded by valid Assistant toolUses. Clearing ToolResults.`
+                        )
+
+                        messages[i].userInputMessageContext!.toolResults = []
+                        messages[i].body = 'Content was blank and has been auto-filled for validation purposes.'
```
jguoamz, line 682: same question here

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -639,79 +642,123 @@ export class ChatDatabase {
...
+
+        //  Make sure there are alternating user and assistant messages
+        let i = 0
+        while (i < messages.length - 1) {
+            const currentMessageType = messages[i].type
+            const nextMessageType = messages[i + 1].type
+
+            if (currentMessageType === nextMessageType) {
+                this.#features.logging.warn(`Deleting message at index ${i + 1} to maintain alternating pattern`)
+                messages.splice(i + 1, 1)
```
jguoamz, line 654: I think we should remove the last message and keep the current one

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.test.ts
tsmithsz: This should not occur, but we have seen instances of validation errors due to: `UserInputMessage must be followed by AssistantResponseMessage`. Likely caused by stop button, but we haven't been able to reliably reproduce

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
tsmithsz: Will update the PR

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
tsmithsz: yes, will update

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
tsmithsz: No we don't. No longer trying to account for existing chat tabs with invalid histories. Will only check last Message in history and new UserMessage

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
tsmithsz: Removed

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
tsmithsz: Removed

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
```diff
@@ -26,6 +26,7 @@ export class ChatSessionService {
     public pairProgrammingMode: boolean = true
     public contextListSent: boolean = false
     #abortController?: AbortController
+    #currentPromptId?: string
```
jngyuamz, line 29: Is this variable thread-safe?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -391,13 +391,13 @@ export class ChatDatabase {
...
-     * 5. If the last message is from the assistant and it contains tool uses, and a next user
-     *    message is set without tool results, then the user message will have cancelled tool results.
+     * 2. The first message is from the user. Oldest messages are dropped if needed.
+     * 3. The last message is from the assistant. The last message is dropped if it is from the user.
+     * 4. The history contains alternating sequene of userMessage followed by assistantMessages
+     * 5. The toolUse and toolResult relationship is valid
+     * 6. The history character length is <= MaxConversationHistoryCharacters - newUserMessageCharacterCount. Oldest messages are dropped.
      */
-    fixHistory(tabId: string, newUserMessage: ChatMessage, conversationId: string): boolean {
+    fixandValidateHistory(tabId: string, newUserMessage: ChatMessage, conversationId: string): boolean {
```
jngyuamz, line 400: nit: `fixAndValidateHistory`

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -634,36 +646,42 @@ export class AgenticChatController implements ChatHandlers {
...
-                    toolUses: Object.keys(result.data?.toolUses!)
-                        .filter(k => result.data!.toolUses[k].stop)
-                        .map(k => ({
-                            toolUseId: result.data!.toolUses[k].toolUseId,
-                            name: result.data!.toolUses[k].name,
-                            input: result.data!.toolUses[k].input,
-                        })),
-                    shouldDisplayMessage: shouldDisplayMessage,
-                })
+                if (token?.isCancellationRequested || !session.isCurrentPrompt(promptId)) {
```
jngyuamz, line 663: nit: worth to have a function for it? such as`isPromptCanceled`

jguoamz (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
tsmithsz: I believe it is. I think because the LSP runs on Node's single-threaded event loop, `#currentPromptId` isn't exposed to multi-threaded contention and `ChatSessionService` instances are per tab. Overlapping async `onChatPrompt` calls can overwrite it, but this is what we want.

Please correct me if my understanding is wrong here.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
tsmithsz: Updated

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
tsmithsz: Fixed

jngyuamz (approved):

leigaol (approved):

PR #1341, opened 2025-05-14
telemetry(amazonq): adding new telemetry metrics and addtional fields for existing metrics

laileni-aws:
## Problem
- Existing metrics in agentic chat is missing some fields like agentic coding mode etc.
- Missing tool use latency.
## Solution
- Addressed the above issues.
- adding a new metric `amazonq_InvokeLLM`
- Implemented additional metrics from these two PR's 
  - https://github.com/aws/aws-toolkit-common/pull/1025
  - https://github.com/aws/aws-toolkit-common/pull/1027

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -62,6 +62,7 @@ import {
 import { v4 as uuid } from 'uuid'
 import {
     AddMessageEvent,
+    ChatConversationType,
     ChatInteractionType,
     ChatTelemetryEventName,
     CombinedConversationEvent,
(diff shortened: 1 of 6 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -657,34 +671,76 @@ export class AgenticChatController implements ChatHandlers {
 
             if (pendingToolUses.length === 0) {
                 // No more tool uses, we're done
+                this.#telemetryController.emitAgencticLoop_InvokeLLM(
+                    response.$metadata.requestId!,
+                    conversationId,
+                    'AgenticChat',
```
chungjac, line 677: Whether LLM was or was not apart of agentic loop, it will eventually arrive here. So it can also be 'AgenticChatWithToolUse'

laileni-aws: This metric is per LLM call so This is used to check is this LLM call has toolUse or not

chungjac (approved):

chungjac (approved):

PR #1343, opened 2025-05-14
fix: fix to remove approved paths for fsWrite

ashishrp-aws:
## Problem
Security requirement to ask for permission for every file modification outisde of workpsace scope.

## Solution
- Skipping approvedPaths check for fsWrite tool.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolShared.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolShared.ts
@@ -114,9 +114,15 @@ export async function requiresPathAcceptance(
     path: string,
     lsp: Features['lsp'],
     logging: Features['logging'],
-    approvedPaths?: Set<string>
+    approvedPaths?: Set<string>,
+    toolName?: string
 ): Promise<CommandValidation> {
     try {
+        // Skip check for fsWrite tool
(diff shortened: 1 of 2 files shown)


laileni-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolShared.ts
```diff
@@ -114,12 +114,16 @@ export async function requiresPathAcceptance(
...
+    toolName?: string
 ): Promise<CommandValidation> {
     try {
         // First check if the path is already approved
-        if (isPathApproved(path, approvedPaths)) {
-            return { requiresAcceptance: false }
+        // Skip this check only if toolName is explicitly 'fsWrite'
+        if (!toolName || toolName !== 'fsWrite') {
+            if (isPathApproved(path, approvedPaths)) {
+                return { requiresAcceptance: false }
```
yueny2020, line 125: after this change, we always ask for permission for fsWrite. I think we should have two approvedPaths sets. one is for read and another is for write.

ashishrp-aws: yes...once we confirm the final requirement, will adjust and bifurcate the approvedPaths for read and write tools.

PR #1348, opened 2025-05-15
feat: launch one remote workspace for all workspace folders

LiGaCu:
## Problem

For server-side workspace context capability, we would like to consolidate all the workspace folders under the same IDE workspace into a single remote workspace so that we could provide better workspace-level context for cross-workspace-folder scenarios.

Prerequisite changes have been made:
- https://github.com/aws/aws-toolkit-vscode/pull/7291
- https://github.com/aws/aws-toolkit-jetbrains/pull/5725

## Solution

Updated the fundamental logic of `workspaceContext/workspaceFolderManager` to map all the events into a single remote workspace. Modified `workspaceContext/workspaceContextServer` logic to accept the workspace identifier passed from extensions before kicking off the workflow. Modified `DependencyDiscoverer`, `LanguageDependencyHandler`, `ArtifactManager`, etc. to accommodate the logic updates.

Performed E2E testing and verified the logic works well.

## License


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyDiscoverer.ts
```diff
@@ -12,6 +12,8 @@ export class DependencyDiscoverer {
     private workspaceFolders: WorkspaceFolder[]
     public dependencyHandlerRegistry: LanguageDependencyHandler<BaseDependencyInfo>[] = []
     private initializedWorkspaceFolder = new Map<WorkspaceFolder, boolean>()
+    // Create a SharedArrayBuffer with 4 bytes (for a 32-bit unsigned integer) for thread-safe counter
```
ege0zcan, line 15: Just for posterity, I believe we don't need to  go with this approach since we are not using any workers or threads in the code and everything is single threaded at the moment

a simple data type like the following would also work here 

```
type DependencySizeTracker = {
    totalSize: number
    maxSize: number
}

const sizeTracker: DependencySizeTracker = {
    totalSize: 0,
    maxSize: 8 * 1024 * 1024 * 1024
}
```

The current approach is still fine and I am not suggesting we change it.

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyHandler/LanguageDependencyHandler.ts
```diff
@@ -308,14 +315,16 @@ export abstract class LanguageDependencyHandler<T extends BaseDependencyInfo> {
 
     dispose(): void {
         this.dependencyMap.clear()
-        this.dependencyUploadedSize.clear()
+        Atomics.store(this.dependencyUploadedSizeSum, 0, 0)
```
ege0zcan, line 318: This should be done once inside the dispose function of the dependencyDiscoverer instead of being done inside each individual dependency handler multiple times

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
```diff
@@ -529,165 +388,160 @@ export class WorkspaceFolderManager {
...
-                    default:
-                        this.logging.warn(`Unknown workspace status: ${metadata.workspaceStatus}`)
-                }
-            } catch (error) {
-                this.logging.error(`Error monitoring workspace ${workspace}: ${error}`)
-            }
-        }, this.CONTINUOUS_MONITOR_INTERVAL)
+        // Otherwise, wait for the promise to resolve or catch the rejection and retry
+        try {
+            return await this.remoteWorkspaceIdPromise
```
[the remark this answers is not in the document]
ege0zcan: The `waitForRemoteWorkspaceId` function is currently being called inside an interval right now from `initializeWorkspaceStatusMonitor`

If the waitForRemoteWorkspaceId call goes into an infinite cycle, new intervals would start getting created before the previous one resolves. So the number of active intervals would linearly increase since none of the open intervals would resolve. This could lead to very unpleasant UX for the unhappy paths

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
```diff
@@ -99,6 +110,32 @@ export class WorkspaceFolderManager {
...
+        }
+
+        this.messageQueueConsumerInterval = setInterval(() => {
+            if (this.workspaceState.webSocketClient && this.workspaceState.webSocketClient.isConnected()) {
+                const message = this.workspaceState.messageQueue[0]
+                if (message) {
+                    this.workspaceState.webSocketClient
+                        .send(message)
+                        .then(() => {
+                            this.logging.log(` Message sent successfully`)
```
ege0zcan, line 130: styling: unnecessary space before the log

ege0zcan: also the log is somewhat ambiguous, would be helpful to at least specify that this is a websocket message

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
```diff
@@ -99,6 +110,32 @@ export class WorkspaceFolderManager {
...
+                            this.workspaceState.messageQueue.shift()
+                        })
+                        .catch(error => {
+                            this.logging.error(`Error sending message: ${error}`)
+                        })
+                }
+            }
+        }, this.MESSAGE_PUBLISH_INTERVAL)
```
ege0zcan, line 138: Given we now have the websocket message frequency being controlled by the workspace folder manager, I believe we should be able to simplify and remove the wrong throttling that we have on the websocket client

We should remove this line -> https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/client.ts#L159 and make the `send` function synchronous. It will also help simplify rest of the code where send was being called

Line 159 linked above is not useful even now since it applies a 100 ms delay to every websocket send call. So if `send` is called with 5 different messages at the same time, all the calls will wait 100 ms and they will all be sent at the same time meaning that the existing delay is actually completely redundant.

The scenario explained in the comment https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/client.ts#L144-L156 is too extreme of an example and as explained in the paragraph above, the artificial delay is redundant

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
```diff
@@ -273,93 +249,42 @@ export class WorkspaceFolderManager {
...
+                                    uri: folder.uri,
+                                    name: folder.name,
                                 },
-                            })
-                        )
-                        .catch(e => {
-                            this.logging.error(`Error sending didChangeWorkspaceFolders message: ${e}`)
-                        })
-                }
-                websocketClient.disconnect()
```
ege0zcan, line 311: is there any case where we would want to disconnect the websocket client to be more efficient with resource usage? For example if the user removes all workspace folders from their workspace. If the user adds a new workspace folder after this happens, the code should probably be able to re-open a new ws connection

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
```diff
@@ -49,7 +57,8 @@ export class WorkspaceFolderManager {
         artifactManager: ArtifactManager,
         dependencyDiscoverer: DependencyDiscoverer,
         workspaceFolders: WorkspaceFolder[],
-        credentialsProvider: CredentialsProvider
+        credentialsProvider: CredentialsProvider,
+        workspaceIdentifier: string
```
ege0zcan, line 61: As a followup we should also update the commented out logic for inline completions so that inline completions server will also be ready to use the workspaceIdentifier

https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts#L52

https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts#L84-L89

https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts#L349-L350

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
```diff
@@ -800,130 +606,50 @@ export class WorkspaceFolderManager {
...
                         workspaceChangeMetadata: {
-                            workspaceId: workspaceId ?? '',
+                            workspaceId: this.workspaceState.workspaceId,
                             s3Path: cleanUrl(s3Url),
                             programmingLanguage: fileMetadata.language,
                         },
                     },
                 })
 
                 // We add this event to the front of the queue here to prevent any race condition that might put events before the didChangeWorkspaceFolders event
```
ege0zcan, line 644: this comment can be removed from here now that we are no longer adding the message to the front of the queue anymore. This was a best effort mechanism to decrease the likelihood of sending other messages before the `didChangeWorkspaceFolders` message but it is not a strong mechanism. 

As far as I can see it is still possible that if the ws connection gets established before than the source code preparation is finished, any LSP event like didSave event can be sent before the `didChangeWorkspaceFolders` event. 

Maybe we should add a flag to the websocket publishing logic in workspace folder manager to keep track of the initial didChangeWorkspaceFolders being sent and until the initial didChangeWorkspaceFolders event(s) is/are sent, the websocket messages would be queued. Not suggesting to do it in this PR but would be a nice followup

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyHandler/LanguageDependencyHandler.ts
LiGaCu: Make sense, will move it.

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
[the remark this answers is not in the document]
LiGaCu: > The waitForRemoteWorkspaceId function is currently being called inside an interval right now from initializeWorkspaceStatusMonitor

Not really, neither ``initializeWorkspaceStatusMonitor`` nor ``checkRemoteWorkspaceStatusAndReact`` would call ``waitForRemoteWorkspaceId``. So, none of these intervals would get blocked by ``remoteWorkspaceIdPromise`` and no loop would be formed.

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
LiGaCu: Make sense, will remove the wait within `WebSocketClient`

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
LiGaCu: yeah, those are the improvements we could consider separately.

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/client.ts
```diff
@@ -141,23 +144,10 @@ export class WebSocketClient {
...
-    // - Both check timeSinceLastMessage at nearly the same time
-    // - Both might determine they need to wait
-    // - They could end up sending in unpredictable order
-    // It might be better to keep an active queue in the client and expose enqueueMessage instead of send
-    public async send(message: string): Promise<void> {
+    public send(message: string): void {
         if (this.ws?.readyState === WebSocket.OPEN) {
-            await new Promise(resolve => setTimeout(resolve, this.messageThrottleDelay))
             this.ws.send(message)
+            this.logging.debug('Message sent successfully to the remote workspace')
```
ege0zcan, line 150: reminder that debug logs are disabled by default on production

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/client.ts
LiGaCu: yeah, not seeing these messages valuable to present in user logs.

PR #1351, opened 2025-05-16
fix: Reduce perceived latency of fsWrite. Show fsWrite errors in the UX 

leigaol:
## Problem

1. Need to show fsWrite errors in the UX , not just in the messages.
2. Need to show fsWrite is starting before the entire file write is done.

## Solution

1. Display fsWrite errors in the UX according to the design

https://github.com/user-attachments/assets/19ca5464-8481-417c-a2d2-08da299b1fa4

3. Render fsWrite progress

https://github.com/user-attachments/assets/0b03604e-fc9a-47a9-b00b-5937c1fd9056

https://github.com/user-attachments/assets/c6b911f1-10a2-4473-b0db-3a2ca97d2c15

https://github.com/user-attachments/assets/a3fd3a76-9ed1-4ac5-87e1-de2234753bfc

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -2270,7 +2270,10 @@ export class AgenticChatController implements ChatHandlers {
                         },
                     })
                 }
-                // render the tool use explanatory as soon as this is received.
+            }
+
+            // render the tool use explanatory as soon as this is received for all tool uses.
+            if (typeof toolUse.input === 'string') {
                 const explanation = extractKey(toolUse.input, 'explanation')
                 const messageId = progressPrefix + toolUse.toolUseId + '_explanation'
                 if (explanation && !chatResultStream.hasMessage(messageId)) {
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -836,8 +836,7 @@ export class AgenticChatController implements ChatHandlers {
                 }
 
                 // remove progress UI
-                const progressMessageId = progressPrefix + toolUse.toolUseId
-                await chatResultStream.removeResultBlockAndUpdateUI(progressMessageId)
+                await chatResultStream.removeResultBlockAndUpdateUI(progressPrefix + toolUse.toolUseId)
 
                 // fsRead and listDirectory write to an existing card and could show nothing in the current position
                 if (!['fsWrite', 'fsRead', 'listDirectory'].includes(toolUse.name)) {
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -834,18 +834,15 @@ export class AgenticChatController implements ChatHandlers {
...
                     await this.#showUndoAllIfRequired(chatResultStream, session)
                 }
-                const { explanation } = toolUse.input as unknown as ExplanatoryParams
-                if (explanation) {
-                    await chatResultStream.writeResultBlock({
-                        type: 'directive',
-                        messageId: toolUse.toolUseId + '_explanation',
-                        body: explanation,
-                    })
-                }
```
jguoamz, line 848: wouldn't this remove the explanation showing for other tools such as executeBash?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -834,18 +834,15 @@ export class AgenticChatController implements ChatHandlers {
                 if (!availableToolNames.includes(toolUse.name)) {
                     throw new Error(`Tool ${toolUse.name} is not available in the current mode`)
                 }
+
+                // remove progress UI
+                await chatResultStream.removeResultBlockAndUpdateUI(progressPrefix + toolUse.toolUseId)
```
jguoamz, line 839: wouldn't this break other tool?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -2203,6 +2230,61 @@ export class AgenticChatController implements ChatHandlers {
         }
     }
 
+    async #showToolUseIntemediateResult(
```
tsmithsz, line 2233: nit: `#showToolUseIntermediateResult`

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
leigaol: This only removes the messages that has the progressPrefix, which is the in progress rendered UX. When the streaming result is finalized, we no longer need these in progress ux.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
leigaol: Explanation is now rendered as soon as they are streamed even in the incomplete json response, but I do missed a if else condition.

leigaol: Added another video to explain bash.

tsmithsz (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
jguoamz: I think the UX is worse for executeBash now, can we do this only for fsWrite?

PR #1353, opened 2025-05-16
fix(amazonq): Use common utility to determine workspaceFolders and fix tests

shruti0085:
## Description
This change address a few different things concerning the usage of `workspaceFolders` across servers.
* Language server runtime version has been bumped to the latest to pull changes made in https://github.com/aws/language-server-runtimes/pull/509
* All locations that determine workspaceFolders have been updated to call `workspace.getAllWorkspaceFolders()` which was introduced in the runtimes PR as a common utility(it respects `didChangeWorkspaceFolder` event alongwith using rootUri of initializeparams in absence of `workspaceFolder`)
* With this update, particularly in the VS client paths associated with tool actions are now correct. Prior to that they were displayed as a relativePath.
* InitializeUtils has been deleted as the logic has been moved to the runtimes repo

In addition, this change fixes test failures seen locally on Windows devices associated with  `workspaceUtils.test.ts` and `fileSearch.test.ts` whereever `fs.readDir` calls were being stubbed out. Error `path is undefined` was being thrown for tests where `getEntryPath` was invoked. 

## License


server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
```diff
@@ -4,20 +4,19 @@ import { TelemetryService } from '../../shared/telemetry/telemetryService'
...
+    ({ credentialsProvider, telemetry, logging, lsp, workspace }) => {
         let localProjectContextController: LocalProjectContextController
         let amazonQServiceManager: AmazonQTokenServiceManager
         let telemetryService: TelemetryService
 
         let localProjectContextEnabled: boolean = false
 
         lsp.addInitializer((params: InitializeParams) => {
-            const workspaceFolders = getWorkspaceFolders(logging, params)
+            const workspaceFolders = workspace.getAllWorkspaceFolders() || params.workspaceFolders
```
awschristou, line 19: Should this be `??` I'm a little rusty...

Similarly in workspaceContextServer.ts

core/aws-lsp-core/src/util/workspaceUtils.test.ts
```diff
@@ -224,7 +236,14 @@ describe('workspaceUtils', function () {
             tempFolder = await TestFolder.create()
             testFeatures = new TestFeatures()
             // Taken from https://github.com/aws/language-server-runtimes/blob/674c02696c150838b4bc93543fb0009c5982e7ad/runtimes/runtimes/standalone.ts#L216
-            testFeatures.workspace.fs.readdir = path => fs.readdir(path, { withFileTypes: true })
+            testFeatures.workspace.fs.readdir = async dirPath => {
+                const entries = await fs.readdir(dirPath, { withFileTypes: true })
+                // Add parentPath to each entry
+                return entries.map(entry => {
+                    ;(entry as any).parentPath = dirPath
```
ege0zcan, line 243: just curious, why didn't we need parentPath information before but now we need it? 

As far as I can see `readDirectoryRecursively` calls `getEntryPath` which uses `parentPath` info so I can see why this might be needed but I wonder why this was not needed before and is needed now

core/aws-lsp-core/src/util/workspaceUtils.ts
```diff
@@ -99,12 +99,16 @@ export function formatListing(entry: Dirent): string {
 }
 
 export function getEntryPath(entry: Dirent) {
+    if (!entry.parentPath) {
```
ege0zcan, line 102: doesn't typescript satisfy this automatically since `parentpath` is a non-optional field in Dirent? 

The only place where the parentpath was not being set before was in the testing stubs that had custom behaviour and that were doing lots of type casting. Throwing and error here doesn't look really right to me, now with the parentpath changes introduced in the PR do we need to throw this error?

ege0zcan (approved):

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
shruti0085: From what I could find, it seems like it the right hand side would be used as fallback when left-hand side is any falsy value (including false, 0, "", null, undefined, and NaN). That seemed fine to me

core/aws-lsp-core/src/util/workspaceUtils.ts
shruti0085: I will remove the throw and error check in getEntryPath as I was seeing errors locally where parentpath was null with tests. I wasn't sure if this only impacted tests but as long as it does not impact actual logic, I am fine with removing this

ege0zcan (approved):

PR #1355, opened 2025-05-17
fix: re-categorize error status code

yueny2020:
## Problem
- error status code in Kibana is 0 or has wrong value

## Solution
- re-categorize error status code

## License


tsmithsz (commented):
What about flipping `Improperly formed request` to 500?

tsmithsz (approved):

yueny2020:
> What about flipping `Improperly formed request` to 500?

discussed offline, the 400 error `Improperly formed request` is between LSP and RTS. From customer perspective, this should be 500. will make the change in a follow up pr.

PR #1359, opened 2025-05-19
fix: add grepSearch implementation

jguoamz:
## Problem
- We are slow and inefficient when resolving tasks
 
## Solution
- Add grepSearch implementation
- This will not enable the tool

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -109,6 +109,7 @@ import { FsWrite, FsWriteParams } from './tools/fsWrite'
 import { ExecuteBash, ExecuteBashParams } from './tools/executeBash'
 import { ExplanatoryParams, ToolApprovalException } from './tools/toolShared'
 import { FileSearch, FileSearchParams } from './tools/fileSearch'
+import { GrepSearch, SanitizedRipgrepOutput } from './tools/grepSearch'
 import { loggingUtils } from '@aws/lsp-core'
 import { diffLines } from 'diff'
 import {
@@ -926,6 +927,7 @@ export class AgenticChatController implements ChatHandlers {
                     case 'fsRead':
                     case 'listDirectory':
                     case 'fileSearch':
+                    case 'grepSearch':
                     case 'fsWrite':
                     case 'executeBash': {
                         const toolMap = {
@@ -933,6 +935,7 @@ export class AgenticChatController implements ChatHandlers {
                             listDirectory: { Tool: ListDirectory },
                             fsWrite: { Tool: FsWrite },
                             executeBash: { Tool: ExecuteBash },
+                            grepSearch: { Tool: GrepSearch },
                             fileSearch: { Tool: FileSearch },
                         }
 
@@ -1042,8 +1045,11 @@ export class AgenticChatController implements ChatHandlers {
                         // no need to write tool result for listDir and fsRead into chat stream
                         // executeBash will stream the output instead of waiting until the end
                         break
-                    case 'codeSearch':
-                        // no need to write tool result for code search.
+                    case 'grepSearch':
+                        const grepSearchResult = this.#processGrepSearchResult(toolUse, result, chatResultStream)
+                        if (grepSearchResult) {
+                            await chatResultStream.writeResultBlock(grepSearchResult)
+                        }
                         break
                     case 'fsWrite':
                         const input = toolUse.input as unknown as FsWriteParams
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1716,6 +1722,73 @@ export class AgenticChatController implements ChatHandlers {
...
+                lineRanges: [{ first: -1, second: -1 }],
+            }
+        }
+
+        // Create sorted array of file paths
+        const sortedFilePaths = Object.keys(fileDetails)
+
+        // Create the context list for display
+        const query = (toolUse.input as any)?.query || 'search term'
+        const matchCount = matches.length
```
laileni-aws, line 1776: Do we need this matchCount var?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1716,6 +1722,73 @@ export class AgenticChatController implements ChatHandlers {
...
+        const matches = output.fileMatches
+        const fileDetails: Record<string, FileDetails> = {}
+
+        // Create file details directly from matches
+        for (const match of matches) {
+            const filePath = match.filePath
+            if (!filePath) continue
+
+            fileDetails[filePath] = {
+                description: `${match.matches.length} matches`,
```
laileni-aws, line 1766: description should show complete filePath IIRC.
There can be same fileName in two different location, users need to know which exact file is being used for context.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
jguoamz: good catch, will remove

grl-wang: yes, following is the UI proposed by Ankit. and we have discussion later to change `grepped` to `searched`
image (6)

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/grepSearch.ts
```diff
@@ -0,0 +1,315 @@
...
+                }
+            }
+
+            if (params.excludePattern) {
+                // Support multiple exclude patterns
+                const patterns = params.excludePattern.split(',')
+                for (const pattern of patterns) {
+                    args.push(`-g`, `!${pattern.trim()}`)
+                }
+            }
```
grl-wang, line 166: we should also update the tool spec and implementation to ignore the common build and dependency directories.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/grepSearch.ts
```diff
@@ -0,0 +1,315 @@
...
+            // Add include/exclude patterns
+            if (params.includePattern) {
+                // Support multiple include patterns
+                const patterns = params.includePattern.split(',')
+                for (const pattern of patterns) {
+                    args.push('-g', `${pattern.trim()}`)
+                }
+            }
+
+            if (params.excludePattern) {
```
Randall-Jiang, line 160: what if user pass an empty string to excludePattern/includePattern?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
grl-wang: as the grep search using the same UI as read file, the complete filePath will show when user hover over the file. In this case, I don't think we need to show the completed filePath explicitly.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/grepSearch.ts
jguoamz: we don't ignore them

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/grepSearch.ts
jguoamz: then it will be false

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/grepSearch.ts
jguoamz: ripgrep respects gitignore rules already and there is no problem searching for more files because it's so fast

grl-wang (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
laileni-aws: In the file read and list we show entire path If user hovers thats a appsec requirement

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1733,7 +1805,7 @@ export class AgenticChatController implements ChatHandlers {
         updatedRequestInput.conversationState!.currentMessage!.userInputMessage!.content = content
 
         for (const toolResult of toolResults) {
-            this.#debug(`ToolResult: ${JSON.stringify(toolResult)}`)
+            this.#debug(`ToolResult: ${JSON.stringify(toolResult)} `)
```
justinmk3, line 1808: @jguoamz why add space at the end all of these logs ? same for other cases in this PR.

PR #1363, opened 2025-05-19
feat: add update server permission method without re-init server

bywang56:
## Problem
update server shouldn't re-init server
## Solution
add update server permission method without re-init server

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpManager.ts
@@ -12,6 +12,7 @@ import type {
     ListToolsResponse,
     McpServerRuntimeState,
     McpServerStatus,
+    MCPServerPermissionUpdate,
 } from './mcpTypes'
 import { loadMcpServerConfigs } from './mcpUtils'
 import { AgenticChatError } from '../../errors'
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpManager.ts
```diff
@@ -434,7 +435,56 @@ export class McpManager {
...
+
+        if (perm.disabled !== undefined && newDisabled !== oldDisabled) {
+            // Server level ENABLED → DISABLED
+            if (newDisabled) {
+                const client = this.clients.get(serverName)
+                if (client) {
+                    await client.close()
+                    this.clients.delete(serverName)
+                }
+                this.setState(serverName, 'DISABLED', 0)
```
ashishrp-aws, line 472: why `0`?

bywang56: I think this is for tools that customer expect to be available when calling LLM right?, so we should set to 0 here if server is disabled entirely.

ashishrp-aws: I think we have show tool numbers regardless whether they are disabled or enabled for better UX....we can talk to product or UX design for this.

ashishrp-aws (approved):

PR #1372, opened 2025-05-19
fix: Truncate API request first, then truncate chat history

leigaol:
## Problem
This is a follow up of https://github.com/aws/language-servers/pull/1368.

The history truncation is already done in https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts#L450

## Solution

Do not truncate history message by length

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -830,20 +830,6 @@ export class AgenticChatController implements ChatHandlers {
             request.conversationState.currentMessage.userInputMessage.userInputMessageContext.editorState.document =
                 truncatedCurrentDocument
         }
```
tsmithsz, line 832: We need to factor in the existing history length into the `remainingCharacterBudget` right?

leigaol: updated.

tsmithsz (approved):

PR #1373, opened 2025-05-19
fix: missing handle connection expired error for inline suggestions

zixlin7:
## Problem
https://github.com/aws/language-servers/pull/1001 (connection expired error handling) was negated by https://github.com/aws/language-servers/pull/1299
## Solution
add back changes in #1001 and add unit test

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
@@ -592,6 +592,14 @@ export const CodewhispererServerFactory =
             logging.log('Recommendation failure: ' + error)
             emitServiceInvocationFailure(telemetry, session, error)
 
+            if (error instanceof AmazonQError) {
+                throw error
+            }
+
+            if (hasConnectionExpired(error)) {
+                throw new AmazonQServiceConnectionExpiredError(getErrorMessage(error))
+            }
+
             sessionManager.closeSession(session)
 
             return EMPTY_RESULT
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
```diff
@@ -592,6 +592,14 @@ export const CodewhispererServerFactory =
             logging.log('Recommendation failure: ' + error)
             emitServiceInvocationFailure(telemetry, session, error)
 
+            if (error instanceof AmazonQError) {
+                throw error
+            }
+
+            if (hasConnectionExpired(error)) {
+                throw new AmazonQServiceConnectionExpiredError(getErrorMessage(error))
```
awschristou, line 600: image
Some of the missing code involves wrapping the exception and returning the standardized `ResponseError` (which is part of the standard LSP Protocol). Is ResponseError being returned from somewhere higher in the call chain, or does this code require some sort of wrapping?

zixlin7: thanks! I added the translate logic back in latest commit, not sure why it was originally nested but I think this should have the same effect?

server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
```diff
@@ -594,6 +594,22 @@ export const CodewhispererServerFactory =
...
+                translatedError = new AmazonQServiceConnectionExpiredError(getErrorMessage(error))
+            }
+
+            if (translatedError instanceof AmazonQError) {
+                throw new ResponseError(
+                    LSPErrorCodes.RequestFailed,
+                    error.message || 'Error processing suggestion requests',
+                    {
+                        awsErrorCode: (error as AmazonQError).code,
+                    }
```
awschristou, line 609: `message` and `code` are pulled from `error`. 
Should these be pulled from `translatedError`?

awschristou (approved):

PR #1374, opened 2025-05-19
Refactor history logic, apply a max 200MB total history size

jngyuamz:
## Problem
1. Current history logic is mixing up the history sent to service and the history persisted in disk.
2. The limit on history size is 600,000 chars or 250 messages, which is way to small.

## Solution
1. Separate the "fix history" logic for service requests and DB
    - History for service requests:
       - Length is less than 250 messages(ensured when retrieving from DB) and 600,000 chars
       - Drop the last empty assistant message and the user prompt before that
       - Ensure the messages have a valid sequence - alternating sequence of user prompt followed by a assistant message
       - Validate and fix tool usage
     - Persisted history in DB:
        - addMessage now skips empty assistant message
        - (Already in place) Skip adding message if the agent loop is cancelled, this is likely the root cause of incorrect message sequence.
2. Maintain a max 200MB total history file size
    - An async check/trim is performed on startup
       - If the total valid history size under the history folder is larger than 200MB, remove the oldest messages across all history files until the total history size is under 75% of the limit. The implementation uses a priority queue to order the messages. The process takes 1.5 sec when avg message size is 5000 chars.
       - Added timestamps to persisted conversations and messages
       - Legacy records don't have timestamps in conversations and messages, use the updatedAt under Tab as the timestamp for all those messages when ordering

## Future improvements
1. Delete empty history file
2. Telemetry for history trimming time

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -1806,7 +1806,7 @@ export class AgenticChatController implements ChatHandlers {
         updatedRequestInput.conversationState!.currentMessage!.userInputMessage!.content = content
 
         for (const toolResult of toolResults) {
-            this.#debug(`ToolResult: ${JSON.stringify(toolResult)} `)
+            this.#debug(`ToolResult: ${JSON.stringify(toolResult)}`)
             updatedRequestInput.conversationState!.currentMessage!.userInputMessage!.userInputMessageContext!.toolResults.push(
                 {
                     ...toolResult,
@@ -1941,13 +1941,13 @@ export class AgenticChatController implements ChatHandlers {
         }
 
         if (authFollowType) {
-            this.#log(`Q auth error: ${getErrorMessage(err)} `)
+            this.#log(`Q auth error: ${getErrorMessage(err)}`)
 
             return createAuthFollowUpResult(authFollowType)
         }
 
         if (customerFacingErrorCodes.includes(err.code)) {
-            this.#features.logging.error(`${loggingUtils.formatErr(err)} `)
+            this.#features.logging.error(`${loggingUtils.formatErr(err)}`)
             if (err.code === 'InputTooLong') {
                 // Clear the chat history in the database for this tab
                 this.#chatHistoryDb.clearTab(tabId)
@@ -1964,10 +1964,10 @@ export class AgenticChatController implements ChatHandlers {
                 buttons: [],
             })
         }
-        this.#features.logging.error(`Unknown Error: ${loggingUtils.formatErr(err)} `)
+        this.#features.logging.error(`Unknown Error: ${loggingUtils.formatErr(err)}`)
         return new ResponseError<ChatResult>(LSPErrorCodes.RequestFailed, err.message, {
             type: 'answer',
-            body: requestID ? `${genericErrorMsg} \n\nRequest ID: ${requestID} ` : genericErrorMsg,
+            body: requestID ? `${genericErrorMsg} \n\nRequest ID: ${requestID}` : genericErrorMsg,
             messageId: errorMessageId,
             buttons: [],
         })
@@ -2003,10 +2003,10 @@ export class AgenticChatController implements ChatHandlers {
             this.#log('Response for inline chat', JSON.stringify(response.$metadata), JSON.stringify(response))
         } catch (err) {
             if (err instanceof AmazonQServicePendingSigninError || err instanceof AmazonQServicePendingProfileError) {
-                this.#log(`Q Inline Chat SSO Connection error: ${getErrorMessage(err)} `)
+                this.#log(`Q Inline Chat SSO Connection error: ${getErrorMessage(err)}`)
                 return new ResponseError<ChatResult>(LSPErrorCodes.RequestFailed, err.message)
             }
-            this.#log(`Q api request error ${err instanceof Error ? JSON.stringify(err) : 'unknown'} `)
+            this.#log(`Q api request error ${err instanceof Error ? JSON.stringify(err) : 'unknown'}`)
             return new ResponseError<ChatResult>(
                 LSPErrorCodes.RequestFailed,
                 err instanceof Error ? err.message : 'Unknown request error'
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -388,86 +421,256 @@ export class ChatDatabase {
...
+            `History total size (${historyTotalSizeInBytes} Bytes) exceeds limit (${maxHistorySizeInBytes} Bytes), trimming history`
+        )
+
+        const trimStart = performance.now()
+        await this.trimHistoryForAllWorkspace()
+        const trimEnd = performance.now()
+        this.#features.logging.info(`Trimming history took ${trimEnd - trimStart} ms`)
+    }
+
+    private async trimHistoryForAllWorkspace() {
```
avi-alpert, line 536:
```suggestion
    private async trimHistoryForAllWorkspaces() {
```

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -388,86 +421,256 @@ export class ChatDatabase {
...
+            this.#features.logging.error(`Error listing database files: ${err}`)
+            return []
+        }
+    }
+
+    /**
+     * If the sum of all history file size exceeds the limit, start trimming the oldest conversation
+     * across all the workspace until the folder size is below maxAfterTrimHistoryFolderSizeInBytes.
+     */
+    async trimHistoryToMaxSize() {
```
avi-alpert, line 515: this new history cleanup mechanism is very complex and makes this file very long and hard to read - could you separate out this functionality into a separate file that can be imported into the ChatDatabase class?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -679,6 +683,13 @@ export class AgenticChatController implements ChatHandlers {
                                 input: result.data!.toolUses[k].input,
                             })),
                         shouldDisplayMessage: shouldDisplayMessage,
+                        timestamp: new Date(),
+                    })
+
+                    // Async process: Trimming history asynchronously if the size exceeds the max
+                    // This process will take several seconds
+                    this.#chatHistoryDb.trimHistoryToMaxSize().catch(err => {
```
avi-alpert, line 691: is this process to check file sizes of all history files happening every time the agent loop? even if its not blocking, i think its overkill to run it every time. can we make this check more periodical, perhaps only on startup

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
jngyuamz: ack

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
jngyuamz: It's on my list actually. There are two dimensions for readability improvement:
1. this mechanism/method itself is complex, I'll need to see if I can simplify it(might be difficult though).
2. chatDb is way too long, ideally, we should only keep DB related operations in this file
    - Separate this mechanism out
    - (Followup PR) Move irrelevant logic to `util` (such as count characters, validation methods, etc.)

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
jngyuamz: Agreed, I'll just move it to startup

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -24,8 +26,24 @@ import { ChatItemType } from '@aws/mynah-ui'
 import { getUserHomeDir } from '@aws/lsp-core/out/util/path'
 
 export const EMPTY_CONVERSATION_LIST_ID = 'empty'
-// Maximum number of messages to keep in history
-const MaxConversationHistoryMessages = 250
+// Maximum number of characters to send in request
+const maxConversationHistoryCharacters = 600_000
```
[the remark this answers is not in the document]
jngyuamz: Yeah, that's a merge conflict issue, and `calculateCurrentMessageCharacterCount` is not being used after that change as well.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -574,25 +575,26 @@ export class AgenticChatController implements ChatHandlers {
...
-                const isHistoryValid = this.#chatHistoryDb.fixAndValidateHistory(
+                // Get preprocessed history messages from DB with size and number of character limits
+                let messages = this.#chatHistoryDb.getPreprocessedRequestHistory(
                     tabId,
                     currentMessage,
-                    conversationIdentifier ?? '',
                     remainingCharacterBudget
                 )
-                if (!isHistoryValid) {
+                if (!this.#chatHistoryDb.validateAndFixNewMessageToolResults(messages, currentMessage)) {
```
tsmithsz, line 587: Calling `validateAndFixNewMessageToolResults` after `getPreprocessedRequestHistory` might result in incorrect history length calculations. We can potentially add or remove characters, the former resulting in errors if we exceed the 600K character limit.

tsmithsz: Ideally the `isPromptCanceled` helps prevent the race conditions when adding messages to the history and we don't require this to return a boolean value and just move it to the `getPreprocessedRequestHistory`

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -574,25 +575,26 @@ export class AgenticChatController implements ChatHandlers {
                 )
             }
             const remainingCharacterBudget = this.truncateRequest(currentRequestInput)
-            //  Fix the history to maintain invariants
+            //  Get and process the messages from history DB to maintain invariants for service requests
+            let messages: DbMessage[] = []
             if (currentMessage) {
-                const isHistoryValid = this.#chatHistoryDb.fixAndValidateHistory(
+                // Get preprocessed history messages from DB with size and number of character limits
+                let messages = this.#chatHistoryDb.getPreprocessedRequestHistory(
```
tsmithsz, line 582: nit: remove `let`

jngyuamz: Good catch, that's another merge conflict issue.

PR #1377, opened 2025-05-19
SSPC dependency upload and watcher fixes

wweitao:
## Problem
1. Symlink inside dependency folder wasn't resolved and didn't get uploaded
2. Dependency watcher can crash Flare when there are lots of file events happening to file watcher.

## Solution

1. Resolve symlink
2. Bundle events for file watcher

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
@@ -4,7 +4,7 @@ import path = require('path')
 import { URI } from 'vscode-uri'
 import JSZip = require('jszip')
 import { EclipseConfigGenerator, JavaProjectAnalyzer } from './javaManager'
-import { isDirectory, isEmptyDirectory } from './util'
+import { resolveSymlink, isDirectory, isEmptyDirectory } from './util'
 import glob = require('fast-glob')
 import { CodewhispererLanguage, getCodeWhispererLanguageIdFromPath } from '../../shared/languageDetection'
 
@@ -45,6 +45,15 @@ const IGNORE_PATTERNS = [
     '**/bin/**',
     // Framework specific
     '**/target/**', // Maven/Gradle builds
+]
+
+const IGNORE_DEPENDENCY_PATTERNS = [
+    // Package management and git
+    '**/.git/**',
+    // Build outputs
+    '**/dist/**',
+    // Logs and temporary files
+    '**/logs/**',
 ]
 
 interface FileSizeDetails {
@@ -161,14 +170,14 @@ export class ArtifactManager {
             const files = await glob(['**/*'], {
                 cwd: filePath,
                 dot: false,
-                ignore: IGNORE_PATTERNS,
-                followSymbolicLinks: false,
+                ignore: IGNORE_DEPENDENCY_PATTERNS,
+                followSymbolicLinks: true,
                 absolute: false,
                 onlyFiles: true,
             })
 
             for (const relativePath of files) {
-                const fullPath = path.join(filePath, relativePath)
+                const fullPath = resolveSymlink(path.join(filePath, relativePath))
                 try {
                     const fileMetadata = await this.createFileMetadata(
                         fullPath,
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
@@ -177,8 +177,8 @@ export class ArtifactManager {
             })
 
             for (const relativePath of files) {
-                const fullPath = resolveSymlink(path.join(filePath, relativePath))
                 try {
+                    const fullPath = resolveSymlink(path.join(filePath, relativePath))
                     const fileMetadata = await this.createFileMetadata(
                         fullPath,
                         path.join(filePathInZipOverride !== undefined ? filePathInZipOverride : '', relativePath),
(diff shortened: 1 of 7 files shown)


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyHandler/DependencyWatcher.ts
```diff
@@ -0,0 +1,66 @@
+import { Logging } from '@aws/language-server-runtimes/server-interface'
+import * as fs from 'fs'
+
+export class DependencyWatcher {
```
ege0zcan, line 4: Not as part of this PR but we should consider replacing the custom watchers with handlers for `didChangeWatchedFiles` event https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles

This would move the file watching responsibility out of servers and into the extensions and some LSP clients (like the one vscode has) already have everything in place to handle file watching correctly

wweitao: That's great callout. We should consider that and will evaluate with other Q dev platforms LSP support.

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyHandler/DependencyWatcher.ts
```diff
@@ -0,0 +1,66 @@
...
+        private readonly callbackFunction: (events: string[]) => void,
+        private readonly logging: Logging,
+        private readonly interval: number = 1000
+    ) {
+        this.watcher = this.setupWatcher()
+    }
+
+    private setupWatcher(): fs.FSWatcher {
+        try {
+            const watcher = fs.watch(this.path, { recursive: false }, async (eventType, fileName) => {
```
ege0zcan, line 21: Should we set a listener for watcher `error` event?

If the watcher errors out it won't be usable anymore and the error thrown by watcher will be propogated upwards 

In the case of watcher erroring out, we can either log the error and stop or we can try to recreate the watcher

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/util.ts
```diff
@@ -57,6 +57,26 @@ export const isDirectory = (path: string): boolean => {
...
+export const resolveSymlink = (dependencyPath: string): string => {
+    let truePath: string = dependencyPath
+    let symlinkTarget: string | null = null
+    try {
+        if (fs.lstatSync(dependencyPath).isSymbolicLink()) {
+            symlinkTarget = fs.readlinkSync(dependencyPath)
+
+            // If the symlink target is relative, resolve it relative to the symlink's directory
+            if (!path.isAbsolute(symlinkTarget)) {
+                symlinkTarget = path.resolve(path.dirname(dependencyPath), symlinkTarget)
```
ege0zcan, line 69: What is symlinkTarget being used for? Are we verifying some scenarios with these calls?

Otherwise the function can be simplified down to: 

```
export const resolveSymlink = (dependencyPath: string): string => {
    try {
        if (fs.lstatSync(dependencyPath).isSymbolicLink()) {
            return fs.realpathSync(dependencyPath)
        }
    } catch (error) {
    }
    return dependencyPath
}
```

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
```diff
@@ -47,6 +47,15 @@ const IGNORE_PATTERNS = [
     '**/target/**', // Maven/Gradle builds
 ]
 
+const IGNORE_DEPENDENCY_PATTERNS = [
```
ege0zcan, line 50: Isn't this list too simple? 

It should at least contain the following from the IGNORE_PATTERNS:

```
    '**/build/**',
    '**/out/**',
    '**/coverage/**',
    '**/logs/**',
    '**/tmp/**',
 ```
 
 I also suggest keeping the two ignore lists tidier by getting rid of duplicates,
 
 We can do 
   ```
const IGNORE_DEPENDENCY_PATTERNS = [ '**/.git/**', `'**/dist/**',`]
const IGNORE_PATTERNS = [...IGNORE_DEPENDENCY_PATTERNS, `'**/node_modules/**',`<the rest of the patterns> ]
```

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyHandler/DependencyWatcher.ts
wweitao: I can add log for error event.

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/util.ts
wweitao: For SymlinkTarget, it can be relative path. Let me double check if realpathSync can resolve this case.

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/util.ts
```diff
@@ -57,6 +57,26 @@ export const isDirectory = (path: string): boolean => {
...
+            }
+
+            // Get the real path (resolves all symlinks in the path)
+            truePath = fs.realpathSync(dependencyPath)
+        }
+    } catch (error) {}
```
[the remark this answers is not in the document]
wweitao: Sure. I can put one.

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
wweitao: I want to separate them. I have found these patterns can skip some valuable source files if customer defines them in a way matching the pattern. So for dependency, I would keep them as simple.

natdurlik (approved):

PR #1382, opened 2025-05-20
fix: Render timeout error, JSON parse error, cancellation to the in progress fs.write UI

leigaol:
## Problem
When there is a timeout error and/or JSON parse error, etc of the tool use and or a user cancel, the error status was not rendered in the progress fsWrite UI. that UI will remain in progress status.

## Solution
At those code path of timeout error and/or JSON parse error of the tool use and or a user cancel error, render the error in the fsWrite UI.

https://github.com/user-attachments/assets/c5958360-1712-462b-b9f1-82f728de9be1

This was tested by manually injecting errors including timeout error, JSON parse error, etc.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -410,10 +410,11 @@ export class AgenticChatController implements ChatHandlers {
                 // depends on it
                 session.conversationId = uuid()
             }
-
+            const chatResultStream = this.#getChatResultStream(params.partialResultToken)
             token.onCancellationRequested(async () => {
                 this.#log('cancellation requested')
                 await this.#showUndoAllIfRequired(chatResultStream, session)
+                await chatResultStream.updateOngoingProgressResult('Canceled')
                 await this.#getChatResultStream(params.partialResultToken).writeResultBlock({
                     type: 'directive',
                     messageId: 'stopped' + uuid(),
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -410,10 +410,11 @@ export class AgenticChatController implements ChatHandlers {
                 // depends on it
                 session.conversationId = uuid()
             }
-
+            const chatResultStream = this.#getChatResultStream(params.partialResultToken)
             token.onCancellationRequested(async () => {
                 this.#log('cancellation requested')
                 await this.#showUndoAllIfRequired(chatResultStream, session)
+                await chatResultStream.updateOngoingProgressResult('Canceled')
                 await this.#getChatResultStream(params.partialResultToken).writeResultBlock({
```
jguoamz, line 418: nit: we can reuse the chatResultStream variable now?

jguoamz (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
leigaol: the this. inside the arrow function does not bind to the same context. 

`chatResultStream` and `this.#getChatResultStream(params.partialResultToken)` is not the same result stream. I don't have context why this was implemented in the first place, but I suggest we use the same `chatResultStream`

jguoamz: okay, thanks for the explanation, but this is very confusing, we should fix it

PR #1387, opened 2025-05-20
fix(amazonq): allow init AmazonQIAMServiceManager multiple times

imykhai:
## Problem
There is a use case, where there is more than one bundle in the same memory space. In this scenario we would like to have these 2 bundles share the same instance of AmazonQIAMServiceManager

## Solution
allow init AmazonQIAMServiceManager multiple times

## License


server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQTokenServiceManager.ts
```diff
@@ -98,7 +98,7 @@ export class AmazonQTokenServiceManager extends BaseAmazonQServiceManager<
             return AmazonQTokenServiceManager.instance
         }
 
-        throw new AmazonQServiceAlreadyInitializedError()
+        return AmazonQTokenServiceManager.instance
```
suprajaven, line 101: Do we want to update it for the token based as well - wondering if there might be any possible unintended consequence?

imykhai: I wanted to keep them consistent. 
As discussed with @viktorsaws , there shouldn't be.

server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/AmazonQIAMServiceManager.ts
```diff
@@ -35,7 +35,7 @@ export class AmazonQIAMServiceManager extends BaseAmazonQServiceManager<
             return AmazonQIAMServiceManager.instance
         }
 
-        throw new AmazonQServiceAlreadyInitializedError()
+        return AmazonQIAMServiceManager.instance
```
viktorsaws, line 38: supernit: we could avoid having 2 identical returns by removing one inside of "if"

viktorsaws (approved):

PR #1394, opened 2025-05-20
fix: handle requestAborted errors silently

tsmithsz:
## Problem
If a user stops the agentic chat mid-execution while the request to Q is in-flight, we abort the request. However, we are currently surfacing that request to customers

## Solution
Swallow exception and handle it in `isUserAction`

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
@@ -7,6 +7,7 @@ type AgenticChatErrorCode =
     | 'InputTooLong' // too much context given to backend service.
     | 'PromptCharacterLimit' // customer prompt exceeds
     | 'ResponseProcessingTimeout' // response didn't finish streaming in the allowed time
+    | 'RequestAborted' // request was aborted by the user
 
 export const customerFacingErrorCodes: AgenticChatErrorCode[] = [
     'QModelResponse',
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
```diff
@@ -52,3 +53,16 @@ export function isInputTooLongError(error: unknown): boolean {
...
     return false
 }
+
+export function isRequestAbortedError(error: unknown): boolean {
+    if (error instanceof AgenticChatError && error.code === 'RequestAborted') {
+        return true
+    }
+
+    if (error instanceof Error) {
+        //  This is fragile (breaks if the backend changes their error message wording)
```
Hweinstock, line 63: appreciate the comment.

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
```diff
@@ -151,11 +151,19 @@ export class ChatSessionService {
             try {
                 return await client.generateAssistantResponse(request, this.#abortController)
             } catch (e) {
+                if (isRequestAbortedError(e)) {
+                    const requestId =
+                        e instanceof CodeWhispererStreamingServiceException ? e.$metadata?.requestId : undefined
```
Hweinstock, line 156: out of curiosity, should this always be defined?
Is it just the SDK generated types that allow for this to be missing?

Hweinstock (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/chatSessionService.ts
tsmithsz: Ideally it should always be there. I can probably remove this check in a follow-up

Hweinstock:
LGTM once CI passes.

jguoamz (approved):

ashishrp-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
```diff
@@ -52,3 +53,16 @@ export function isInputTooLongError(error: unknown): boolean {
...
+        return error.message.includes('Request aborted')
```
yueny2020, line 64: i cannot find this exception from RTS? which component does it from?

tsmithsz: The "Request aborted" error originates from the underlying fetch API used by the AWS SDK when a request is aborted.

PR #1404, opened 2025-05-21
fix: the new prompt wont stop the process properly

Randall-Jiang:
## Problem
When a new prompt was sent while the first was still processing, the telemetry emission for the cancelled prompt was happening before the request was aborted, which prevented the system from properly handling the second prompt.

## Solution
Moved the telemetry emission to occur after the request is aborted

https://github.com/user-attachments/assets/d60d4ed2-b31a-496d-8147-800bc864721b

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -427,11 +427,12 @@ export class AgenticChatController implements ChatHandlers {
                     session.pairProgrammingMode,
                     session.getConversationType()
                 )
-                await this.#telemetryController.emitAddMessageMetric(params.tabId, metric.metric, 'Cancelled')
 
                 session.abortRequest()
                 void this.#invalidateAllShellCommands(params.tabId, session)
                 session.rejectAllDeferredToolExecutions(new CancellationError('user'))
+
+                await this.#telemetryController.emitAddMessageMetric(params.tabId, metric.metric, 'Cancelled')
             })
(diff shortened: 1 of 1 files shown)


tsmithsz (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -427,11 +427,12 @@ export class AgenticChatController implements ChatHandlers {
                     session.pairProgrammingMode,
                     session.getConversationType()
                 )
-                await this.#telemetryController.emitAddMessageMetric(params.tabId, metric.metric, 'Cancelled')
 
                 session.abortRequest()
                 void this.#invalidateAllShellCommands(params.tabId, session)
                 session.rejectAllDeferredToolExecutions(new CancellationError('user'))
+
```
laileni-aws, line 434: How does this impact the streaming? due to latency of emitting the telemetry?

Randall-Jiang: I think the await statement delay the actual cancellation. And if telemetry call takes time or fails, it will block the cancellation.

tsmithsz: I think it has to do with the order of operations. Due to the latency of emitting the telemetry there is a delay before the request is aborted

laileni-aws: Got it thank you.

laileni-aws (approved):

PR #1406, opened 2025-05-21
feat(amazonq): enhanced utg

Will-ShaoHua:
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/focalFileResolution.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/focalFileResolution.ts
@@ -1,0 +1,357 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as path from 'path'
+import * as os from 'os'
+import * as fs from 'fs'
+
+import { pathUtils } from '@aws/lsp-core'
+
+type Metadata = {
+    lang: string
+    extensions: string[]
+    testAffixes: string[]
+    packageMarker?: RegExp
+    importPatterns?: RegExp[]
+    packageSeparator: string
+    moduleAliases?: Record<string, string>
+}
+
+const LANGUAGE_CONFIG: Record<string, Metadata> = {
+    java: {
+        lang: 'java',
+        extensions: ['.java'],
+        testAffixes: ['Test', 'Tests'],
+        packageMarker: /^package\s+([a-zA-Z0-9_.]+);/,
+        packageSeparator: '.',
+    },
+    python: {
+        lang: 'python',
+        extensions: ['.py'],
+        testAffixes: ['test_', '_test'],
+        importPatterns: [/^import\s+([a-zA-Z0-9_.]+)/, /^from\s+([a-zA-Z0-9_.]+)/],
+        packageSeparator: '.',
+    },
+    javascript: {
+        lang: 'javascript',
+        extensions: ['.js'],
+        testAffixes: ['.test', '.spec'],
+        importPatterns: [/^import.*from\s+[\'"](.+)[\'"]/],
+        packageSeparator: '/',
+    },
+    typescript: {
+        lang: 'typescript',
+        extensions: ['.ts'],
+        testAffixes: ['.test', '.spec'],
+        importPatterns: [/^import.*from\s+[\'"](.+)[\'"]/],
+        packageSeparator: '/',
+        moduleAliases: {
+            '@src': 'src',
+        },
+    },
+}
+
+class FocalFileResolver {
+    constructor() {}
+
+    inferSourceFile(testFilePath: string, projectRoot: string, language: string): string | undefined {
+        const inferredSrcFilename = this.inferFocalFilename(testFilePath, language)
+
+        if (!inferredSrcFilename) {
+            console.log(`not able to infer source file name`)
+            return
+        }
+
+        const config = LANGUAGE_CONFIG[language]
+
+        // find candidate focal files based on naming conventions
+        const candidates: { fullPath: string; relativePath: string }[] = []
+        const files = this.walk(projectRoot)
+
+        for (const file of files) {
+            const ext = path.extname(file)
+            const base = path.basename(file, ext)
+
+            if (base === inferredSrcFilename) {
+                // TODO: not correct, fix fullPath & relativePath once walk is implemented
+                candidates.push({
+                    fullPath: file,
+                    relativePath: file,
+                })
+            }
+        }
+
+        if (candidates.length === 0) {
+            return undefined
+        }
+
+        if (candidates.length === 1) {
+            return candidates[0].fullPath
+        }
+
+        // filter based on the imported path and symbols
+        const importedFiles = this.extractImportedPaths(testFilePath, config, projectRoot)
+        const filteredCandidate = []
+        for (const candidate of candidates) {
+            for (const importedFileAbsPath of importedFiles) {
+                const ext = path.extname(importedFileAbsPath)
+                // os.path.splittext(importedFileAbsPath)[0]
+                const p = importedFileAbsPath.substring(0, importedFileAbsPath.length - ext.length)
+                if (candidate.fullPath.startsWith(importedFileAbsPath) || candidate.fullPath.includes(p)) {
+                    filteredCandidate.push(candidate)
+                    break
+                }
+            }
+        }
+
+        if (filteredCandidate.length) {
+            switch (language) {
+                case 'javascript':
+                case 'typescript':
+                    const importedSymbols = this.extractImportedSymbols(testFilePath)
+                    let bestMatch = undefined
+                    let bestOverlap = 0
+
+                    for (const candidate of filteredCandidate) {
+                        const exportedSymbols = this.extractExportedSymbolsFromFile(candidate.fullPath)
+                        const overlap = exportedSymbols.length // TODO: intersection of imported & exported
+                        if (overlap > bestOverlap) {
+                            bestOverlap = overlap
+                            bestMatch = candidate
+                        }
+                    }
+
+                    if (bestMatch) {
+                        return bestMatch.fullPath
+                    }
+                    return undefined
+
+                default:
+                    return filteredCandidate[0].fullPath
+            }
+        }
+
+        return undefined
+    }
+
+    inferFocalFilename(testFilePath: string, language: string): string | undefined {
+        const config = LANGUAGE_CONFIG[language]
+        const filename = path.basename(testFilePath)
+        const ext = path.extname(testFilePath)
+
+        let inferredSrcFilename: string | undefined
+        for (const affix of config['testAffixes']) {
+            if (filename.endsWith(affix)) {
+                inferredSrcFilename = filename.substring(0, filename.length - affix.length) + ext
+                break
+            } else if (filename.endsWith(affix)) {
+                inferredSrcFilename = filename.substring(affix.length) + ext
+                break
+            }
+        }
+
+        return inferredSrcFilename
+    }
+
+    extractImportedPaths(testFilePath: string, config: Metadata, projectRoot: string): string[] {
+        const content = fs.readFileSync(testFilePath)
+        const lines = content.toString().split(os.EOL)
+        // TODO: original science source code use set
+        const result: string[] = []
+        let buffer = ''
+        let insideImportBlock = false
+        try {
+            for (const l of lines) {
+                const line = l.trim()
+
+                if (config.lang === 'java') {
+                    const match = config.packageMarker?.exec(line)
+                    if (match) {
+                        // TODO: 0 or 1 ?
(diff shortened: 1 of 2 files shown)


Will-ShaoHua:
`unitTestIntentDetection` is complete
`focalFileIdentity` is WIP

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/unitTestIntentDetection.ts
```diff
@@ -0,0 +1,149 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { TextDocument } from '@aws/language-server-runtimes/server-interface'
+
+const testFileNameRegex: Record<string, RegExp[]> = {
+    python: [/^test_.*\.py$/, /.*_test\.py$/],
+    java: [/^test.*\.java$/, /.*Test\.java$/],
```
molingbo1220, line 10: For the Java regex, it seems to be missing 'Tests' as a valid affix. Also, the first 'test' in the pattern should have a capital 'T'.

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/unitTestIntentDetection.ts
```diff
@@ -0,0 +1,149 @@
...
+                '\\s*\\{',
+            'gm'
+        )
+
+        return this.curlyBracesSyntaxUtil(signaturePattern, content)
+    }
+
+    // @VisibleForTesting
+    jsTsTestIntent(content: string): boolean {
+        const signaturePattern = new RegExp(/(it|test)\s*\(\s*["\'].*?["\']\s*,\s*(async\s+)?\(\s*\)\s*=>\s*\{/, 'gm')
```
molingbo1220, line 80: Should we pass the pattern as a string, similar to how it's done in functions like javaTestIntent and pyTestIntent?

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/unitTestIntentDetection.test.ts
```diff
@@ -0,0 +1,272 @@
...
+
+        describe('should return false if filename follows test naming convertion BUT content doesnt contain unit test keywords', function () {
+            const testCases: { filepath: string; language: string; content: string }[] = [
+                {
+                    filepath: 'fooTest.java',
+                    language: 'java',
+                    content: ``,
+                },
+                {
+                    filepath: 'test_foo_class.py',
```
molingbo1220, line 77: repeated test cases in Line 77-84

Will-ShaoHua: fixed

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/focalFileResolution.ts
```diff
@@ -0,0 +1,439 @@
...
+    importPatterns?: RegExp[]
+    packageSeparator: string
+    moduleAliases?: Record<string, string>
+}
+
+const LANGUAGE_CONFIG: Record<string, Metadata> = {
+    java: {
+        lang: 'java',
+        extensions: ['.java'],
+        testAffixes: ['Test', 'Tests'],
```
molingbo1220, line 28: Change the order of the affixes into ['Tests', 'Test'], so that the longer affix 'Tests' will be checked first. This should solve the failed test case about "TestsFoo.java"

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/focalFileResolution.test.ts
```diff
@@ -0,0 +1,292 @@
...
+        fs.rmSync(tmpProjectRoot, { recursive: true, force: true })
+    })
+
+    describe('inferFocalFilename', function () {
+        describe('java', function () {
+            const testCases = [
+                'FooTest.java',
+                'FooTests.java',
+                'TestFoo.java',
+                // 'TestsFoo.java' // TODO: current implementation this case will fail
```
molingbo1220, line 31: Check the comment on Line 28 in focalFileResolution.ts

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/focalFileResolution.ts
```diff
@@ -0,0 +1,439 @@
...
+        }
+
+        // If there are multiple matches of source files, filter based on the imported path and symbols
+        const importedFiles = this.extractImportedPaths(testFilePath, language, projectRoot)
+        const filteredCandidate = []
+        for (const candidate of candidates) {
+            for (const importedFileAbsPath of importedFiles) {
+                const ext = path.extname(importedFileAbsPath)
+                // os.path.splittext(importedFileAbsPath)[0]
+                const p = importedFileAbsPath.substring(0, importedFileAbsPath.length - ext.length)
```
molingbo1220, line 129: In Java, the importedFileAbsPath may be like "com/amazon/q/service" without file extension. In such cases, minus ext.length could break the obtained path.

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/focalFileResolution.ts
```diff
@@ -0,0 +1,439 @@
...
+        try {
+            for (const l of lines) {
+                const line = l.trim()
+
+                if (config.lang === 'java') {
+                    const match = config.packageMarker?.exec(line)
+                    if (match) {
+                        const pkg = this.resolvePackageToPath(match[1], config.packageSeparator)
+                        result.add(pkg)
+                        continue
```
molingbo1220, line 205: Maybe place "continue" outside the if (match) block? That way, if the current line is not matched, it will still directly go to check the next line without going through the remaining logic. Similar to the "continue" on Line 215

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/unitTestIntentDetection.ts
```diff
@@ -0,0 +1,136 @@
...
+
+const testFileNameRegex: Record<string, RegExp[]> = {
+    python: [/^test_.*\.py$/, /.*_test\.py$/],
+    java: [/^Tests?.*\.java$/, /.*Tests?\.java$/],
+    typescript: [/.*\.(test|spec)\.ts$/],
+    javascript: [/.*\.(test|spec)\.js$/],
+}
+
+const testKeywordsRegex: Record<string, RegExp[]> = {
+    python: [/^import unittest/m, /^from unittest/m, /^def test_/m],
```
molingbo1220, line 16: Tried several test cases in Python, but the test intent was not triggered, even when the file contained the expected keyword patterns.

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/unitTestIntentDetection.ts
```diff
@@ -0,0 +1,136 @@
...
+import { TextDocument } from '@aws/language-server-runtimes/server-interface'
+
+const testFileNameRegex: Record<string, RegExp[]> = {
+    python: [/^test_.*\.py$/, /.*_test\.py$/],
+    java: [/^Tests?.*\.java$/, /.*Tests?\.java$/],
+    typescript: [/.*\.(test|spec)\.ts$/],
+    javascript: [/.*\.(test|spec)\.js$/],
+}
+
+const testKeywordsRegex: Record<string, RegExp[]> = {
```
molingbo1220, line 15: Keyword regex could be expanded a bit for Python and Java, 

Python: `[r"^import\s+pytest", r"^from\s+pytest"]`

Java: `[r"import\s+org\.junit\.", r"import\s+org\.testng\.", r"import\s+org\.mockito\.", r"^import\s+org\.assertj\.", r"^import\s+org\.springframework\.test\."]`

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/focalFileResolution.ts
Will-ShaoHua: yes you are right, fixed

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/supplementalContextUtil.ts
```diff
@@ -27,18 +33,40 @@ export async function fetchSupplementalContext(
...
         let supplementalContextValue:
             | Pick<CodeWhispererSupplementalContext, 'supplementalContextItems' | 'strategy'>
             | undefined
 
         if (isUtg) {
-            return
+            supplementalContextValue = await waitUntil(
+                async function () {
+                    const focalFile = await utgFocalFileResolver.inferFocalFile(document, workspace)
+                    if (focalFile) {
```
molingbo1220, line 47: When `isUtg` is true but `inferFocalFile` fails to identify a focal file, should we fall back to using BM25 to retrieve supplemental context? Currently, the logic seems to return an empty result in this case.

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/focalFileResolution.ts
Will-ShaoHua: yes thanks for capturing this, i missed this. The original src code is as. you mentioned

Will-ShaoHua: https://github.com/aws/language-servers/pull/1406/commits/d872700aa4d98e5bab9f4b744143ca18b91b1004

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/unitTestIntentDetection.ts
Will-ShaoHua: should be fixed in https://github.com/aws/language-servers/pull/1406/commits/d25cbfc16c6ae774251983d5ae232a1d7e148917

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/unitTestIntentDetection.ts
Will-ShaoHua: addressed by
https://github.com/aws/language-servers/pull/1406/commits/2cd3b2fb5300cba2d0350862509ab69673d61431

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/supplementalContextUtil.ts
Will-ShaoHua: will follow up

Will-ShaoHua:
```
 765 passing (6s)
  8 pending
  1 failing

  1) focalFileResolver
       extractImportedPaths
         python
           case1:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

1 !== 5

      + expected - actual

      -1
      +5
      
      at Context.<anonymous> (src\shared\supplementalContextUtil\focalFileResolution.test.ts:120:24)
      at processImmediate (node:internal/timers:476:21)
```

Will-ShaoHua:
also not work in windows
```
   1 failing

  1) focalFileResolver
       extractImportedPaths
         ts
           case1:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

1 !== 3

      + expected - actual

      -1
      +3
      
      at Context.<anonymous> (src\shared\supplementalContextUtil\focalFileResolution.test.ts:147:24)
      at processImmediate (node:internal/timers:476:21)
```

server/aws-lsp-codewhisperer/src/shared/supplementalContextUtil/unitTestIntentDetection.ts
```diff
@@ -0,0 +1,146 @@
...
+
+        if (matches.length === 0) {
+            return false
+        }
+
+        // Get content after the last test
+        const lastMatch = matches[matches.length - 1]
+        const lastMatchPos = lastMatch.index! + lastMatch[0].length
+        const tailFromLastTest = content.slice(lastMatchPos)
+        const lines = tailFromLastTest.split('\n')
```
Will-ShaoHua, line 100: should use either `\r?\n` or `os.EOL`

Will-ShaoHua: nm these are not used at this point. Will consider removing them when we're sure they are 100% not needed anymore

Will-ShaoHua:
followup AIs
- https://github.com/aws/language-servers/pull/1406#discussion_r2121638757 fallback to use bm25 when utg fail
- add logging

molingbo1220 (approved):

leigaol (approved):

PR #1411, opened 2025-05-21
feat: support persona file for MCP permission

bywang56:
## Problem
We need to support persona file and move all permission settings into it.
https://quip-amazon.com/B3Z9ADFBZxHi/Agent-personas-in-Q-Developer-CLI-and-IDE

## Solution
Support persona file

## TODOs
* Fix remaining issues from front end integration.
* There is also requirement changes that not been addressed yet like the default content in persona file, which needs to be addressed.
* Manual test all permission use cases
* Rewrite broken uni tests

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
@@ -7,8 +7,9 @@ import {
     ListMcpServersParams,
     McpServerClickParams,
 } from '@aws/language-server-runtimes/protocol'
-import { getGlobalMcpConfigPath } from './mcpUtils'
-import { MCPServerConfig } from './mcpTypes'
+
+import { getGlobalMcpConfigPath, getGlobalPersonaConfigPath, getWorkspacePersonaConfigPaths } from './mcpUtils'
+import { MCPServerConfig, MCPServerPermission } from './mcpTypes'
 
 interface PermissionOption {
     label: string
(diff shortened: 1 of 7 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
```diff
@@ -287,15 +290,19 @@ export class McpEventHandler {
...
-            disabled: false,
         }
 
+        // TODO: handle ws/global selection
         let configPath = ''
         if (params.optionsValues.scope === 'global') {
             configPath = getGlobalMcpConfigPath(this.#features.workspace.fs.getUserHomeDir())
         }
+        let personaPath = ''
+        if (params.optionsValues.scope === 'global') {
```
laileni-aws, line 301: Thanks for adding this

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
```diff
@@ -436,24 +447,24 @@ export class McpEventHandler {
     #buildPermissionOptions(currentPermission: string) {
         const permissionOptions: PermissionOption[] = []
 
-        if (currentPermission !== 'Always run') {
+        if (currentPermission !== 'alwaysAllow') {
```
laileni-aws, line 450: nit: We can create a enum for permission and use this here for permission states.

but this is not blocker.

laileni-aws (approved):
LGTM

PR #1419, opened 2025-05-22
fix: adding chat summary card for mcp tools

laileni-aws:
## Problem
- Missing UX for MCP chat summary card view and permission ux.
## Solution

https://github.com/user-attachments/assets/e540cf00-f210-4958-b2c9-780911becd05

## License


chat-client/src/client/mynahUi.ts
```diff
@@ -900,6 +900,49 @@ export const createMynahUi = (
 
         let processedHeader = header
         if (message.type === 'tool') {
+            // Handle MCP tool summary with accordion view
+            if (message.summary) {
```
bywang56, line 904: Is this only for MCP? Will build-in tool also has summary?

bywang56 (commented):
will this PR fix this issue for runtime version upgrade?

chat-client/src/client/mynahUi.ts
ashishrp-aws: No only mcp tool messages

ashishrp-aws (approved):

PR #1440, opened 2025-05-24
fix: properly tokenize command args using shlex.split()

tsmithsz:
## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -458,12 +458,11 @@ export class ExecuteBash {
...
             }
 
-            this.childProcess = new ChildProcess(
-                this.logging,
-                shellName,
-                [shellFlag, params.command],
-                childProcessOptions
-            )
+            const shellArgs = IS_WINDOWS_PLATFORM
+                ? [shellFlag, ...split(params.command)] // Windows: split for proper arg handling
```
awschristou, line 462: Should we have some tests that assert given certain inputs that the args are being split as expected before being sent to ChildProcess?

tsmithsz: Added

jguoamz (approved):

PR #1454, opened 2025-05-28
fix: adding validation for list of mcp servers

laileni-aws:
## Problem
- No validation in list of mcp servers
## Solution
- Added validation.

https://github.com/user-attachments/assets/cb6a6f5e-7d90-47f9-bfc6-4fb08037d1cb

## TODO:
- Edit server name should be disabled and need to provide option to delete the server.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
@@ -72,37 +72,8 @@ export class McpEventHandler {
         }
         const mcpManagerServerConfigs = mcpManager.getAllServerConfigs()
 
-        // Validate each server configuration and collect all validation errors
-        const validationErrors: { serverName: string; errors: string[] }[] = []
-
-        for (const [serverName, config] of mcpManagerServerConfigs.entries()) {
-            // Create a values object that matches the expected format for validateMcpServerForm
-            const values = {
-                name: serverName,
-                command: config.command,
-                timeout: config.timeout?.toString() || '',
-            }
-
-            const validation = this.#validateMcpServerForm(values)
-            if (!validation.isValid) {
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
```diff
@@ -68,6 +70,38 @@ export class McpEventHandler {
         }
         const mcpManagerServerConfigs = mcpManager.getAllServerConfigs()
 
+        // Validate each server configuration and collect all validation errors
+        const validationErrors: { serverName: string; errors: string[] }[] = []
+
+        for (const [serverName, config] of mcpManagerServerConfigs.entries()) {
+            // Create a values object that matches the expected format for validateMcpServerForm
+            const values = {
```
bywang56, line 78: I think we should include all attributes from the forms here, so that if we add additional validations—like the one Jacob just added for env vars—we won’t need to modify this logic each time we update validateMcpServerForm.

bywang56 (commented):
At a high level, could we move the error checking and display logic into a separate helper function? I think we could reuse it for other errors, like JSON syntax errors. Also, have you tried reading mcpServerStates from McpManager to handle syntax errors?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
laileni-aws: I can pass additional values but right now `validateMcpServerForm` function does not have any logic to validate other fields.

laileni-aws:
> At a high level, could we move the error checking and display logic into a separate helper function? I think we could reuse it for other errors, like JSON syntax errors.

I can move this to a separate function

>have you tried reading mcpServerStates from McpManager to handle syntax errors?

No, thats my next task to try on.

PR #1456, opened 2025-05-28
fix: decode UTF-16LE shell output on Windows

tsmithsz:
## Problem

On Windows we launch shell commands through `cmd.exe /c`. `cmd.exe `writes legacy code‑page bytes.
Because our `ExecuteBash` tool always assumed `UTF‑8`, any non‑ASCII characters in the output were being mis‑decoded and displayed as `??`. This affected customers

## Solution
- Invoke cmd with /u on Windows: Forces `UTF‑16 LE` encoding (this is how Windows handles Unicode) for all output written to the process pipes, independent of the console code‑page.
- Heuristic decoder: 
	•	Re‑creates the raw bytes from the received string `(Buffer.from(text, 'binary'))`.
	•	Detects `UTF‑16 LE` by checking whether every odd byte in the first 32 bytes is `0x00`.
	•	Decodes with `buffer.toString('utf16le')` when the pattern matches, otherwise falls back to `UTF‑8`.

### Before Fix:
Screenshot 2025-05-28 at 2 17 34 AM

### After Fix:

Screenshot 2025-05-28 at 2 15 53 AM

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -536,6 +536,19 @@ export class ExecuteBash {
         })
     }
 
+    private static decodeWinUtf(raw: string): string {
+        const buffer = Buffer.from(raw, 'binary')
+
+        let utf16 = true
+        for (let i = 1, n = Math.min(buffer.length, 32); i < n; i += 2) {
```
awschristou, line 543: I don't know how to detect if a string is utf8 or utf16, but is https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/isWellFormed valid for this purpose?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -536,6 +536,19 @@ export class ExecuteBash {
         })
     }
 
+    private static decodeWinUtf(raw: string): string {
```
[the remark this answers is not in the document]
tsmithsz: Added

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
tsmithsz: I do not believe `isWellFormed()` will work here. It only validates UTF-16 strings already in JS memory, but our issue is earlier: determining if incoming bytes are UTF-16LE or UTF-8. By the time Node decodes these bytes into a string, they'll be "well-formed" either way.

bryceitoc9: If `/u` forces UTF-16LE, under what circumstances would we have to fall back onto UTF-8?

tsmithsz: It forces it for built-in commands, but if the user is running an external program, it's free to use whatever encoding it wants

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -536,6 +536,24 @@ export class ExecuteBash {
         })
     }
 
+    /**
+     * Re‑creates the raw bytes from the received string (Buffer.from(text, 'binary')).
+     * Detects UTF‑16 LE by checking whether every odd byte in the first 32 bytes is 0x00.
+     * Decodes with buf.toString('utf16le') when the pattern matches, otherwise falls back to UTF‑8.
```
rli, line 542: is this actually a reasonable assumption? isnt the encoding defined by the user's shell locale?

tsmithsz: `cmd.exe /u` forces built‑ins to emit `UTF‑16LE`, bypassing whatever code page the user’s console is set to.

I'm not sure if true “universal” detection is practical. This heuristic covers the dominant cases (`UTF‑16LE` and `UTF‑8`).

PR #1458, opened 2025-05-28
Fix: Convert array values to comma-separated strings in telemetry metrics

XiaoxuanLu:
## Problem
Flare is sending the telemetry payloads as JSON blobs, but each IDE is re-serializing this data differently. When array values like toolName, toolUseId, and latency are passed directly to the telemetry service, inconsistencies occur across different clients. The telemetry service expects literal strings, not arrays, which causes data processing issues downstream.

## Solution
Modified the emitAgencticLoop_InvokeLLM method to flatten array values into comma-separated strings before emitting the telemetry metric. This ensures that each IDE receives properly formatted string values instead of arrays. 


server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
```diff
@@ -187,11 +187,11 @@ export class ChatTelemetryController {
                 [CONVERSATION_ID_METRIC_KEY]: conversationId,
                 cwsprChatConversationType: conversationType,
                 credentialStartUrl: this.#credentialsProvider.getConnectionMetadata()?.sso?.startUrl,
-                cwsprToolName: toolName,
-                cwsprToolUseId: toolUseId,
+                cwsprToolName: toolName?.join(','),
```
leigaol, line 190: nit: rename toolName to toolNames

awschristou (approved):

leigaol (approved):

server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
leigaol: No it won't I mean rename the toolName variable not the cwsprToolName

PR #1462, opened 2025-05-28
fix: update executeBash UI for failures during command existence check

tsmithsz:
## Problem
UI for stop button does not go away when commands fail the  `whichCommand()` check for confirming their existence on the user's system.

## Solution

Update the `executeBash` UI to show `Completed` for non-user related failures. 

##Testing

### Before fix:

image

### After fix:

https://github.com/user-attachments/assets/04a15638-0f95-4c35-98e8-b01442fbd13c

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -1167,6 +1167,36 @@ export class AgenticChatController implements ChatHandlers {
                     } else {
                         await chatResultStream.writeResultBlock(errorResult)
                     }
+                } else if (toolUse.name === 'executeBash' && toolUse.toolUseId) {
+                    const existingCard = chatResultStream.getMessageBlockId(toolUse.toolUseId)
+                    const command = (toolUse.input as unknown as ExecuteBashParams).command
+                    const completedErrorResult = {
+                        type: 'tool',
+                        messageId: toolUse.toolUseId,
+                        body: `\`\`\`shell\n${command}\n\`\`\``,
+                        header: {
+                            body: 'shell',
+                            status: {
+                                status: 'success',
+                                icon: 'ok',
+                                text: 'Completed',
+                            },
+                            buttons: [],
+                        },
+                    } as ChatResult
+
+                    if (existingCard) {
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1167,6 +1167,36 @@ export class AgenticChatController implements ChatHandlers {
...
+                    const completedErrorResult = {
+                        type: 'tool',
+                        messageId: toolUse.toolUseId,
+                        body: `\`\`\`shell\n${command}\n\`\`\``,
+                        header: {
+                            body: 'shell',
+                            status: {
+                                status: 'success',
+                                icon: 'ok',
+                                text: 'Completed',
```
awschristou, line 1182: Is success the right status to surface? It seems like the task completes, but it seems like this part of the code represents some sort of error or failure.

laileni-aws: According to figma, Irrespective of result it may be success or failure, IDE shows `Completed` as text and status as `success`.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1167,6 +1167,36 @@ export class AgenticChatController implements ChatHandlers {
                     } else {
                         await chatResultStream.writeResultBlock(errorResult)
                     }
+                } else if (toolUse.name === 'executeBash' && toolUse.toolUseId) {
+                    const existingCard = chatResultStream.getMessageBlockId(toolUse.toolUseId)
+                    const command = (toolUse.input as unknown as ExecuteBashParams).command
```
laileni-aws, line 1172: Just curious, Can we write this as 
```
const command = (toolUse.input as ExecuteBashParams).command
```

If yes please include it!

tsmithsz: Yeah this raises errors since `toolUse.input` is not guaranteed to be of type `ExecuteBashParams`. Need to use as unknown as ExecuteBashParams

PR #1470, opened 2025-05-29
fix line number display in code context commands

XiaoxuanLu:
## Problem
Line numbers displayed in the webview were off by 1 when showing code symbol ranges. This is because the editor's internal line numbering starts at 0, but UI displays typically show line numbers starting at 1. This inconsistency affects all IDEs and creates confusion for users when they see line numbers that don't match what's shown in their editor. 
VS:
image
VSC:
image
JB:
image

## Solution
Added +1 to both the start and end line numbers when displaying the line range in the code context command description. This ensures that the line numbers shown in the UI match what users see in their editor interface, providing a consistent experience across all IDEs.

## License


server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
awschristou, line 1: This change is from a different PR

awschristou (approved):

PR #1472, opened 2025-05-29
fix: show invalid configuration error when fixing server config

chungjac:
## Problem
- When this line errors `await Promise.race([connectPromise, timeoutPromise]) `, we know that the server configuration is not valid so the server int failed. We do not show this error to the user right now

## Solution
- Show invalid configuration error to user when user clicks on fix configuration

https://github.com/user-attachments/assets/3491002a-9d0d-4c48-ba37-79d9fdc41e40

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpManager.ts
```diff
@@ -165,7 +165,17 @@ export class McpManager {
                     )
                     timer.unref()
                 })
-                await Promise.race([connectPromise, timeoutPromise])
+                try {
+                    await Promise.race([connectPromise, timeoutPromise])
```
bywang56, line 169: I think what we should do is not catch on the race, instead just try catch client.connect(transport), then add our own message, also, we should add the actual error message in the error. 

something like this 

```
const connectPromise = client.connect(transport).catch(err => {
    throw new AgenticChatError(
        `MCP: server '${serverName}' failed to connect: ${err.message}`,
        'MCPServerConnectionFailed'
    )
})
```

this way we also don't swallow the init timeout error.

bywang56 (approved):

PR #1479, opened 2025-05-30
fix: use updated version of vecLib and use local context controller to raise context command updates

shruti0085:
## Problem
DidWorkspaceFolderChange events did not update context commands. This issue is particularly pronounced in Visual Studio where this event is used to send information about projects contained in a solution on intial load which should be indexed.
In addition, the updateIndex calls for vecLib was gated by the indexingEnabled flag which meant when a change event came with the indexer off, the context commands were not updated. The context command updates should happen unconditionally even when indexer is disabled.
## Solution
This change addresses the issues above:
* A new version of the vecLib has been added to the local development copy with the tgz file. A corresponding release PR is being tracked in a separate repo.
* The update above adds support for updating indexer and context commands when workspace folders change.
* A corresponding change has been made to the language server to support sending the workspace folders updates to the vecLib update index call.
* The context command provider and controller were very strongly coupled causing issues updating the two independently. With this change, the controller now becomes the sole authority that raises an event to context command provider to get and push latest context commands after indexing finishes.
* IndexingEnabled flag has been removed as check from the updateIndex calls. It is now expected that the vecLib knows when to not index locally(for @workspace) when the buildIndex('default') is invoked with this value. 

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
```diff
@@ -24,33 +18,13 @@ export class ContextCommandsProvider implements Disposable {
...
-        })
-    }
 
-    async updateContextCommandsOnFileChange(filePaths: string[], isAdd: boolean) {
-        try {
-            const localProjectContextController = await LocalProjectContextController.getInstance()
-            const shouldUpdate = await localProjectContextController.shouldUpdateContextCommand(filePaths, isAdd)
-            if (shouldUpdate) {
-                const contextItems = await localProjectContextController.getContextCommandItems()
+        void (async () => {
```
leigaol, line 22: can you keep the try catch so that there won't be uncaught exceptions from promoises

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -138,11 +140,13 @@ export class LocalProjectContextController {
 
             // build index if vecLib was initialized but indexing was not enabled before
             if (this._vecLib) {
+                // if indexing is turned being on, build index with 'all' that supports vector indexing
                 if (enableIndexing && !this._isIndexingEnabled) {
```
leigaol, line 144: why are 2 booleans required

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -227,19 +229,30 @@ export class LocalProjectContextController {
...
+            const actualAdditions = added.filter(
+                addition => !this.workspaceFolders.some(existing => this.areWorkspaceFoldersEqual(existing, addition))
             )
 
-            const merged = [...afterRemovals]
-            for (const addition of added) {
-                if (!merged.some(existing => this.areWorkspaceFoldersEqual(existing, addition))) {
-                    merged.push(addition)
+            this.workspaceFolders.push(...actualAdditions)
+            // Only update index if we have actual changes and indexing is enabled
```
leigaol, line 245: and indexing is enabled => and indexing library is present.

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -154,7 +158,9 @@ export class LocalProjectContextController {
             if (vecLib) {
                 this._vecLib = await vecLib.start(LIBRARY_DIR, this.clientName, this.indexCacheDirPath)
                 if (enableIndexing) {
-                    void this.buildIndex()
+                    void this.buildIndex('all')
```
leigaol, line 161: the errors inside this void buildIndex() will not be caught by the try catch.

You can do buildIndex().catch((e) => {} )

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
shruti0085: this is to gate unnecessary update. It checks if the old value has changed and only then makes the buildIndex calls accordingly

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
leigaol: plus it makes sense to await here.. why void

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
awschristou: Please log the error though

awschristou (approved):

PR #1482, opened 2025-05-30
fix: remove unused util

Randall-Jiang:
## Problem
code removed in https://github.com/aws/language-servers/pull/1473 was accidentally re-added in https://github.com/aws/language-servers/pull/1455. This pr is to move those code
## License


awschristou (commented):
Looks like this change is fixing the situation where code removed in #1473 was accidentally re-added in #1455. Is this correct?

If so, please update the PR description with these details.

PR #1483, opened 2025-05-30
fix(amazonq): normalize line endings to match local file endings in strReplace

LokeshDogga13:
## Problem

image

When using Amazon Q on Windows systems, the service responses contain LF (`\n`) line endings, but local files use CRLF (`\r\n`). This mismatch causes string comparison failures in fsWrite's strReplace command because:

1. Service returns: `oldStr` with `\n` endings
2. Local file has: `\r\n` endings
3. Direct string comparison fails even when content is semantically same
4. Results in "No occurrences found" errors requiring chat retries

## Solution

Added line ending normalization in getStrReplaceContent to handle the mismatch:

1. Detect line ending style from the target file
2. Normalize both oldStr and newStr to match file's line ending style before comparison
3. Ensures consistent comparison regardless of service response line endings

## Testing

1. Added unit tests for line ending scenarios
2. Verified Amazon Q responses now work correctly on Windows without requiring retries

## License


awschristou (approved):
Great job! We may want to change this so that the normalization attempt only happens if the original getStrReplaceContent attempt fails. We can always revisit this though.

PR #1493, opened 2025-05-30
fix: clear the set to make sure truncate function work as expected

Randall-Jiang:
## Problem
When user ask agentic what tools it has, if users ask multiple times, the tool name agentic return would with numbers. 
image

## Solution
Add a clear function to reset the Set so every time we truncate the same server_tool, it would be the same name.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpUtils.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpUtils.ts
@@ -305,12 +305,6 @@ export function enabledMCP(params: InitializeParams | undefined): boolean {
 
 export const MAX_TOOL_NAME_LENGTH = 64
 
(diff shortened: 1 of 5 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpUtils.ts
```diff
@@ -290,6 +290,14 @@ export function getOriginalToolNames(namespacedName: string): { serverName: stri
     return toolNameMapping.get(namespacedName)
 }
 
+/**
+ * Clears the tool name mapping. This should be called when starting fresh tool registration
+ * to avoid conflicts from previous registrations.
+ */
+export function clearToolNameMapping(): void {
+    toolNameMapping.clear()
```
bywang56, line 298: if MCP is not enabled, isn't this will cause type error? we should make sure we consider feature flag for any changes in the files outside of mcp folder.

PR #1497, opened 2025-05-31
fix(amazonq): pagination request should also used truncated left/right context

andrewyuq:
In the first request in a session, below, it truncates left/right context, but didn't store them in the session,
when sending the second request it should use the same truncation logic.
## License


LiGaCu (approved):
Do we need to add a unit test for this?

PR #1510, opened 2025-06-03
feat: bundle nupkg files into artifact.zip

cslong:
## Problem
* Need to bundle nupkg files into transformation request if provided

## Solution
* From the IDE pass file paths for each nupkg file in the PackageReferenceMetadata. In the artifact manager copy these files into the artifact folder and put relative path in the requirement.json

## License


server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -144,6 +144,25 @@ export class ArtifactManager {
         }
         this.logging.log('Total project references: ' + projects.length)
 
+        let packages: string[] = []
+        if (request.PackageReferences != null) {
+            for (const pkg of request.PackageReferences) {
+                if (pkg.NetCompatiblePackageFilePath == null) {
+                    continue
+                }
```
jonlouie, line 152: Do we want to guard against empty strings as well?

huawenm: also NetCompatiblePackageFilePath field may not exist, depend on toolkits or lsp released first, better to check the case

huawenm (approved):

Issue #797, opened 2025-02-21
SSO stores token file in random location


justinmk3:
# Problem

SSO auth implementation stores the cache token to whatever directory it happens to be installed to:

https://github.com/aws/language-servers/blob/17a6f6f864125c8cb2c5c936e66bf69e1eebf2ee/server/device-sso-auth-lsp/src/language-server/sso/builderId.ts#L15

# Expected behavior

- Consumer application (e.g. IDE extensions) can control the storage location.
- Default behavior should store token in the (quasi) standard `~/.aws/sso/...` location.
- Ideally, it should also try to re-use token already found in that location which may have been set by other tools.

sadderchris:
I use [nix](https://nixos.org/download/), which installs all of its artifacts to an _immutable_ store directory.  This assumes that the directory it's loaded from is always writable however, which obviously won't be the case here.  Can this use the user's configured `XDG_CACHE_HOME` for this purpose perhaps, since it appears to be writing something to a cache in this case?

justinmk3:
@sadderchris the default will be `~/.aws/sso/cache` does that work for you? Though, the lsp should plan to support XDG basedirs eventually...

sadderchris:
I think this will be fine, as long as it doesn't try to write to a directory alongside where the actual library is stored.  XDG basedir support would be even better, but I can live without it for now.

PR #1517, opened 2025-06-04
feat: added configurable token cache location

viktorsaws:
## Problem

Address an issued in Device SSO auth LSP package to make token cache location writable.

fix https://github.com/aws/language-servers/issues/797

## Solution

- Add support for configuring token cache location through LSP initialization options
- Move default token cache from local .cache directory to $HOMEDIR/.aws/device-sso-lsp/cache
- Update README with configuration documentation
- Pass token cache location through the authentication flow components

## License

### DIFF
--- a/server/device-sso-auth-lsp/src/language-server/sso/builderId.ts
+++ b/server/device-sso-auth-lsp/src/language-server/sso/builderId.ts
@@ -14,7 +14,7 @@ import * as fs from 'fs'
 import * as path from 'path'
 import * as os from 'os'
 
-export const DEFAULT_TOKEN_CACHE_DIR = path.join(os.homedir(), '.aws/device-sso-lsp/cache')
+export const DEFAULT_TOKEN_CACHE_DIR = path.join(os.homedir(), '.aws/sso/cache')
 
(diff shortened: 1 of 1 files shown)


server/device-sso-auth-lsp/src/language-server/sso/builderId.ts
```diff
@@ -12,7 +12,9 @@ import { SsoProfile as BaseSsoProfile, ClientRegistration, SsoToken, isExpired }
 
 import * as fs from 'fs'
 import * as path from 'path'
-const TOKEN_CACHE_DIR = path.join(__dirname, '.cache')
+import * as os from 'os'
+
+export const DEFAULT_TOKEN_CACHE_DIR = path.join(os.homedir(), '.aws/device-sso-lsp/cache')
```
justinmk3, line 17: why not `~/.aws/sso/cache/` ? that is the standard location. we can name the file something specific to avoid conflict with other apps

viktorsaws: right, good call, I'll change that

justinmk3 (approved):

PR #1523, opened 2025-06-04
fix: return QModelResponse as a response not an error

XiaoxuanLu:
## Problem
1. Inconsistent Error Handling Across IDEs 
Currently, Flare throws _QModelResponse_ errors directly, requiring each IDE to catch and handle them separately, which leads to inconsistent UI experiences across different IDEs compared to VS Code.
image/

image/
image/

2. Model Selection Error Code Issues
_QModelResponse_ errors don't send a error code back to IDE, as we only receive a body with errorMessageId inside instead of standardized error codes, which makes us hard to have a generalized way to handle logic for QModelResponse Error. 

## Solution
VS Code as a source of truth tries to show the error message from QModelResponse error directly to chat UI. We've modified the error handling approach to return a result instead of throwing exceptions for QModelResponse errors. This ensures that errors are displayed consistently in the chat UI across all IDE integrations and reduces the implementation burden on individual IDE teams. 

This change serves as a quick fix for the current UI inconsistency and reduces the implementation load on the IDE side. **A more comprehensive standardized error code system will be implemented when more bandwidth is available, but this interim solution addresses the most pressing issues with error handling across our IDE integrations.**

## Current Status
VS Code and Visual Studio implementations have been tested and confirmed working, and we're waiting for test from JetBrains and Eclipse teams before merging.
image
image

  

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -1011,9 +1011,8 @@ describe('AgenticChatController', () => {
             )
 
             // These checks will fail if a response error is returned.
-            const typedChatResult = chatResult as ResponseError<ChatResult>
-            assert.strictEqual(typedChatResult.message, errorMsg)
-            assert.strictEqual(typedChatResult.data?.body, errorMsg)
+            const typedChatResult = chatResult as ChatResult
+            assert.strictEqual(typedChatResult.body, errorMsg)
         })
 
         it('truncate input to 500k character ', async function () {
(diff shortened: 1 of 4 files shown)


ctlai95 (approved):

zuoyaofu:
Thanks for the PR! Do you mind providing a screenshot of what the error now looks like after the change in VS/Eclipse in the description? We can then cross verify that it still works in VSC.

XiaoxuanLu:
> Thanks for the PR! Do you mind providing a screenshot of what the error now looks like after the change in VS/Eclipse in the description? We can then cross verify that it still works in VSC/other IDEs

I tested in VS and VSC right now, wait from JB and Eclipse confirmation. I will attach what we have in the description

PR #1530, opened 2025-06-05
fix: mcp server should always load persona from workspace first

Randall-Jiang:
## Problem
If a project persona is present, we seem to be reading from it, but writing to the global persona. This means that if a project persona is present, any changes you make in the IDE are not persisted.
## Solution
We are trying to use the persona from workspace right now? If we can not find one, we use global's.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpUtils.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpUtils.ts
@@ -198,14 +198,15 @@ export async function loadPersonaPermissions(
         .filter(p => p !== globalPath)
 
     const globalExists = await workspace.fs.exists(globalPath).catch(() => false)
-
-    // global first, then workspace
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpUtils.ts
```diff
@@ -198,49 +198,43 @@ export async function loadPersonaPermissions(
...
                 logging.error(`Failed to create default persona file: ${e.message}`)
             })
-        files.push(globalPath)
-        logging.info(`Created default persona file at ${globalPath}`)
+        selectedFile = globalPath
+        logging.info(`Using newly created default persona file: ${selectedFile}`)
     }
 
-    // merge configs: later files override earlier ones
-    let wsHasStar = false
```
ashishrp-aws, line 221: how are you handling the `*` permission?

ashishrp-aws (approved):

PR #1533, opened 2025-06-05
fix: add fsReplace tool to batch edits

jguoamz:
## Problem
- LLM does not batch small edits on the same file together and instead calls fsWrite multiple times which adds latency
- LLM sometimes does not respect the the fact that parent directory must exist before file creation

## Solution
- Break fsWrite into fsWrite and fsReplace tool where fsReplace accepts an array of diffs to allow LLM to batch edits on the same file together
- Having a simplified fsWrite tool now leads to better performance as LLM can better understand the toolSpec and reduces the chance of it trying to create a new file in a directory that does not exist yet

Summary of the new tools
```
fsWrite
Purpose: Creates new files or adds to existing ones
Commands: create (new/overwrite) and append (add to end)
Key point: Parent directories must exist beforehand
Best for: New files, complete rewrites, or adding to file ends

fsReplace
Purpose: Makes targeted changes within existing files
Method: Uses oldStr/newStr pairs for precise replacements
Multiple changes: Supports batch edits via diffs array
Key point: Each oldStr must match exactly and be unique in the file
Best for: Surgical edits to specific parts of files
```

## License


leigaol:
can you do a global search in the agenticChatController and search for where 'fsWrite' is used and consider adding 'fsReplace' when necessary?

For example, at 
```
#getTools(session: ChatSessionService) {
        const tools = this.#features.agent.getTools({ format: 'bedrock' })

        // it's disabled so filter out the write tools
        if (!session.pairProgrammingMode) {
            return tools.filter(tool => !['fsWrite', 'executeBash'].includes(tool.toolSpecification?.name || ''))
        }
        return tools
    }
``

File: server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsReplace.ts [hunk 1]
```diff
@@ -0,0 +1,159 @@
...
+            diff.newStr = ''
+        }
+        // Normalize oldStr and newStr to match fileContent's line ending style
+        const normalizedOldStr = diff.oldStr.split(/\r\n|\r|\n/).join(lineEnding)
+        const normalizedNewStr = diff.newStr.split(/\r\n|\r|\n/).join(lineEnding)
+
+        const matches = [...fileContent.matchAll(new RegExp(escapeRegExp(normalizedOldStr), 'g'))]
+
+        if (matches.length === 0) {
+            throw new Error(`No occurrences of "${diff.oldStr}" were found`)
```
san2488, line 147: Should we return "X/Y diffs were applied before this failure" to help with partial failures?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsReplace.ts
```diff
@@ -0,0 +1,159 @@
...
+            name: 'fsReplace',
+            description:
+                'A tool for search and replace contents of an existing file.\n\n' +
+                '## Overview\n' +
+                'This tool replaces sections of content in an existing file using `oldStr`/`newStr` blocks that define exact changes to specific parts of the file. You MUST ALWAYS bundle as many changes as you can by populating the diffs array with different `oldStr`/`newStr` pairs. You MUST ALWAYS read the contents of the file you want to update first before using the tool to ensure successful result.\n\n' +
+                '## When to use\n' +
+                '- When you need to make targeted changes to specific parts of a file\n' +
+                '- When you need to update multiple sections of the same file\n' +
+                '## When not to use\n' +
+                '- When `oldStr` and `newStr` are the exact same, skip calling this tool because no changes are needed\n' +
```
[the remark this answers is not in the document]
jguoamz: I added this because I observed this happen once. The explanation for this tool call was like "no changes was needed" which is why I added this

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/fsReplace.ts
jguoamz: good callout, I will update the tool

jguoamz: updated

ctlai95 (approved):

PR #1542, opened 2025-06-05
fix: incorrect history when user aborts in-progress toolUse

jguoamz:
## Problem
- incorrect history when user aborts in-progress toolUse, this leads to incorrect history which confuses LLM and takes long time to recover

## Solution
- Populating the previous toolResult back when user aborts instead of simply dropping it

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -589,6 +592,13 @@ export class ChatDatabase {
 
         //  Make sure the last stored message is from the assistant (type === 'answer'), else drop
         if (messages.length > 0 && messages[messages.length - 1].type === ('prompt' as ChatItemType)) {
+            // When user aborts some in-progress tooluse event, we should still send the previous toolResult back
+            if (messages[messages.length - 1].userInputMessageContext?.toolResults) {
+                if (newUserMessage.userInputMessage && newUserMessage.userInputMessage?.userInputMessageContext) {
```
jngyuamz, line 597: Can this be simplified to:
```diff
if (messages[messages.length - 1].userInputMessageContext?.toolResults) {
-  if (newUserMessage.userInputMessage && newUserMessage.userInputMessage?.userInputMessageContext) {
+  if (newUserMessage.userInputMessage?.userInputMessageContext) {

```

jguoamz: fixed

jngyuamz (approved):

jguoamz:
> nit: Could we add a unit test for this case?

normally yes, but since we have a new refactoring soon, will skip to deploy the fix first

PR #1548, opened 2025-06-06
Remove auto

ctlai95:
## License

### DIFF
--- a/chat-client/src/client/texts/modelSelection.ts
+++ b/chat-client/src/client/texts/modelSelection.ts
@@ -10,8 +10,8 @@ type ModelDetails = {
 }
 
 const modelRecord: Record<BedrockModel, ModelDetails> = {
-    [BedrockModel.CLAUDE_3_7_SONNET_20250219_V1_0]: { label: 'Claude Sonnet 3.7' },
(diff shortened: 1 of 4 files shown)


XiaoxuanLu (approved):
Tested and worked for VS

chat-client/src/client/texts/modelSelection.ts
```diff
@@ -23,7 +23,8 @@ export const modelSelection: ChatItemFormItem = {
     type: 'select',
     id: 'model-selection',
     options: modelOptions,
-    placeholder: 'Auto',
+    mandatory: true,
+    hideMandatoryIcon: true,
     border: false,
     autoWidth: true,
```
zuoyaofu, line 29: Nit: Do you think adding `value: defaultModelId` here to set default value could potentially avoid the bug @XiaoxuanLu produced? (Not important as users will not have memory of previous builds we tested)

zuoyaofu (approved):

chat-client/src/client/texts/modelSelection.ts
ctlai95: Yeah I'm not sure if adding a default value here would solve that problem anyway because it would only change the UI and not the history

PR #1549, opened 2025-06-06
fix: diff reports no lines added or removed

awschristou:
## Problem

When agentic chat modifies existing files, chat often reports the delta as +0 -0 (no file changes) on Windows systems.

image

This causes the following problems:
- clicking on the diff shows the entire file as a new change, instead of just the changed contents
- clicking Undo will delete the file, because chat thinks the file was newly added as a result of the modification

## Causes

- [getTextDocument](https://github.com/aws/language-servers/blob/9082323d1affe9cb71001aa76a216b690e892b06/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts#L265) is called with a mixture of strings that contain [uris](https://github.com/aws/language-servers/blob/9082323d1affe9cb71001aa76a216b690e892b06/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts#L245) and local filesystem paths ([1](https://github.com/aws/language-servers/blob/9082323d1affe9cb71001aa76a216b690e892b06/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts#L1006), [2](https://github.com/aws/language-servers/blob/9082323d1affe9cb71001aa76a216b690e892b06/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts#L1061))
- getTextDocument calls `workspace.getTextDocument` which is expecting a string in **uri** format. The calls made using local filesystem paths fail 100% of the time.
- when calls to `workspace.getTextDocument` fail, getTextDocument falls back to directly reading from the filesystem, however it tries to do this by parsing the given input (which is sometimes a local file path) into a Uri. This frequently throws an error, which causes getTextDocument to return undefined. Calling code incorrectly assumes this means its working with a new file.
- when getTextDocument is successful, the chat controller calls it again after modifying the file, which attempts to load from the LSP workspace (workspace.getTextDocument). There are times when the LSP workspace is not updated with the latest content, because some IDEs are slow to pick up changes from files altered outside of the IDE process. For example, some IDEs don't register file content changes when they do not have active focus until after the IDE receives focus again. As a result, the chat controller thinks the file contents before and after a change are the same.
- on Windows, there is some variation in the URI used to store each file in the LSP workspace. For example, VS Code currently looks like `file:///C%3A/Foo/bar.txt` but Visual Studio looks like `file:///C:/Foo/bar.txt`. When requesting LSP workspace files, the URI must be exact, and is case sensitive.

## Solution

- separate the getTextDocument calls, so callers can explicitly request content using a uri (`getTextDocumentFromUri`) or a filepath (`getTextDocumentFromPath`)
- when requesting a testDocument using a filepath on Windows, attempt to retrieve the document from the LSP workspace using a combination of different URI styles, and drive casings.
  - This increases the success rate obtaining the content from the LSP workspace.
  - This reduces the likelihood the system thinks the file is newly created
- only fetch file contents from the filesystem (not the LSP workspace) after the fsWrite tool alters a file, so that calling code has the true file contents from after a file modification.
  - This reduces the likelihood the system thinks the file has not been modified

This change was tested in Visual Studio, and using the sample VS Code extension in this repo. In both cases, I was able to see the problem occur before my change, and disappear after my change.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -1003,7 +1003,7 @@ export class AgenticChatController implements ChatHandlers {
 
                 if (toolUse.name === 'fsWrite') {
                     const input = toolUse.input as unknown as FsWriteParams
-                    const document = await this.#triggerContext.getTextDocument(input.path)
+                    const document = await this.#triggerContext.getTextDocumentFromPath(input.path, true, true)
                     session.toolUseLookup.set(toolUse.toolUseId, {
                         ...toolUse,
                         fileChange: { before: document?.getText() },
@@ -1058,7 +1058,13 @@ export class AgenticChatController implements ChatHandlers {
                         break
                     case 'fsWrite':
                         const input = toolUse.input as unknown as FsWriteParams
-                        const doc = await this.#triggerContext.getTextDocument(input.path)
+                        // Load from the filesystem instead of workspace.
+                        // Workspace is likely out of date - when files
+                        // are modified external to the IDE, many IDEs
+                        // will only update their file contents (which
+                        // then propagates to the LSP) if/when that
+                        // document receives focus.
+                        const doc = await this.#triggerContext.getTextDocumentFromPath(input.path, false, true)
                         const chatResult = await this.#getFsWriteChatResult(toolUse, doc, session)
                         const cachedToolUse = session.toolUseLookup.get(toolUse.toolUseId)
                         if (cachedToolUse) {
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
@@ -242,7 +242,7 @@ export class AgenticChatTriggerContext {
         if (textDocumentIdentifier?.uri === undefined) {
             return
         }
-        const textDocument = await this.getTextDocument(textDocumentIdentifier.uri)
+        const textDocument = await this.getTextDocumentFromUri(textDocumentIdentifier.uri)
 
         return textDocument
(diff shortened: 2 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1003,7 +1003,7 @@ export class AgenticChatController implements ChatHandlers {
 
                 if (toolUse.name === 'fsWrite') {
                     const input = toolUse.input as unknown as FsWriteParams
-                    const document = await this.#triggerContext.getTextDocument(input.path)
+                    const document = await this.#triggerContext.getTextDocumentFromPath(input.path, true, true)
```
jguoamz, line 1006: I think even here we should only load from fs because edits requires fs content, not LSP content

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/agenticChatTriggerContext.ts
```diff
@@ -242,7 +242,7 @@ export class AgenticChatTriggerContext {
         if (textDocumentIdentifier?.uri === undefined) {
             return
         }
-        const textDocument = await this.getTextDocument(textDocumentIdentifier.uri)
+        const textDocument = await this.getTextDocumentFromUri(textDocumentIdentifier.uri)
```
jguoamz, line 245: can we load from fs here as well to keep it consistent?

awschristou: I'm not familiar with most of this code flow. Do you know what the impact would be if we avoid using the workspace altogether?

jguoamz (approved):
will approve this to address customer issue, track follow up action items separately

ctlai95 (approved):

PR #1566, opened 2025-06-07
fix(amazonq): differentiate listWorkspaceMetadata failure and empty result

LiGaCu:
## Problem

On backend side, we are seeing occasional `ConflictException` from `CreateWorkspace` API, which means users are trying to call that API for an existing running workspace. 

If `listWorkspaceMetadata` call fails due to factors like network issues, throttling, etc. , the current code would mistakenly treat that the same as getting an empty result, which would then triggers an unnecessary `CreateWorkspace` call.

## Solution

Differentiate `listWorkspaceMetadata` failure and empty result, skipping `resetRemoteWorkspaceId` and `handleWorkspaceCreatedState` when the `listWorkspaceMetadata` API fails.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
@@ -417,13 +417,18 @@ export class WorkspaceFolderManager {
 
     private async checkRemoteWorkspaceStatusAndReact(skipUploads: boolean = false) {
         this.logging.log(`Checking remote workspace status for workspace [${this.workspaceIdentifier}]`)
-        const { metadata, optOut } = await this.listWorkspaceMetadata(this.workspaceIdentifier)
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceFolderManager.ts
```diff
@@ -667,14 +672,17 @@ export class WorkspaceFolderManager {
...
     }> {
         let metadata: WorkspaceMetadata | undefined | null
         let optOut = false
+        let error: any
         try {
             const params = workspaceRoot ? { workspaceRoot } : {}
             const response = await this.serviceManager.getCodewhispererService().listWorkspaceMetadata(params)
             metadata = response && response.workspaces.length ? response.workspaces[0] : null
         } catch (e: any) {
+            error = e
```
ege0zcan, line 685: Given we are now returning the error, it would be better to use a type to define the error. I think it should be possible

PR #1572, opened 2025-06-09
fix(amazonq): utg shouldnt throw when there is no corresponding config as its not handled at callers

Will-ShaoHua:
## Problem
```
Error: {
  "code": -32603,
  "message": "Request aws/textDocument/inlineCompletionWithReferences failed with message: lang not supported by utg completion"
}
```

will break generate completion flow if config is not found

## License


leigaol (approved):

Will-ShaoHua:
```
781 passing (5s)
  10 pending
  1 failing

  1) TestIntentDetector
       isTestFile
         should throw if language is not in the supported language set:
     AssertionError [ERR_ASSERTION]: Missing expected exception.
      at Context.<anonymous> (src/shared/supplementalContextUtil/unitTestIntentDetection.test.ts:21:24)
      at processImmediate (node:internal/timers:476:21)

```

PR #1416, opened 2025-05-22
feat: apply a max 200MB total history size

jngyuamz:
## Moved to https://github.com/aws/language-servers/pull/1587.
Cleaned up the commit history, use rebase instead of merge. 

## Problem
1. Current history logic is mixing up the history sent to service and the history persisted in disk.
2. The limit on history size is 600,000 chars or 250 messages, which is way to small.

## Solution
1. Separate the "fix history" logic for service requests and DB
    - History for service requests:
       - Length is less than 250 messages(ensured when retrieving from DB) and 600,000 chars
       - Drop the last empty assistant message and the user prompt before that
       - Ensure the messages have a valid sequence - alternating sequence of user prompt followed by a assistant message
       - Validate and fix tool usage
     - Persisted history in DB:
        - addMessage now skips empty assistant message
        - (Already in place) Skip adding message if the agent loop is cancelled, this is likely the root cause of incorrect message sequence.
2. Maintain a max 200MB total history file size
    - An async check/trim is performed on startup
       - If the total valid history size under the history folder is larger than 200MB, remove the oldest messages across all history files until the total history size is under 75% of the limit. The implementation uses a priority queue to order the messages. The process takes 1.5 sec when avg message size is 5000 chars.
       - Added timestamps to persisted conversations and messages
       - Legacy records don't have timestamps in conversations and messages, use the updatedAt under Tab as the timestamp for all those messages when ordering

## Pending items
- Fix the agenticChatController unit tests(which is causing the CI failure) ✅ 
- Fix the merge issue ✅ 
- Fix improper variable names ✅ 
- Dedupe `calculateDatabaseSize` ✅ 
- ~~Error handling within `calculateDatabaseSize`~~
- Improvement: Store messages in pairs ✅ 
- Handle this edge case: https://github.com/aws/language-servers/pull/1527, will do in the next PR.
- When a cancellation happen, the last user prompt got removed. Need to retain that user prompt instead of removing. ✅ 
- Need a similar logic as `trimHistoryToMaxLength` to make sure the sequence of the message starts at a valid point.

## Future improvements
1. Delete empty history file
2. Telemetry for history trimming time

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1587, opened 2025-06-10
feat: apply a max 200MB total history size

jngyuamz:
## Problem
1. Current history logic is mixing up the history sent to service and the history persisted in disk.
2. The limit on history size is 600,000 chars or 250 messages, which is way to small.

## Solution
1. Separate the "fix history" logic for service requests and DB
    - History for service requests:
       - Length is less than 250 messages(ensured when retrieving from DB) and 600,000 chars
       - Drop the last empty assistant message and the user prompt before that
       - Ensure the messages have a valid sequence - alternating sequence of user prompt followed by a assistant message
       - Validate and fix tool usage
     - Persisted history in DB:
        - addMessage now skips empty assistant message
        - (Already in place) Skip adding message if the agent loop is cancelled, this is likely the root cause of incorrect message sequence.
2. Maintain a max 200MB total history file size
    - An async check/trim is performed on startup
       - If the total valid history size under the history folder is larger than 200MB, remove the oldest messages across all history files until the total history size is under 75% of the limit. The implementation uses a priority queue to order the messages. The process takes 1.5 sec when avg message size is 5000 chars.
       - Added timestamps to persisted conversations and messages
       - Legacy records don't have timestamps in conversations and messages, use the updatedAt under Tab as the timestamp for all those messages when ordering

## Pending items
- Fix the agenticChatController unit tests(which is causing the CI failure) ✅ 
- Fix the merge issue ✅ 
- Fix improper variable names ✅ 
- Dedupe `calculateDatabaseSize` ✅ 
- ~~Error handling within `calculateDatabaseSize`~~
- Improvement: Store messages in pairs ✅ 
- Handle this edge case: https://github.com/aws/language-servers/pull/1527, will do in the next PR.
- When a cancellation happen, the last user prompt got removed. Need to retain that user prompt instead of removing. ✅ 
- Need a similar logic as `trimHistoryToMaxLength` to make sure the sequence of the message starts at a valid point. ✅ 

## Future improvements
1. Delete empty history file
2. Telemetry for history trimming time

## Previous PRs: 
- https://github.com/aws/language-servers/pull/1374
- https://github.com/aws/language-servers/pull/1416

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
@@ -35,7 +35,7 @@ export class ToolResultValidationError extends Error {
 }
 
 export const EMPTY_CONVERSATION_LIST_ID = 'empty'
-// Maximum number of messages to send in request
+// Maximum number of history messages to include in each request to the LLM
 const maxConversationHistoryMessages = 250
 
 /**
@@ -396,7 +396,7 @@ export class ChatDatabase {
                     isOpen: true,
                     tabType: tabType,
                     title: tabTitle,
-                    conversations: [{ conversationId, clientType, messages: [message] }],
+                    conversations: [{ conversationId, clientType, updatedAt: new Date(), messages: [message] }],
                 })
             }
         }
@@ -423,9 +423,9 @@ export class ChatDatabase {
     }
 
     /**
-     * Prepare the history messages for service request `GenerateAssistantResponseCommandInput` to maintain the following invariants:
+     * Prepare the history messages for service request and fix the persisted history in DB to maintain the following invariants:
      * 1. The history contains at most MaxConversationHistoryMessages messages. Oldest messages are dropped.
-     * 2. The first message is from the user and the last message is from the assistant.
+     * 2. The first message is from the user and without any tool usage results, and the last message is from the assistant.
      *    The history contains alternating sequene of userMessage followed by assistantMessages
      * 3. The toolUse and toolResult relationship is valid
      * 4. The history character length is <= MaxConversationHistoryCharacters - newUserMessageCharacterCount. Oldest messages are dropped.
@@ -467,6 +467,17 @@ export class ChatDatabase {
         return allMessages
     }
 
+    /**
+     * Finds a suitable "break point" index in the message sequence.
+     *
+     * It ensures that the "break point" is at a clean conversation boundary where:
+     * 1. The message is from a user (type === 'prompt')
+     * 2. The message doesn't contain tool results that would break tool use/result pairs
+     * 3. The message has a non-empty body
+     *
+     * @param allMessages The array of conversation messages to search through
+     * @returns The index to trim from, or undefined if no suitable trimming point is found
+     */
     private findIndexToTrim(allMessages: Message[]): number | undefined {
         for (let i = 2; i < allMessages.length; i++) {
             const message = allMessages[i]
@@ -588,13 +599,23 @@ export class ChatDatabase {
             return
         }
 
-        //  Make sure the first message is from the user (type === 'prompt'), else drop
+        // Make sure the first message sent to LLM is from the user (type === 'prompt'), else drop
         while (messages.length > 0 && messages[0].type === ('answer' as ChatItemType)) {
             messages.shift()
             this.#features.logging.debug('Dropped first message since it is not from user')
         }
 
-        //  Make sure the last message is from the assistant (type === 'answer'), else add a cancellation response
+        // Make sure the first user message doesn't have tool usage results.
+        while (
+            messages.length > 0 &&
+            messages[0].type === ('prompt' as ChatItemType) &&
+            !this.isValidUserMessageWithoutToolResults(messages[0])
+        ) {
+            messages.splice(0, 2) // Remove first user-assistant pair
+            this.#features.logging.debug('Dropped the first message pair since the user message has tool usage results')
+        }
+
+        //  Make sure the last message is from the assistant (type === 'answer'), else add a dummy response
         if (messages.length > 0 && messages[messages.length - 1].type === ('prompt' as ChatItemType)) {
             // Add an assistant response to both request and DB to maintain a valid sequence
             const dummyResponse: Message = {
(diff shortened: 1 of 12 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatHistoryMaintainer.ts
```diff
@@ -0,0 +1,360 @@
...
+        this.#features.logging.info(`Trimming history took ${trimEnd - trimStart} ms`)
+    }
+
+    /**
+     * Trims chat history across all workspaces to reduce storage size.
+     *
+     * This method:
+     * 1. Loads all database files from the history directory
+     * 2. Creates a priority queue of tabs sorted by oldest message date
+     * 3. Iteratively removes oldest messages in batches until total size is below target
```
jngyuamz, line 83: > [jguoamz](https://github.com/jguoamz): this could result in invalid history right?



> [jngyuamz](https://github.com/jngyuamz): Yes and no
> 
>     * Yes - We make sure we remove messages in pairs, so the persisted messages are always in the right sequence.
> 
>     * No - ^that's the only validation we do here.
> 
> 
> For tool usage, we need the `fixHistory` method to take care of them, that method anyway needs to validate it, I'll revisit that logic

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -326,7 +347,15 @@ export class ChatDatabase {
      * - Updates tab's last updated time
      */
     addMessage(tabId: string, tabType: TabType, conversationId: string, message: Message) {
-        if (this.#initialized) {
+        if (this.isInitialized()) {
+            // Skip adding to history DB if it's an empty response
+            if (isEmptyAssistantMessage(message)) {
+                this.#features.logging.debug(
+                    'The assistant response is empty. Skipped adding this message and removed last user prompt from the history DB'
```
jngyuamz, line 354: > [jguoamz](https://github.com/jguoamz): why?



> [jngyuamz](https://github.com/jngyuamz): Moved from the `fixHistory` method. Instead of looking back to remove it, we make sure we don't add empty assistant message
> 
> ```diff
> -        this.handleEmptyAssistantMessage(allMessages)
> ```

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -453,38 +464,7 @@ export class ChatDatabase {
...
-        tabData.updatedAt = new Date()
-        tabCollection.update(tabData)
-        this.#features.logging.info(`Updated tab data in collection`)
-        return isValid
-    }
-
-    private trimHistoryToMaxLength(messages: Message[]): Message[] {
-        while (messages.length > MaxConversationHistoryMessages) {
-            // Find the next valid user message to start from
-            const indexToTrim = this.findIndexToTrim(messages)
```
jngyuamz, line 474: > [jguoamz](https://github.com/jguoamz): imo, we still need logic similar to this for trimming



> [jngyuamz](https://github.com/jngyuamz): This logic is in the call at line 441. We pass a param to `getMessage` to ensure the max number of message < 250.
> 
> ```diff
> +        // 1. Make sure the length of the history messages don't exceed MaxConversationHistoryMessages
> +        let allMessages = this.getMessages(tabId, maxConversationHistoryMessages)
> ```



> [jngyuamz](https://github.com/jngyuamz): I get what you mean. We need a trim logic to make sure this sequence of message starts at a valid point.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/util.ts
```diff
@@ -322,3 +415,14 @@ function getTabTypeIcon(tabType: TabType): IconType {
...
     }
 }
+
+/**
+ * Calculates the size of a database file
+ * @param features Features object containing workspace filesystem access
+ * @param dbPath Path to the database file
+ * @returns Promise that resolves to the file size in bytes, or 0 if there's an error
+ */
+export async function calculateDatabaseSize(features: Features, dbPath: string): Promise<number> {
```
jngyuamz, line 425: > [avi-alpert](https://github.com/avi-alpert): i think you should also add a try catch to handle if getFileSize fails







> [jngyuamz](https://github.com/jngyuamz): We don't need try/catch within this method. The callers should handle the errors and decide what they want to do(just use 0 or pass the error to downstream)

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/util.ts
```diff
@@ -74,9 +78,27 @@ export type Message = {
...
     toolUses?: ToolUse[]
+    timestamp?: Date
     shouldDisplayMessage?: boolean
 }
 
+/**
+ * Represents a tab with its database collection reference, database name, and timestamp information
+ * for use in history trimming operations.
+ */
+export type TabWithContext = {
```
avi-alpert, line 89: nitpick: could we rename this variable? I'm concerned that calling it TabWithContext will cause confusion since we do store actually store Context in history

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/util.ts
```diff
@@ -160,6 +182,14 @@ export function chatMessageToMessage(chatMessage: StreamingMessage): Message {
     }
 }
 
+export function isEmptyAssistantMessage(message: Message) {
```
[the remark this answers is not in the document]
jngyuamz: I have the same question. Need to follow up on the original code author

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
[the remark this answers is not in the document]
jguoamz: I don't think this ever happens? (emptyAssistantMessage)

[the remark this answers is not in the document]
jngyuamz: Checked with [tsmithsz](https://github.com/tsmithsz), he doesn't think this would happen, I'm removing it.

PR #1601, opened 2025-06-10
fix(amazonq): remove storing zips under workspaceContextArtifacts

LiGaCu:
## Problem

Currently, workspace context workflow is creating zip files and writing them to disk before further processing (uploading to S3, sending notification to backend, etc.), however, when examining the logic, there is no place where it reads from these zip files.

We should avoid saving these unused files to reduce disk usage and prevent leaving residues on users disk.

## Solution

Remove storing zips under `workspaceContextArtifacts`

## Testing

Tested locally and verified that:
1. Zipping and uploading logic works correctly as before.
2. No temporary files are created under `/tmp/aws-language-servers/workspaceContextArtifacts` anymore.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceContextServer.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceContextServer.test.ts
@@ -2,6 +2,7 @@ import { InitializeParams, Server } from '@aws/language-server-runtimes/server-i
 import { TestFeatures } from '@aws/language-server-runtimes/testing'
 import sinon from 'ts-sinon'
 import { WorkspaceContextServer } from './workspaceContextServer'
+import { AmazonQTokenServiceManager } from '../../shared/amazonQServiceManager/AmazonQTokenServiceManager'
 
 describe('WorkspaceContext Server', () => {
     let features: TestFeatures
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceContextServer.test.ts
```diff
@@ -34,6 +35,10 @@ describe('WorkspaceContext Server', () => {
                 },
             } as InitializeParams)
 
+            // Create a stub for the static getInstance method
+            const getInstanceStub = sinon.stub(AmazonQTokenServiceManager, 'getInstance')
+            getInstanceStub.throws(new Error('Deliberate error to exit the test at onInitialized'))
```
ege0zcan, line 40: I didn't understand why we need this error now

LiGaCu: The previous test code would exit at the step of initializing `ArtifactManager` when running this test, due to the exception thrown from `this.tempDirPath = path.join(this.workspace.fs.getTempDirPath(), ARTIFACT_FOLDER_NAME)`, which is removed by this change.

Now the test code can proceed beyond that step and something afterwards (open handles? unresolved promises?) keeps Node.js process alive and hanging:
- https://github.com/aws/language-servers/actions/runs/15570836209/job/43846060469?pr=1601

So, this deliberate error is just a workaround until we figure out what is wrong with this specific test setup and how to gracefully exit it.

ege0zcan: Good to know, thanks for clarifying this

PR #1609, opened 2025-06-10
fix: remove the tool from the mapping after user delete the server

Randall-Jiang:
## Problem
When user remove a server, we did not remove the server's tool name from set. Which might case when next time user add the same server, there would be a conflict when we create name mapping for serverName_toolName

https://github.com/user-attachments/assets/6ffd66fb-3054-4e7f-98ec-fbc71203d526

(00:00:55 in this video)

## Solution
We are removing the tool name in set for creating the Map. To make sure when user remove the sever, the tool would be completely removed from the memory as well.

https://github.com/user-attachments/assets/86648754-9555-4e9f-8472-b8959e2181ff

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolServer.ts
@@ -91,6 +91,7 @@ export const McpToolsServer: Server = ({ credentialsProvider, workspace, logging
         // 1) remove old tools
         for (const name of registered[server] ?? []) {
             agent.removeTool(name)
+            allNamespacedTools.delete(name)
         }
         registered[server] = []
 


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/toolServer.ts
```diff
@@ -91,6 +91,7 @@ export const McpToolsServer: Server = ({ credentialsProvider, workspace, logging
         // 1) remove old tools
         for (const name of registered[server] ?? []) {
             agent.removeTool(name)
+            allNamespacedTools.delete(name)
```
ashishrp-aws, line 94: Are we sure we are not deleting tools from servers that are not deleted?

ashishrp-aws: we do not have handling for servers with duplicate names right now...we are tracking it as a P1

PR #1614, opened 2025-06-11
feat: language keywords detector impl for NEP

Will-ShaoHua:
## License


floralph (approved):

server/aws-lsp-codewhisperer/src/language-server/inline-completion/auto-trigger/languageDetector.ts
```diff
@@ -0,0 +1,379 @@
...
+        }
+
+        return this.detectors.get(normalizedLanguage)!
+    }
+}
+
+/**
+ * Base class for language detectors with common functionality
+ */
+abstract class BaseLanguageDetector implements LanguageDetector {
```
leigaol, line 54: please add doc string to explain what this class does. It is hard to follow by reading the code and function name

Will-ShaoHua: let me follow up this as the logic of this class is naive and all it does is to determine if a word/string belongs to a language's keywords

PR #1624, opened 2025-06-11
fix: fix for overwriting in workspace level config and persona files

ashishrp-aws:
## Problem
At workspace level config and persona files are getting overwritten and deleting existing mcps when an mcp initialization fails on saving. as workspace level paths arrive as URIs with `file:/` at the start of path.

## Solution
- fixes for URI path format for config and persona files at workspace
- storing newly added servers which have failed, to delete only them and not manually added mcps by users directly to config file.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
@@ -25,6 +25,7 @@ import {
     McpServerStatus,
 } from './mcpTypes'
 import { TelemetryService } from '../../../../shared/telemetry/telemetryService'
+import { URI } from 'vscode-uri'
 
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
```diff
@@ -1162,10 +1177,15 @@ export class McpEventHandler {
 
             if (Array.isArray(workspacePersonaPaths) && workspacePersonaPaths.length > 0) {
                 try {
+                    // Convert URI format to filesystem path if needed
+                    const personaPath = workspacePersonaPaths[0].startsWith('file:')
```
laileni-aws, line 1181: nit: We can move this to `mcpUtils.ts`

laileni-aws (approved):
LGTM with some nits

PR #1653, opened 2025-06-13
fix: adding normalizePathFromUri to mcpUtils to handle uri paths

ashishrp-aws:
## Problem
normalize path from URI was being used in multiple places in mcpEventHandler

## Solution
- created a function to handle it along with unit tests.

## License


laileni-aws (approved):
Awesome, Thanks for refactoring!

PR #1676, opened 2025-06-16
fix: update MCP tools implementation

ashishrp-aws:
## Problem
Current tools setup for MCP servers follows mcpServerName___toolName format. product requirement is to send just tools with tool names and include serverName as prefix only in case a tool has duplicate/ other tool with same name.

## Solution
- Updated to use only toolNames and removed serverName from request.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -1655,7 +1655,7 @@ export class AgenticChatController implements ChatHandlers {
                         content: {
                             header: {
                                 icon: 'tools',
-                                body: `${McpManager.instance.getOriginalToolNames(toolUse.name!)?.toolName}`,
+                                body: `${originalToolName ?? (toolType || toolUse.name)}`,
                                 status: {
                                     status: isAccept ? 'success' : 'error',
                                     icon: isAccept ? 'ok' : 'cancel',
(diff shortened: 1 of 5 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpUtils.ts
```diff
@@ -379,9 +379,19 @@ export function createNamespacedToolName(
         }
     }
 
+    // Sanitize the tool name
+    const sanitizedToolName = sanitizeName(toolName)
+
+    // First try to use just the tool name if it's not already in use
+    if (!allNamespacedTools.has(sanitizedToolName)) {
+        allNamespacedTools.add(sanitizedToolName)
+        toolNameMapping.set(sanitizedToolName, { serverName, toolName })
```
Randall-Jiang, line 388: why we are adding [sanitizedToolName, {serverName, toolName}] here, but adding [serverName___sanitizedToolName, {serverName, toolName} ] later on line 399?

ashishrp-aws: because the toolName for a tool by default will just be tool name itself, if it has duplicate then we add the mcp server name prefix and namespacing which is a map considers that as the tool name. for example, if `fetch` exists, then another `fetch` tool becomes `server___fetch`. namespacing will consider this as a tool.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1655,7 +1655,7 @@ export class AgenticChatController implements ChatHandlers {
                         content: {
                             header: {
                                 icon: 'tools',
-                                body: `${McpManager.instance.getOriginalToolNames(toolUse.name!)?.toolName}`,
+                                body: `${originalToolName ?? (toolType || toolUse.name)}`,
```
laileni-aws, line 1658: nit: We can use `toolName` directly.

PR #1684, opened 2025-06-17
fix(amazonq): filter languages at workspace context server onDeleteFiles

LiGaCu:
## Problem

On backend, we are receiving unexpected WebSocket requests that is complaining about the programming language field:
>   Input should be 'java', 'python', 'javascript' or 'typescript' [type=enum, input_value='json', input_type=str]

## Solution

Filter languages at workspace context server onDeleteFiles

Tested this change filtered out unexpected events:

image

## License


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/workspaceContextServer.ts
```diff
@@ -421,9 +421,11 @@ export const WorkspaceContextServer = (): Server => features => {
                     continue
                 }
 
-                const programmingLanguages = artifactManager.handleDeletedPathAndGetLanguages(file.uri, workspaceFolder)
+                const programmingLanguages = artifactManager
+                    .handleDeletedPathAndGetLanguages(file.uri, workspaceFolder)
+                    .filter(lang => SUPPORTED_WORKSPACE_CONTEXT_LANGUAGES.includes(lang))
```
leigaol, line 426: consider adding unit test for this.

also makes sense to move this logic inside artifactManager

LiGaCu: > consider adding unit test for this.

None of these listeners are having unit test setup nowadays. We probably should add for all of them. Will track that separately.

> also makes sense to move this logic inside artifactManager

Makes sense. Moved it.

PR #1691, opened 2025-06-18
feat: adding current working directory to the stdio transport for mcp…

ashishrp-aws:
## Problem

For some MCP servers, like in case of SQLite MCP server, when given relative path of the db file in the workspace the server initialization is failing. Q cli on the other hand supports relative paths in arguments.

## Solution

- adding CWD to the stdio transport to fix the issue.

https://github.com/user-attachments/assets/7577a567-cb9f-4e5b-9c02-5daf0dcac32e

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpManager.ts
@@ -206,6 +206,7 @@ export class McpManager {
                 command: cfg.command,
                 args: cfg.args ?? [],
                 env: mergedEnv,
+                cwd: URI.parse(this.features.workspace.getAllWorkspaceFolders()[0].uri).fsPath,
             })
             const client = new Client({
                 name: `mcp-client-${serverName}`,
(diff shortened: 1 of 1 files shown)


chungjac (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpManager.ts
```diff
@@ -206,6 +206,7 @@ export class McpManager {
                 command: cfg.command,
                 args: cfg.args ?? [],
                 env: mergedEnv,
+                cwd: URI.parse(this.features.workspace.getAllWorkspaceFolders()[0].uri).fsPath,
```
laileni-aws, line 209: nit: Its good to have a try catch here if there is any error while parsing we can return undefined.

chungjac (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpManager.ts
```diff
@@ -202,11 +202,22 @@ export class McpManager {
...
-            })
+            }
+
+            try {
+                const workspaceFolders = this.features.workspace.getAllWorkspaceFolders()
+                if (workspaceFolders.length > 0) {
+                    transportConfig.cwd = URI.parse(workspaceFolders[0].uri).fsPath
+                }
+            } catch {
+                // No workspace folder available, continue without cwd
```
laileni-aws, line 217: Can we add a log if this throws any error? So, that it will be easy to debug issues!

PR #1701, opened 2025-06-19
fix: make file collection for indexing non blocking

leigaol:
## Problem
Before doing local indexing for `@code`, `@file` ,etc and `@workspace`,  there is a function `processWorkspaceFolders` that collects files. That function is compute extensive, which blocks the node js event loop even though it is behind an async method. As node js is single threaded, when the node js event loop is blocked, the new events like the JSON RPC request cannot be handled in time, causing the UI to get stuck in "Thinking....".

## Solution

Use ignore-walk API (which does not block event loop) to collect files in OS and let it early terminates when max size limit hits. 

Tested by opening entire workspace, can use chat when `processWorkspaceFolders`  is going on.

`listFilesWithGitignore` is borrowed from the workspace indexing library. 

The duplicate vector indexing invocation bug is also avoided in this PR...

Note: This does not fully resolve the stuck in "Thinking...." problem as the vecLib.buildIndex() needs some improvements to not block the event loop. Once user is past `processWorkspaceFolders` phrase, the `vecLib.buildIndex()` can still block event loop. I will work on another change for this. ( only visible internally)

Before the agentic chat LSP migration, the previous local workspace LSP had mechanism to free the event loop but that was not implemented during agentic chat Flare LSP migration. 

I will not invest more time in the unit test of this PR until this gets tested by others.

## License


server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -433,89 +412,33 @@ export class LocalProjectContextController {
...
+
+        for (const folder of workspaceFolders) {
+            const absoluteFolderPath = path.resolve(URI.parse(folder.uri).fsPath)
+            const filesUnderFolder = await listFilesWithGitignore(absoluteFolderPath)
+            for (const file of filesUnderFolder) {
+                const fileExtName = '.' + getFileExtensionName(file)
+                if (!uniqueFilesToIndex.has(file) && fileExtensions?.includes(fileExtName)) {
+                    try {
+                        const fileSize = fs.statSync(file).size
+                        if (fileSize < sizeConstraints.maxFileSize && sizeConstraints.remainingIndexSize > fileSize) {
```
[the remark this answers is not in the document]
yueny2020: if there is a log of these logs, does that mean we need to increase the maxFileSize?

server/aws-lsp-codewhisperer/package.json
```diff
@@ -51,6 +51,7 @@
         "fast-glob": "^3.3.3",
         "fastest-levenshtein": "^1.0.16",
         "fdir": "^6.4.3",
+        "fs-extra": "^11.3.0",
```
justinmk3, line 54: There really should be no need for fs-extra since I thought the [fs.ts](https://github.com/aws/aws-toolkit-vscode/blob/0819eb3bcfe591a7d18b71e9513856f27b75b10d/packages/core/src/shared/fs/fs.ts) module was ported from aws-toolkit-vscode, or was planned to be.

Adding fs-extra will cause problems with supporting web-mode, that's why we removed it from aws-toolkit-vscode. It's also not needed at all if you port fs.ts from aws-toolkit-vscode (which is more robust anyway).

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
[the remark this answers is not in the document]
leigaol: I can add a log to print how many files are skipped.. just one line

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -433,89 +412,42 @@ export class LocalProjectContextController {
...
+                const fileExtName = '.' + getFileExtensionName(file)
+                if (!uniqueFilesToIndex.has(file) && fileExtensions?.includes(fileExtName)) {
+                    try {
+                        const fileSize = fs.statSync(file).size
+                        if (fileSize < sizeConstraints.maxFileSize) {
+                            if (sizeConstraints.remainingIndexSize > fileSize) {
+                                uniqueFilesToIndex.add(file)
+                                sizeConstraints.remainingIndexSize = sizeConstraints.remainingIndexSize - fileSize
+                            } else {
+                                this.log.info(
```
yueny2020, line 430: nit: should be warn for the log?

leigaol: not a warn because this is expected. we restrict on max size.

yueny2020 (approved):

san2488 (approved):

PR #1713, opened 2025-06-20
feat: surface file operation errors in tooltip

ctlai95:
## Problem

File operation errors should be visible.

## Solution

- Map file operation errors to user-friendly error messages
- Surface error message to the error status description so that it appears on hover

image

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
@@ -88,13 +88,13 @@ export class FileOperationError extends Error {
 // Common file system errors
 export class DirectoryNotFoundError extends FileOperationError {
     constructor(message: string) {
-        super(message, 'The directory does not exist. Please create the directory first.')
+        super(message, 'The directory does not exist.')
     }
 }
 
 export class PermissionDeniedError extends FileOperationError {
     constructor(message: string) {
-        super(message, 'Permission denied. You do not have sufficient permissions to modify this file.')
+        super(message, 'Permission denied.')
     }
 }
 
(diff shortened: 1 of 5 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
```diff
@@ -71,3 +71,194 @@ export function isRequestAbortedError(error: unknown): boolean {
...
+}
+
+// Common file system errors
+export class DirectoryNotFoundError extends FileOperationError {
+    constructor(message: string) {
+        super(message, 'The directory does not exist. Please create the directory first.')
+    }
+}
+
+export class PermissionDeniedError extends FileOperationError {
```
jguoamz, line 95: where did we get the list of errors?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
```diff
@@ -71,3 +71,194 @@ export function isRequestAbortedError(error: unknown): boolean {
...
+            'The file does not exist. Please create the file first or check the path.'
+        )
+    }
+}
+
+export class TextNotFoundError extends FileOperationError {
+    constructor(text: string) {
+        super(
+            `No occurrences of "${text}" were found`,
+            'The text to replace was not found in the file. Please check your input.'
```
jguoamz, line 173: I think we should reword the customer facing string here. "you" does not make sense because it's the LLM that is making the mistake, same for other cases

server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
```diff
@@ -71,3 +71,194 @@ export function isRequestAbortedError(error: unknown): boolean {
...
+        return new TextNotFoundError(noOccurrencesMatch[1])
+    }
+
+    const multipleOccurrencesMatch = message.match(/Multiple occurrences of "(.+)" were found/)
+    if (multipleOccurrencesMatch) {
+        return new MultipleMatchesError(multipleOccurrencesMatch[1])
+    }
+
+    // Default fallback - wrap the original error with the same message
+    return new FileOperationError(message, message)
```
jguoamz, line 246: what about invalid json?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
ctlai95: That is true, I'll update it

jguoamz (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
ctlai95: I organized all the tool errors we were throwing into this file, and also added a few more common errors from nodejs fs

server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
ctlai95: The invalid json case won't happen here, that will be caught outside of the tool altogether

jguoamz (approved):

PR #1732, opened 2025-06-21
fix: include .classpath related files as source code of SSPC

wweitao:
## Problem
.classpath was ignored when uploading source code to language server
## License


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
```diff
@@ -35,6 +35,9 @@ export const IGNORE_PATTERNS = [
     '**/coverage/**',
     // Hidden files
     '**/.*',
+    '!**/.classpath',
```
ege0zcan, line 38: did you test that fast-glob is able to process the negation correctly? in my previous testing I remember not all libraries were able to handle negation

PR #1752, opened 2025-06-24
fix(amazonq): add jitter for websocket client re-connections

LiGaCu:
## Problem

We need to prevent the thundering herd problem when there is a network issue or backend issue causing a large amount of WebSocket clients disconnected.

## Solution

Introduce randomness into the timing of retries to avoid simultaneous requests to re-establish WebSocket connection.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/client.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/client.ts
@@ -82,7 +82,9 @@ export class WebSocketClient {
         // Apply exponential backoff for both unclean closures and failed reconnection attempts
         if (this.reconnectAttempts < this.maxReconnectAttempts) {
             this.reconnectAttempts++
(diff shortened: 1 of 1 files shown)


ramana-keerthi (commented):
It's probably less of a concern now, but we might want to add a similar jitter when we open the connection if we start to see high spikes during the start of the business day

LiGaCu:
> It's probably less of a concern now, but we might want to add a similar jitter when we open the connection if we start to see high spikes during the start of the business day

That wouldn't help the traffic contributing from different users, since the jitter method would only help distribute the traffic across the jitter interval (i.e. it only helps actions that happen simultaneously at second-level but not for the traffic that might be relatively and consistently high for minutes or hours).

But I agree something like that would help the surge traffic from the same user / machine, when opening the IDE that automatically brings up multiple previous windows. It would be something affecting all AmazonQ functionalities, and there is already a recent PR to introduce a Jitter when selecting profile (and our WorkspaceContextServer will wait until profile is selected): 
- https://github.com/aws/aws-toolkit-vscode/pull/7534

PR #1757, opened 2025-06-25
Add persistent pair programming mode setting with database storage and UI synchronization

laileni-aws:
## Problem:
- Agentic coding toggle doesn't persist across chat tabs

## Solution:
- Add persistent pair programming mode setting with database storage and UI synchronization

### Technical Details:
• Maintains backward compatibility by making pairProgrammingMode optional in settings
• Uses type assertions where needed to extend existing interfaces
• Preserves existing settings when updating pair programming mode
• Follows established patterns for model selection persistence

### Tested in these cases:
- new user --> default agentic coding ON
- new chat tab
  - user turns agentic coding OFF
  - user opens new tab
  - agentic coding should be OFF

### Demo:

https://github.com/user-attachments/assets/66b17490-57df-473d-a3b2-7f1bac1f0035

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -215,19 +215,6 @@ export const createMynahUi = (
     let programmingModeCardActive = !pairProgrammingCardAcknowledged
     let contextCommandGroups: ContextCommandGroups | undefined
 
-    // Fetch and set the saved pair programming mode from the server
-    // This will be called later when the UI is ready
-    const fetchSavedPairProgrammingMode = () => {
-        // Request to get the saved pair programming mode
-        messager.onChatPrompt(
-            {
-                prompt: { prompt: '', escapedPrompt: '' },
-                tabId: tabFactory.initialTabId,
-            },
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts
```diff
@@ -844,4 +844,15 @@ export class ChatDatabase {
     setModelId(modelId: string | undefined): void {
         this.updateSettings({ modelId: modelId === '' ? undefined : modelId })
```
ctlai95, line 845: I think here you'll have to do what you did below, otherwise it might reset pairProgrammingMode?

ctlai95: oh in that case why do we need to do the existing settings check for setPairProgrammingMode?

ctlai95 (approved):

PR #1759, opened 2025-06-25
fix(amazonq): fix to include explanation field in mcp tools schema

ashishrp-aws:
## Problem
Many MCP servers do not allow custom fields in tool schema like `explanation` which we use to add a directive message for tool executions. The  mcp tools are breaking and initialization is failing due to this.

## Solution
- Added explanation back to tool schema.
- removed it right before user facing messages with parameters and tool execution.

## License


laileni-aws (approved):
LGTM

PR #1761, opened 2025-06-25
fix(amazonq): fix the order of publishing the chat stop ack message

abhraina-aws:
## Problem
We were printing the chat processing stopped acknowledging message in the wrong order where it was coming after the new message. 

## Solution
We moved the message publication to the mynah.ts file so that we can handle it at source and hence we dont publish it after the message is published.

image
image

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -146,7 +146,32 @@ export const handleChatPrompt = (
     agenticMode?: boolean
 ) => {
     let userPrompt = prompt.escapedPrompt
-    messager.onStopChatResponse(tabId)
+
+    // Check if there's an ongoing request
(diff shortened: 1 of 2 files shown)


chat-client/src/client/mynahUi.ts
```diff
@@ -146,7 +146,32 @@ export const handleChatPrompt = (
     agenticMode?: boolean
 ) => {
     let userPrompt = prompt.escapedPrompt
-    messager.onStopChatResponse(tabId)
+
+    // Check if there's an ongoing request
+    const isLoading = mynahUi.getTabData(tabId)?.getStore()?.loadingChat
```
laileni-aws, line 151: nit: inProgress is better than isLoading because request may be streaming too.

NOT A BLOCKER

tsmithsz (approved):

PR #1763, opened 2025-06-25
fix: emit latency as an int for creating visualizations

chungjac:
## Problem
- https://github.com/chungjac/aws-toolkit-common/blob/e3f6ceb877c560179bc1748e7f3ace2ff7e43b85/telemetry/definitions/commonDefinitions.json#L1817-L1819

## Solution
- emit `latency` as a number

## License


server/aws-lsp-codewhisperer/src/language-server/chat/telemetry/chatTelemetryController.ts
```diff
@@ -193,7 +193,7 @@ export class ChatTelemetryController {
                 cwsprToolUseId: toolUseId?.join(',') ?? '',
                 result,
                 languageServerVersion: languageServerVersion,
-                latency: latency?.join(','),
+                latency: latency,
```
laileni-aws, line 196: nit: we don't need to assign value here 
```
languageServerVersion: languageServerVersion,
latency,
toolCallLatency: toolCallLatency?.join(','),
```
Not a blocker

PR #1766, opened 2025-06-25
fix(amazonq): return empty if nep auto trigger is not triggered

Will-ShaoHua:
## Problem
empty prediction type will case 4xx error

## License


Will-ShaoHua:
followup: should handle manual trigger vs. auto trigger

atonaamz (approved):

PR #1769, opened 2025-06-26
fix(amazonq): make workspace context server upload dependency chunks sequentially

LiGaCu:
## Problem

The current workspace context server dependency upload logic has two key performance issues:
1. **Memory inefficiency**: During dependency discovery, all chunked zip files are collected into a single array before upload, causing them to remain in memory simultaneously (up to 8GB before compression) and preventing individual garbage collection.
2. **Upload concurrency spikes**: While compression is sequential, uploads run concurrently across workspace folders and languages, creating traffic spikes to the backend.

## Solution

Refactor from EventEmitter to callback-based approach in LanguageDependencyHandler:
1. **Immediate processing**: Process individual zip files via callback instead of batching, enabling immediate garbage collection after each upload.
2. **Sequential uploads:** Make the callback return a promise that's awaited before generating the next zip, eliminating upload concurrency issues.

This approach reduces memory footprint as well as provides smoother, more predictable backend traffic patterns.

### Testing

Tested the change on a local machine successfully:
image

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyHandler/LanguageDependencyHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyHandler/LanguageDependencyHandler.ts
@@ -2,7 +2,6 @@ import { Logging, Workspace, WorkspaceFolder } from '@aws/language-server-runtim
 import * as fs from 'fs'
 import { ArtifactManager, FileMetadata } from '../../artifactManager'
 import path = require('path')
-import { EventEmitter } from 'events'
 import { CodewhispererLanguage } from '../../../../shared/languageDetection'
 import { isDirectory } from '../../util'
 import { DependencyWatcher } from './DependencyWatcher'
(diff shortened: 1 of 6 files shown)


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyHandler/LanguageDependencyHandler.ts
```diff
@@ -227,15 +219,15 @@ export abstract class LanguageDependencyHandler<T extends BaseDependencyInfo> {
...
                 const totalChunkSize = chunk.reduce((sum, dep) => sum + dep.size, 0)
                 // Log chunk statistics
                 this.logging.log(
-                    `Created zip: ${singleZip.filePath} for #${chunkIndex} chunk containing ${chunk.length} ${this.language} dependencies with total size: ${(
+                    `Created a zip for chunk #${chunkIndex} containing ${chunk.length} ${this.language} dependencies with total size: ${(
                         totalChunkSize /
                         (1024 * 1024)
                     ).toFixed(2)}MB under ${workspaceFolder.name}`
                 )
+                await this.uploadDependencyZip(workspaceFolder, singleZip)
```
yzhangok, line 230: Will this zippped file be cleaned up after upload?

LiGaCu: Historically, these zips are written to disk and there might be leftovers under the corner cases.

This previous PR has already avoided writing these zip files to disk entirely:
- https://github.com/aws/language-servers/pull/1601

PR #1783, opened 2025-06-27
feat: adding logic to merge with previous suggestion

atonaamz:
## License


server/aws-lsp-codewhisperer/src/language-server/inline-completion/mergeRightUtils.ts
```diff
@@ -13,19 +18,115 @@ export function getPrefixSuffixOverlap(firstString: string, secondString: string
...
+        i--
+    }
+    return idx
+}
+
+export function truncateOverlapWithRightContext(
+    rightFileContent: string,
+    suggestion: string,
+    userEdit?: string
+): string {
```
Will-ShaoHua, line 45: @atonaamz i just have a small request, as these "Edits" specific reconciliation logic is new, can we have a simple 
```
if (suggestionType === EDITS) {

} else {

}
```

to branch out these 2 types of suggestions and it can also ensure the "Completion" type suggestion wont regress?

Will-ShaoHua: since these context merge logic is pretty delicate and will potentially require some iterations, it will be easier to branch out these 2 types of suggestions

Will-ShaoHua (approved):

PR #1801, opened 2025-07-01
fix(amazonq): add handling for relative paths for isInWorkspace

abhraina-aws:
## Problem
Sometimes when dealing with relative paths to the workspace our isInWorkspace function breaks. 
image

## Solution
Added handling for relative paths such that we verify its existance.
image

## License

### DIFF
--- a/core/aws-lsp-core/src/util/workspaceUtils.ts
+++ b/core/aws-lsp-core/src/util/workspaceUtils.ts
@@ -3,6 +3,7 @@ import { URI } from 'vscode-uri'
 import { Features } from '@aws/language-server-runtimes/server-interface/server'
 import { CancellationToken } from '@aws/language-server-runtimes/server-interface'
 import { CancellationError } from './awsError'
+import * as fs from 'fs'
 
 type ElementType<T> = T extends (infer U)[] ? U : never
 type Dirent = ElementType<Awaited<ReturnType<Features['workspace']['fs']['readdir']>>>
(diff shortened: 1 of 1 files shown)


core/aws-lsp-core/src/util/workspaceUtils.ts
```diff
@@ -125,7 +126,20 @@ export function isParentFolder(parentPath: string, childPath: string): boolean {
...
 
 export function isInWorkspace(workspaceFolderPaths: string[], filepath: string) {
-    return workspaceFolderPaths.some(wsFolder => isParentFolder(wsFolder, filepath) || wsFolder === filepath)
+    if (path.isAbsolute(filepath)) {
+        return workspaceFolderPaths.some(wsFolder => isParentFolder(wsFolder, filepath) || wsFolder === filepath)
+    } else {
+        // For relative paths, try each workspace folder
+        for (const wsFolder of workspaceFolderPaths) {
+            const absolutePath = path.join(wsFolder, filepath)
+            if (fs.existsSync(absolutePath)) {
```
jguoamz, line 135: we don't need the exists check right?

abhraina-aws: We do. We are practically making up the path here:
            const absolutePath = path.join(wsFolder, filepath)
If we dont check this we will practically get true everytime. What I can do howeer is get rid of the inner loop totally. If a file exists at this location then we can surely say that its in workspace. Let me know what you feel about that.

chungjac (approved):

tsmithsz (approved):

jguoamz (approved):

PR #1814, opened 2025-07-02
fix(amazonq): allow taking .jpg file as image context, add image cont…

yzhangok:
…ext used to transparency list

## Problem
In bug bash, found 2 issues
1) Images sent as context don't show up in context transparency list. All other items sent as context appear in Context dropdown at top of response
2) .JPG file is not accepted, .JPEG is

## Solution
1) Add image context to the trigger context, so the image contexts used can be rendered in ui
2) Both .jpg and .jpeg files use the exact same JPEG compression algorithm and file structure, so read a file as binary, we're getting the raw data, replace the format to 'jpeg' if it's originally 'jpg'.
Screenshot 2025-07-02 at 1 25 18 AM

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -681,7 +681,12 @@ export class AgenticChatController implements ChatHandlers {
                 params.context,
                 params.tabId
             )
-
+            // Add image context to triggerContext.documentReference for transparency
+            await this.#additionalContextProvider.appendCustomContextToTriggerContext(
+                triggerContext,
+                params.context,
+                params.tabId
+            )
             // Get the initial request input
             const initialRequestInput = await this.#prepareRequestInput(
                 params,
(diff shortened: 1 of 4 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -684,7 +684,12 @@ export class AgenticChatController implements ChatHandlers {
                 params.context,
                 params.tabId
             )
-
+            // Add image context to triggerContext.documentReference for transparency
+            await this.#additionalContextProvider.appendCustomContextToTriggerContext(
+                triggerContext,
+                params.context,
```
jngyuamz, line 690: There's a trim logic `return mergedContext.slice(0, MAX_IMAGE_CONTEXT_COUNT)` in `getImageBlocksFromContext`, if some image context got trimmed, we should make sure not to add it to the `triggerContext`.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/additionalContextProvider.ts
```diff
@@ -401,36 +402,81 @@ export class AdditionalContextProvider {
...
+                ) {
                     mergedContext.push(pc)
                 }
             }
         }
+        return mergedContext.slice(0, MAX_IMAGE_CONTEXT_COUNT)
+    }
+
+    public async appendCustomContextToTriggerContext(
+        triggerContext: any,
```
ctlai95, line 424: Use `TriggerContext` type here?

yzhangok: Fixed

PR #1820, opened 2025-07-02
fix: adding files on VSC windows properly triggers reindexing

chungjac:
## Problem / Context
- Adding, deleting, renaming files within the VSC IDE specifically on Windows does not properly trigger reindexing. All other IDE/OS combinations work
- For all IDE/OS combination except VSC on Windows, using URI.parse() works
- For VSC on Windows, using URI.parse() chops off the Windows drive letter, so need to use URI.file() to preserve it
- Discussed with the team and this is only a temporary solution until further investigation is done on how the pathing works

## Solution
- If we are specifically on VSC on Windows, we use URI.file() instead of URI.parse()
- We understand that flare is IDE agnostic. We are doing this temporary solution for a time-sensitive customer ticket, until we can go back and investigate the longer term solution

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
@@ -17,6 +17,7 @@ export const LocalProjectContextServer =
         let telemetryService: TelemetryService
 
         let localProjectContextEnabled: boolean = false
+        let VSCWindowsOverride: boolean = false
 
         lsp.addInitializer((params: InitializeParams) => {
             const workspaceFolders = workspace.getAllWorkspaceFolders() || params.workspaceFolders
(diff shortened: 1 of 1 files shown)


tsmithsz (approved):

avi-alpert:
I think the reason URI.file broke other IDEs is because we're still using URI.parse to pass in workspaceFolder paths [here](https://github.com/aws/language-servers/blob/9f1ddb327778dba6da49337b79c5fef19023b52d/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts#L229) and [here](https://github.com/aws/language-servers/blob/9f1ddb327778dba6da49337b79c5fef19023b52d/server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts#L258) so we were passing inconsistent file paths

chungjac:
Will merge now, and also look into what @avi-alpert has suggested

PR #1826, opened 2025-07-03
fix: should always trigger EDIT suggestion if triggering via acceptance

atonaamz:
## License


server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
```diff
@@ -483,18 +484,24 @@ export const CodewhispererServerFactory =
...
-                                char: triggerCharacter,
-                                previousDecision: previousDecision,
-                                cursorHistory: cursorTracker,
-                                recentEdits: recentEditTracker,
-                            })
-                            predictionTypes = [
-                                ...(autoTriggerResult.shouldTrigger ? [['COMPLETIONS']] : []),
-                                ...(editPredictionAutoTriggerResult.shouldTrigger && editsEnabled ? [['EDITS']] : []),
-                            ]
+                            if (previousDecision === 'Accept' && previousSuggestionType === SuggestionType.EDIT) {
```
floralph, line 487: question: Do you need to check editsEnabled here?

floralph (approved):

server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
Will-ShaoHua: > Do you need to check editsEnabled

should be fine cuz it's checked at L481

Will-ShaoHua: as a short term hack, could we add another field called prevoiusDecisionType `previousDecisionTimestamp` and only go this path if timestamp is less than a threshold, for example 1000ms or 2000ms

```
if (previousDecision === 'Accept' && previousSuggestionType === 'Edits' && clock.now() - previousDecisionTimestamp <= threshold) {
        predictionTypes = [['EDITS']]
} else {
      // normal flow
}
```

Will-ShaoHua: long term solution we maybe still want to have something like sticky session which requires introducing another new field in the completion API. But let's start with this quick short term for now to unblock us.

Will-ShaoHua:
aidan you may want to sync with mainline first as there are small refactor changes i did in mainline thus conflicts

PR #1827, opened 2025-07-03
feat: produce agentic chat bundle from repo

awschristou:
This change allows us to produce agentic chat artifacts directly from the repo. A new GitHub Action has been created that will build these artifacts for every commit made to the `main` and `feature/*` branches. The artifacts will be available (for a limited time) from the GitHub commit check.

These changes were developed on an Ubuntu WSL image. There are bash scripts involved, which will not run on Windows. The intention is that these are run from a linux based CI compute.

To support this change: 
- a new npm script command `ci:generate:agent-standalone` has been created in the workspace `app/aws-lsp-codewhisperer-runtimes`. This is the command that CI compute will run to produce the artifacts.
- native binary dependencies (ripgrep, workspace indexer) have been directly added to the repo using [git LFS](https://git-lfs.com/). While not ideal, this gives us a faster path to quicker build iteration. In the future we can relocate these away from the repo.
- most of the scripting has been pulled from the internal repo DEXPRuntimeServerBuildConfiguration, and is mostly unchanged. This way our resulting artifacts are identical to what is currently produced internally. It also makes it easier to review these scripts, comparing them to the internal versions. Optimizations can be made to the scripts over time if desired. Here is where each script came from:
  - download-node.sh: update-node-assets.sh
  - package.sh: package.sh
  - copy-node-assets.ts: copy-node-assets.ts
  - webpack.config.prod.js: webpack.config.js

Upcoming changes will make use of this one in order to make agentic chat builds easier to obtain.

## License


app/aws-lsp-codewhisperer-runtimes/_bundle-assets/qserver-darwin-arm64.zip
leigaol, line 1: What is the size limit of Github LFS? Can it suppport 10GB? 

At development time, do we need to pull these?

app/aws-lsp-codewhisperer-runtimes/scripts/download-node.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
```
leigaol, line 1: since we are using git lfs anyway, maybe we can also check in node binary?

app/aws-lsp-codewhisperer-runtimes/_bundle-assets/qserver-darwin-arm64.zip
awschristou: The largest file is 2gb, but I don't know the cumulative space we're allowed to take up. Ideally we would try to reduce our reliance on LFS over time, but for now the goal is the ability to quickly produce builds.

We do not need these at development time, the intent is that the changes I'm adding are only used by CI jobs.

PR #1832, opened 2025-07-04
chore: create GitHub releases for agentic chat alpha, feature, and rc builds

awschristou:
This builds on #1827, to automatically produce GitHub releases containing agentic chat builds. The GitHub releases have fixed names, and are updated with every commit that is made to `main`, `feature/*`, and `release/agentic/*` branches (support for the release candidate branch support will be built out in an upcoming change).

| Commits made to branch... | Update a release called... | Note |
| ------------- | ------------- | ------------- |
| `main`  | `agentic-alpha` | Parallels the current "alpha" builds produced internally |
| `feature/foo` | `agentic-pre-foo` | |
| `release/agentic/1.2.3` | `agentic-rc-1.2.3` | Parallels the current "preprod" builds produced internally |

This is inspired by the AWS VS Code extension, which [produces releases for each commit to main and feature branches](https://github.com/aws/aws-toolkit-vscode/blob/7b161da2610ccde33639e67af4a763819037f9e5/.github/workflows/release.yml).

A manifest is also uploaded to the GitHub release, which can be consumed by plugins, to directly download the builds from that release. The manifests will not contain cumulative links to older versions, instead they will have a pointer to the version that is hosted by that GitHub release. This allows for quick and easy testing against these alpha/preprod(rc) builds.

To support this change:
- a new npm script command `ci:generate:manifest` has been created in the workspace `app/aws-lsp-codewhisperer-runtimes`. This is the command that CI compute runs to produce the artifact manifest.
- the typescript code (create-repo-manifest.ts) to produce the manifest has been sourced from the internal repo DEXPDistributionCDK (update-manifest.ts) with moderate alterations. This avoided building the script from scratch.
- the agentic chat server's version is now driven from this repo (app/aws-lsp-codewhisperer-runtimes/src/version.json). This file gives us a clean way to query for the version, which will help upcoming infrastructure changes, as well as in this change.
  - this value drives the version assigned to release candidate builds and branches (that work is coming up soon)
  - this value is emitted in the GitHub release descriptions
  - internal builds continue to assign the artifact from internal scripting. When that internal infrastructure is updated to start consuming these artifacts, the internal scripting will query from this version.json file, which will make it the source of truth.

This change was made side-by-side to #1827, and may even be merged to the repo first. It will be dormant until #1827 is merged, because the GitHub action in this change only runs after the successful completion of the GitHub action added by #1827.

## License


.github/workflows/create-agentic-github-prerelease.yml
```diff
@@ -0,0 +1,164 @@
...
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.event.workflow_run.head_sha }}
+
+            - name: Setup Node.js
+              uses: actions/setup-node@v4
+              with:
+                  node-version: '20'
```
leigaol, line 104: we are delivering node 18 to customers, should we use node 18 instead?

app/aws-lsp-codewhisperer-runtimes/scripts/create-repo-manifest.ts
```diff
@@ -0,0 +1,346 @@
...
+ * @param {string} params.gitHubReleaseName - the name of the GitHub release this will refer to (pre-main)
+ *
+ * @description
+ * This function performs the following operations:
+ * 1. Calculates SHA384 checksums and file sizes for clients.zip and servers.zip files.
+ * 2. Generates a new manifest entry with the provided and calculated information.
+ * 3. Produces a manifest file, containing only this one version.
+ * 4. Saves the updated manifest to a file.
+ */
+export async function updateManifest(
```
leigaol, line 78: if anything inside this function fails, does github CI show as aborted?

.github/workflows/create-agentic-github-prerelease.yml
awschristou: Internally we're using node 20 to produce the artifacts, so I kept this the same. I have node 18 installed on my system, and I noticed that the webpack bundling produces warnings because it is expecting node 20, so I figured I'd try to keep things consistent with the existing build system.

app/aws-lsp-codewhisperer-runtimes/src/agent-standalone.ts
```diff
@@ -18,13 +18,11 @@ import {
...
-const MAJOR = 0
-const MINOR = 1
-const PATCH = 0
-const VERSION = `${MAJOR}.${MINOR}.${PATCH}`
+const versionJson = require('./version.json')
+const version = versionJson.agenticChat
 
 const props = {
-    version: VERSION,
+    version: version,
```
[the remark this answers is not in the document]
awschristou: That is correct -- this will be a revised part of the preflight steps.
Specifically, running the action in #1842 will apply the new version.

app/aws-lsp-codewhisperer-runtimes/scripts/create-repo-manifest.ts
awschristou: The GitHub action will show as aborted on https://github.com/aws/language-servers/actions, however it will not be surfaced as an individual commit check.

leigaol (approved):

PR #1834, opened 2025-07-07
fix: imagecontext image name bug, mutliple images in pinned context

LokeshDogga13:
## Problem

During integration of Flare changes with Visual Studio, following bugs have been identified
- On windows, when we add an image using the options @image or @image: options, full file name is shown, we should only show file name.
- I was able to add only one image when I used @image context menu option, when I add another image, it does not show up in pinned context.
Images for the these bugs are below.

image

## Solution
- After fixing these issues.
image

## License


yzhangok (commented):
Can you please ensure that there's not regression on MacOS?

PR #1847, opened 2025-07-08
feat(chat-client): add stringOverrides to createChat config

avi-alpert:
## Problem
Clients need a way to override Mynah-UI strings. For example, a "workspace" in one IDE might be called a "project" in another IDE. Or when displaying keyboard shortcut hints, an IDE might want to display `Alt Enter` instead of `⌥ Enter`.

## Solution
Add an optional `stringOverrides` property to ChatClientConfig. This allows clients to override Mynah-UI [ConfigTexts](https://github.com/aws/mynah-ui/blob/aabb08d924a27ba19a0db3b01c3e3e371e7f6226/src/static.ts#L726) by passing in strings to createChat.

## License

### DIFF
--- a/chat-client/src/client/chat.ts
+++ b/chat-client/src/client/chat.ts
@@ -108,7 +108,7 @@ import {
     OPEN_FILE_DIALOG_METHOD,
     OpenFileDialogResult,
 } from '@aws/language-server-runtimes-types'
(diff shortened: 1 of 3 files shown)


awschristou (approved):
Is it possible to add a test for the change?

shruti0085 (approved):

PR #1868, opened 2025-07-10
test: add integration test workflow

ctlai95:
## Problem

We don't have any integration tests to validate built servers.

## Solution

Create a new workflow to run integration tests. The workflow job will run after successful agent-standalone builds and do the following steps for all supported target platforms:
1. Download the servers.zip for the current platform
2. Refresh/Fetch a new SSO token for the agent server test
3. Runs the integration test which will start the lsp server (with a simple hello chat prompt test for now)

## License


server/aws-lsp-codewhisperer/src/shared/amazonQServiceManager/configurationUtils.ts
```diff
@@ -108,6 +108,14 @@ export async function getAmazonQRelatedWorkspaceConfigs(
     lsp: Lsp,
     logging: Logging
 ): Promise<Readonly<Partial<AmazonQWorkspaceConfig>>> {
+    const clientParams = lsp.getClientInitializeParams()
+    const supportsWorkspaceConfiguration = clientParams?.capabilities?.workspace?.configuration !== false
+
+    if (!supportsWorkspaceConfiguration) {
+        logging.debug('Client does not support workspace configuration, returning default config.')
+        return {}
+    }
```
ctlai95, line 117: This check was added because the server always sends `workspace/configuration` requests to the client but `ts-lsp-client` doesn't support it which results in failures due to out-of-order messages.

Real clients should use `true` or `undefined` for `workspace.configuration` in client capabilities.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+                  path: app/aws-lsp-codewhisperer-runtimes/build/archives/agent-standalone/${{ matrix.target }}/
+            - name: Configure AWS Credentials
+              uses: aws-actions/configure-aws-credentials@v4
+              with:
+                  role-to-assume: arn:aws:iam::964765661569:role/GitHubActionsTokenRefresherRole
+                  role-session-name: language-servers-github
+                  aws-region: us-east-1
+            - name: Build
+              run: |
+                  npm ci
```
leigaol, line 56: ideally we want the build step to use the exact artifact from https://github.com/aws/language-servers/pull/1827.

leigaol: Is this PR depending on #1827?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+              uses: actions/setup-node@v4
+              with:
+                  node-version: 18
+            - name: Download build artifacts
+              uses: actions/download-artifact@v4
+              with:
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  run-id: ${{ github.event.workflow_run.id }}
+                  name: ${{ matrix.target }}
+                  path: app/aws-lsp-codewhisperer-runtimes/build/archives/agent-standalone/${{ matrix.target }}/
```
leigaol, line 47: this seems like it depends on https://github.com/aws/language-servers/pull/1827/files#diff-989018ac7fbcc4a7edb5805c9b80cc242890442e00c8ec8d9033bbca6b4602dcR41

.github/workflows/integration-tests.yml
ctlai95: Yes this depends on #1827

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+            - name: Configure AWS Credentials
+              uses: aws-actions/configure-aws-credentials@v4
+              with:
+                  role-to-assume: arn:aws:iam::964765661569:role/GitHubActionsTokenRefresherRole
+                  role-session-name: language-servers-github
+                  aws-region: us-east-1
+            - name: Build
+              run: |
+                  npm ci
+                  npm run compile
```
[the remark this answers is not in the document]
ctlai95: This is actually needed to compile the servers, not the `app/aws-lsp-codewhisperer-runtimes` workspace. But we can probably do `npm run compile:core && npm run compile:servers` instead here

app/aws-lsp-codewhisperer-runtimes/src/tests/testFixture/test.py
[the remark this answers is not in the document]
ctlai95: These were copy/pasted from a different integ test but they are currently not used. I'll remove them

integration-tests/q-agentic-chat-server/src/tests/testFixture/test.py
ctlai95, line 1: Planning to keep these files for now since they are acting as the workspace files for the test

app/aws-lsp-codewhisperer-runtimes/src/tests/testFixture/test.py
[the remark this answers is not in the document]
ctlai95: Actually I'm planning to keep these for now, since they are acting as the workspace files for the tests

awschristou (approved):

PR #1892, opened 2025-07-14
feat: add active user tracking with state persistence

chungjac:
## Problem
- We need active user data for CW

## Solution
**Implemented window-based active user tracking with state persistence:**
1. When a user sends a message and no window exists yet, a new window is created and a telemetry event is emitted
2. If a user sends additional messages within that same window, no new telemetry event will be emitted
3. The window is NOT reset when additional messages are sent within it
4. Once the window expires, the next message will create a new window and emit a telemetry event

**Persisting state:**
1. Window state is persisted to disk in ~/.aws/amazonq/active-user-state.json
2. State is loaded on initialization based on clientId to maintain window continuity across IDE restarts
3. Multiple client IDs are supported

**Notes:**
- Active user is defined as one who has sent a message within the last 24 hours
- This telemetry is just for kibana. More work will need to be done to post data to CW

## Verification 
- Kibana metric gets emitted only once per window
- Window state persists across IDE restarts
- Unit tests

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -194,7 +194,6 @@ import { ActiveUserTracker } from '../../shared/activeUserTracker'
 import { UserContext } from '../../client/token/codewhispererbearertokenclient'
 import { CodeWhispererServiceToken } from '../../shared/codeWhispererService'
 
-
 type ChatHandlers = Omit<
     LspHandlers<Chat>,
     | 'openTab'


yueny2020:
if user logs out, then the time window will reset, right? 
if yes, then we will have more active user than the actual number, right?

chungjac:
> if user logs out, then the time window will reset, right?
if yes, then we will have more active user than the actual number, right?

Correct, it would emit again

server/aws-lsp-codewhisperer/src/shared/activeUserTracker.ts
```diff
@@ -0,0 +1,191 @@
...
+     * Gets the client ID from the LSP initialization parameters or generates a machine ID if not available
+     * @returns The client ID
+     */
+    private getClientId(): string {
+        const clientId = this.features.lsp.getClientInitializeParams()?.initializationOptions?.aws?.clientInfo?.clientId
+        if (clientId) {
+            return clientId
+        }
+
+        // Generate a machine-specific ID if no client ID is available
```
yueny2020, line 109: should we create a clientId or throw exception?

yueny2020 (approved):

server/aws-lsp-codewhisperer/src/shared/activeUserTracker.ts
chungjac: resolved with @yueny2020 offline

PR #1894, opened 2025-07-14
fix: delete empty history db file when starting trimming history

jngyuamz:
## Problem
- If there are empty history db files, the IDE plugins don't have a mechanism to delete them.

## Solution
- In history trimming loop, we load all non-empty db files but ignore empty db files. This PR added a logic to delete the empty files instead of ignoring them.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatHistoryMaintainer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatHistoryMaintainer.ts
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import * as Loki from 'lokijs'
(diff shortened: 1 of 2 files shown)


avi-alpert:
Would this delete files that have a Settings collection but no Tab collection?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatHistoryMaintainer.ts
```diff
@@ -319,7 +341,9 @@ export class ChatHistoryMaintainer {
                 if (!this.isEmptyCollection(collection)) {
                     allDbsMap.set(dbFile, { collection: collection, db: db })
                 } else {
```
avi-alpert, line 343: Would this delete files that have a Settings collection but no Tab collection?

PR #1904, opened 2025-07-16
fix(amazonq): change to use promptStickyCard to show image verificati…

yzhangok:
…on notification

## Problem
all errors are displayed near the top of the chat view and it's very easy to miss.

Screenshot 2025-07-15 at 8 47 54 PM

## Solution

change to use promptStickyCard which is closer to prompt input to show

Screenshot 2025-07-15 at 8 48 53 PM

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -777,13 +777,21 @@ export const createMynahUi = (
                 // Add valid files to context commands
                 mynahUi.addCustomContextToPrompt(tabId, commands, insertPosition)
             }
-            const uniqueErrors = [...new Set(errors)]
-            for (const error of uniqueErrors) {
-                mynahUi.notify({
-                    content: error,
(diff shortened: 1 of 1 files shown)


chat-client/src/client/mynahUi.ts
```diff
@@ -777,13 +777,21 @@ export const createMynahUi = (
...
-                })
+
+            const imageVerificationBanner: Partial<ChatItem> = {
+                messageId: 'image-verification-banner',
+                header: {
+                    icon: 'warning',
+                    iconStatus: 'warning',
+                    body: '### Invalid Image',
+                },
+                body: `${errors.join('\n')}`,
```
leigaol, line 788: dedup. the errors here?

yzhangok: The error message is FileName + Violated Restriction (i.e. `diagram.png: Image must be no more than 3.75MB in size.`). If there's conflict in the error message we still should display them

PR #1916, opened 2025-07-17
fix: fix blocking regex calls being made before indexing

manodnyab:
## Problem
In large projects, chat gets stuck in Thinking stage for a long time. Similarly the @ Pin Context button is not shown.
## Solution
There is a delay in the server getting the onChatPrompt command that starts chat. This delay is because regex events being used to filter files to send over to the indexing library were blocking the IO thread. Replacing the ignoreWalk library with fast glob to remove the regex events. 

## License


server/aws-lsp-codewhisperer/src/shared/utils.ts
```diff
@@ -24,7 +24,9 @@ import {
 } from '@amzn/codewhisperer-streaming'
 import * as path from 'path'
 import { ServiceException } from '@smithy/smithy-client'
-import * as ignoreWalk from 'ignore-walk'
```
avi-alpert, line 27: could we now uninstall it from https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/package.json#L85

server/aws-lsp-codewhisperer/src/shared/utils.ts
```diff
@@ -506,36 +508,53 @@ export function hasConnectionExpired(error: any) {
...
     try {
-        let ignoreWalkFiles = await ignoreWalk({
-            path: directory,
-            // Use multiple ignore files
-            ignoreFiles: ['.gitignore', '.npmignore'],
-            // Additional options
-            follow: false, // follow symlinks
-            includeEmpty: false, // exclude empty directories
-        })
-        // hard limit of 500k files to avoid hitting memory limit
```
avi-alpert, line 526: why did we get rid of this 500k limit

manodnyab: Added it back

avi-alpert (approved):

rli (approved):

PR #1933, opened 2025-07-18
fix(amazonq): fix for mcp server unnecessary refresh from file watchers

ashishrp-aws:
## Problem
File Watchers are triggering mcp server refresh during permission changes and random clicks. Permission page close results into mcp list re-opening and refresh

## Solution
- Added debounce for file watchers and programmatic change tracking update to correctly track them.
- permission page close is also marked as mcp servers list page as not active.

## License

### DIFF
--- a/chat-client/src/client/mcpMynahUi.ts
+++ b/chat-client/src/client/mcpMynahUi.ts
@@ -508,6 +508,7 @@ export class McpMynahUi {
                         },
                         onClose: () => {
                             this.messager.onMcpServerClick(MCP_IDS.SAVE_PERMISSION_CHANGE)
+                            this.isMcpServersListActive = false
                         },
                         onBackClick: () => {
                             this.messager.onMcpServerClick(MCP_IDS.SAVE_PERMISSION_CHANGE)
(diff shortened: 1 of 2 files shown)


chat-client/src/client/mcpMynahUi.ts
```diff
@@ -508,6 +508,7 @@ export class McpMynahUi {
                         },
                         onClose: () => {
                             this.messager.onMcpServerClick(MCP_IDS.SAVE_PERMISSION_CHANGE)
+                            this.isMcpServersListActive = false
```
bywang56, line 511: do we need to do this for other action as well like onBackCLick below?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
```diff
@@ -867,23 +867,25 @@ export class McpEventHandler {
 
         try {
             await McpManager.instance.removeServer(serverName)
+
+            return { id: params.id }
         } catch (error) {
             this.#features.logging.error(`Failed to delete MCP server: ${error}`)
-        } finally {
-            // Reset flag after operations
             this.#isProgrammaticChange = false
```
bywang56, line 874: why putting reset flag and return in catch block here and outside of the catch in other places? is this intended?

chat-client/src/client/mcpMynahUi.ts
ashishrp-aws: yes would need it...other wise. It'd prompt mcp list to open

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
ashishrp-aws: intended. we can't have finally block. we are resetting it somewhere because it was not working out

bywang56 (approved):

PR #1944, opened 2025-07-21
fix: bug for credential refresh in StreamingClientServiceIAM

sdharani91:
Added expiration time to credential provider so SDK will refresh creds automatically.

## License


sdharani91:
Tested by passing credentials with different expiry times to the language server and validating it uses the expected credentials through logs.
Will be  further testing in SMUS environment.

mr-lee (approved):

PR #1946, opened 2025-07-22
fix(amazonq): avoid workspace context server missing historical dependency events

LiGaCu:
## Problem

There are two issues with workspace context server receiving `didChangeDependencyPaths` events:
- **Lost initial events:** Before the workspace context server properly determines user eligibility (`isUserEligibleForWorkspaceContext()`) and completes initialization, all `didChangeDependencyPaths` events sent from the IDE extension are lost.
- **Lost historical events:** When `dependencyDiscoverer` is disposed (due to user opt-out, admin opt-out or user logout), all previously received `didChangeDependencyPaths` events are lost as dependency maps are reset. These events cannot be recovered if the workspace context server needs to run for the user again later.

## Solution

Implement an in-memory recording system for all didChangeDependencyPaths events that:
- Stores events regardless of workspace context server initialization status
- Ingests these recorded events when building initial dependency maps
- Only processes events according to the user's current qualification status

## Testing

Tested the changes locally with JetBrains Amazon Q extension and verified we are no longer missing ``didChangeDependencyPaths`` events sent at the start of IDE. Also verified the original dependency discovery and upload logic is not disturbed by this change.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyDiscoverer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyDiscoverer.ts
@@ -10,6 +10,7 @@ import {
 } from './dependencyHandler/LanguageDependencyHandler'
 import { ArtifactManager } from '../artifactManager'
 import { supportedWorkspaceContextLanguages } from '../../../shared/languageDetection'
+import { DependencyEventBundler } from './dependencyEventBundler'
 
 export class DependencyDiscoverer {
     private logging: Logging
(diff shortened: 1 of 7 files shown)


Will-ShaoHua (approved):

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/dependency/dependencyDiscoverer.ts
```diff
@@ -144,12 +149,45 @@ export class DependencyDiscoverer {
...
+                    if (!recordedPaths) {
+                        continue
+                    }
+                    dependencyHandler.updateDependencyMapBasedOnLSP(recordedPaths, workspaceFolder)
+                    ingestedDependencyCount += recordedPaths.length
+                } catch (error) {
+                    this.logging.debug(`Error ingesting dependency events for ${workspaceFolder.uri}: ${error}`)
+                }
+            }
+            this.dependencyEventsIngestedFolderUris.add(workspaceFolder.uri)
```
leigaol, line 174: this is marking workspaceFolder.uri as ingested even if some of the dependencies are failed or thrown errors.

PR #1952, opened 2025-07-22
test(amazonq): add integ tests for built-in tools

ctlai95:
## Problem

Basic tool execution is not tested

## Solution

- Replace `ts-lsp-client` with a custom jsonrpc endpoint implementation that supports asynchronous out-of-order messages
- Log errors without needing to set `DEBUG=true` for easier debugging of test failures
- Add integration tests for built-in tools: `fsRead`, `listDirectory`, `executeBash`, `fsWrite`, `fsReplace`, and `fileSearch`

## License

### DIFF
--- a/integration-tests/q-agentic-chat-server/src/tests/agenticChatInteg.test.ts
+++ b/integration-tests/q-agentic-chat-server/src/tests/agenticChatInteg.test.ts
@@ -4,12 +4,13 @@ import * as chaiAsPromised from 'chai-as-promised'
 import { ChildProcessWithoutNullStreams, spawn } from 'child_process'
 import { describe } from 'node:test'
 import * as path from 'path'
-import { JSONRPCEndpoint, LspClient } from 'ts-lsp-client'
+import { JSONRPCEndpoint, LspClient } from './lspClient'
 import { pathToFileURL } from 'url'
 import * as crypto from 'crypto'
 import { EncryptionInitialization } from '@aws/lsp-core'
 import { authenticateServer, decryptObjectWithKey, encryptObjectWithKey } from './testUtils'
 import { ChatParams, ChatResult } from '@aws/language-server-runtimes/protocol'
+import * as fs from 'fs'
 
 chai.use(chaiAsPromised)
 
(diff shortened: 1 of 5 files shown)


integration-tests/q-agentic-chat-server/src/tests/agenticChatInteg.test.ts
```diff
@@ -113,20 +125,248 @@ describe('Q Agentic Chat Server Integration Tests', async () => {
...
         expect(decryptedResult).to.have.property('messageId')
         expect(decryptedResult).to.have.property('body')
         expect(decryptedResult.body).to.not.be.empty
     })
+
+    it('reads file contents using fsRead tool', async () => {
+        const encryptedMessage = await encryptObjectWithKey<ChatParams>(
+            {
+                tabId,
+                prompt: { prompt: 'Read the contents of the test.py file using the fsRead tool.' },
```
jguoamz, line 163: does test.py exist?

ctlai95: yes it exists in testFixtures folder with some dummy code

integration-tests/q-agentic-chat-server/src/tests/lspClient.ts
```diff
@@ -0,0 +1,126 @@
...
+            },
+            () => this.nextId++
+        )
+
+        let buffer = ''
+        readable.on('data', (chunk: Buffer) => {
+            buffer += chunk.toString()
+
+            while (true) {
+                const headerEnd = buffer.indexOf('\r\n\r\n')
```
leigaol, line 56: why is it \r\n? Is this test supposed to run in Windows only?

jguoamz (approved):
nit: we don't have auto retries in flare which means test here can fail, do we have retry on the test level?

integration-tests/q-agentic-chat-server/src/tests/lspClient.ts
ctlai95: It's a language server specification requirement: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#baseProtocol

Not required to be run on windows

ctlai95:
> nit: we don't have auto retries in flare which means test here can fail, do we have retry on the test level?

Yes, we have `--retries 2` flag passed to the mocha process, see https://github.com/aws/language-servers/blob/main/integration-tests/q-agentic-chat-server/package.json#L9

PR #1954, opened 2025-07-22
fix: emit metric for tool error

chungjac:
## Problem
- missing metric for tool error

## Solution
- emit metric when there is a tool error
    - exclude tool errors relating to user rejecting/canceling the shell command, as this is intended by the user

## License


laileni-aws (approved):
LGTM

PR #1955, opened 2025-07-22
feat(amazonq): enhance workspaceContext classpath generation

liumofei-amazon:
## Problem

- In previous setup, when user open up workspace that's not at java root folder, we won't able to generating `.classpath` and `.project` correctly because there's no /src /tst folder available. For exmaple if I have `workspace/src/package1` and `workspace/src/package2`. When user open the folder at `workspace/src` flare will generating `.classpath` and `.project` under `src` with no source folder

-  Introducing this change to identify the actual project root folder by looking at project files like `build.xml, gradle.build, build.gradle.kts, pom.xml`  Loop through those directory to generate `.classpath` and `.project` for each folder.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
@@ -6,7 +6,11 @@ import JSZip = require('jszip')
 import { EclipseConfigGenerator, JavaProjectAnalyzer } from './javaManager'
 import { resolveSymlink, isDirectory, isEmptyDirectory } from './util'
 import glob = require('fast-glob')
-import { CodewhispererLanguage, getCodeWhispererLanguageIdFromPath } from '../../shared/languageDetection'
+import {
+    CodewhispererLanguage,
+    getCodeWhispererLanguageIdFromPath,
+    isJavaProjectFileFromPath,
+} from '../../shared/languageDetection'
 
 export interface FileMetadata {
     filePath: string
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
```diff
@@ -633,25 +637,49 @@ export class ArtifactManager {
...
+        const projectRoots = this.extractJavaProjectRoots(files)
         const additionalFiles: FileMetadata[] = []
 
-        // Generate Eclipse configuration files
-        const javaManager = new JavaProjectAnalyzer(workspacePath)
-        const structure = await javaManager.analyze()
-        const generator = new EclipseConfigGenerator(workspaceFolder, this.logging)
+        // Process each project root separately
+        for (const projectRoot of projectRoots) {
+            const isRootProject = projectRoot === '.'
```
leigaol, line 646: why can it be '.'

liumofei-amazon: This is returned from the extraction methods above. If it's already within a root folder (with those build file) it will treat the projects roots array returned as ['.']  and `const projectPath = path.join(workspacePath, projectRoot)`  projectPath won't change still be `workspacePath`.

server/aws-lsp-codewhisperer/src/shared/languageDetection.ts
```diff
@@ -302,3 +304,12 @@ export function getCodeWhispererLanguageIdFromPath(filePath: string): Codewhispe
 
     return undefined
 }
+
+export function isJavaProjectFileFromPath(filePath: string): boolean {
```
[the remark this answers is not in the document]
liumofei-amazon: That will needs to be handled when we do support kotlin in SSPC. Right now the logic will only check there's java programming language file exist, I'll add a check to also look for extension names. 

Will add UT

server/aws-lsp-codewhisperer/src/shared/languageDetection.ts
```diff
@@ -294,6 +294,8 @@ export function getCodeWhispererLanguageIdFromPath(filePath: string): Codewhispe
         return 'javascript'
     }
 
+    if (isJavaProjectFileFromPath(filePath)) return 'java'
```
LiGaCu, line 297: Actually, instead of hard-coding these files as ``java``, could we integrate them here?
- https://github.com/aws/language-servers/blob/528f8206b101e8f0c785b7fc0aceb87d6ef3de7b/server/aws-lsp-codewhisperer/src/language-server/workspaceContext/javaManager.ts#L861C19-L861C41

server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
```diff
@@ -627,31 +631,55 @@ export class ArtifactManager {
         files: FileMetadata[]
     ): Promise<FileMetadata[]> {
         const workspacePath = URI.parse(workspaceFolder.uri).path
-        const hasJavaFiles = files.some(file => file.language === 'java')
+        const hasJavaFiles = files.some(file => file.language === 'java' && file.relativePath.endsWith('java'))
```
leigaol, line 634: how about .endsWith('.java')

server/aws-lsp-codewhisperer/src/shared/languageDetection.ts
liumofei-amazon: The problem is we need to `analyze`  to get the structure result then we can generate to write the structure into .classpath and .project. Those analyze needs to be done for each project roots identified before this.

Upload wise it maybe fine, but this is not the entrypoint to the generation

PR #1990, opened 2025-07-27
refactor: trigger code path

Will-ShaoHua:
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/codeWhispererService.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/codeWhispererService.ts
@@ -251,12 +251,39 @@ export class CodeWhispererServiceToken extends CodeWhispererServiceBase {
     }
 
     async generateSuggestions(request: GenerateSuggestionsRequest): Promise<GenerateSuggestionsResponse> {
+        const isCompletionRequest =
+            request.predictionTypes === undefined ||
+            (request.predictionTypes.length > 0 && request.predictionTypes[0] === SuggestionType.COMPLETION)
+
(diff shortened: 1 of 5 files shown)


server/aws-lsp-codewhisperer/src/language-server/inline-completion/codewhispererController.ts
```diff
@@ -0,0 +1,128 @@
...
+    editCompletionHandler: EditCompletionHandler | undefined
+    private inlineCompletionHandler: InlineCompletionHandler | undefined
+    private logInlineCompletionSessionResultsHandler: LogInlineCompletionSessionResultsHandler | undefined
+
+    constructor(
+        readonly sessionManager: SessionManager,
+        readonly editsSessionManager: SessionManager
+    ) {}
+
+    async init(
```
leigaol, line 37: can you move this init to the constructor so that `inlineCompletionHandler` does not need to be undefined?

server/aws-lsp-codewhisperer/src/shared/codeWhispererService.ts
```diff
@@ -181,6 +270,8 @@ export class CodeWhispererServiceToken extends CodeWhispererServiceBase {
     /** If user clicks "Upgrade" multiple times, cancel the previous wait-promise. */
     #waitUntilSubscriptionCancelSource?: CancellationTokenSource
 
+    private cache: Map<string, GenerateSuggestionsResponse> = new Map()
```
leigaol, line 273: this is not used

PR #2003, opened 2025-07-29
fix: use fast glob streaming when collecting files

leigaol:
## Problem

fast glob on a extremely large workspace (870GB, 2.4m files) is still causing issues that blocks the node js event loop.

## Solution

1. Use fast glob streaming method that early terminates without doing actual scan of entire workspace. Previously fg won't stop until it finish scanning entire workspace.

2. Lower the threshold of workspace files to 100k

## License


server/aws-lsp-codewhisperer/src/shared/utils.ts
```diff
@@ -533,15 +533,28 @@ export async function listFilesWithGitignore(directory: string): Promise<string[
...
+    const absolutePaths: string[] = []
+    let fileCount = 0
+    const MAX_FILES = 100_000
+
+    const stream = fg.stream(['**/*'], {
         cwd: directory,
         dot: true,
         ignore: ignorePatterns,
-        onlyFiles: false,
+        onlyFiles: true,
```
manodnyab, line 544: We need folders here too, right?

server/aws-lsp-codewhisperer/src/shared/utils.ts
```diff
@@ -533,15 +533,28 @@ export async function listFilesWithGitignore(directory: string): Promise<string[
...
         followSymbolicLinks: false,
         absolute: true,
     })
-    return absolutePaths.slice(0, 500_000)
+
+    for await (const entry of stream) {
```
manodnyab, line 549: Do we need the for loop, can we just change the value in the slice above to 100K?

leigaol: No. This is for early termination of fast glob. Without for loop it iterate first.

server/aws-lsp-codewhisperer/src/shared/utils.ts
leigaol: No it should only be files

XiaoxuanLu (approved):

PR #2017, opened 2025-07-31
fix(amazonq): fix processing empty unsupported workspace file

liumofei-amazon:
## Problem

From the priors change workspace context is not handling `.scala` `.kt` project correctly by uploading an empty project file. 

This PR to address this issue, only adding the `pom.xml` `gradle.build.kts` `gradle.build` `build.xml` if there's java source file existing. 

## Solution

- Move the logic assign the language override inside `artifactManager`, only process these projectFile if there's `.java` source file exist so we do not upload it for kotlin/scala/smithy projects which will result an empty zip with project file only

- Also introduce the fix using glob.stream and have upper limit of file count from this [commit](
https://github.com/liumofei-amazon/language-servers/commit/f7c0a0b0ef9ce3ecd620acfef00e55745db3d71f)

## License


server/aws-lsp-codewhisperer/src/language-server/workspaceContext/artifactManager.ts
```diff
@@ -368,11 +368,19 @@ export class ArtifactManager {
             onlyFiles: true,
         })
 
+        const hasJavaFile = files.some(file => file.endsWith('.java'))
```
leigaol, line 371: in this function when you get files, please use fast glob streaming methods. See https://github.com/aws/language-servers/pull/2003

liumofei-amazon: Ack let me refactor this to streaming methods

liumofei-amazon: @leigaol  added the same glob.stream with early exist on MAX_FILES

leigaol (approved):

PR #2026, opened 2025-08-01
fix(amazonq): fix for mcp server permissions

ashishrp-aws:
## Problem
MCP server permissions save has seen regression due to recent changes.

## Solution
fixed the permission save for mcp server tools and its telemetry

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
@@ -1174,11 +1174,6 @@ export class McpEventHandler {
      * Applies the stored permission changes
      */
     async #handleSavePermissionChange(params: McpServerClickParams) {
-        if (!params.optionsValues) {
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpEventHandler.ts
```diff
@@ -1174,11 +1174,6 @@ export class McpEventHandler {
      * Applies the stored permission changes
      */
     async #handleSavePermissionChange(params: McpServerClickParams) {
-        if (!params.optionsValues) {
```
tsmithsz, line 1177: Not familiar with the entire MCP flow. Why is there no longer a need for this initial validation check?

ashishrp-aws: optionsValues are actually not present in save mcp params, so it is returning without saving permissions. It was added couple of days ago by @bywang56 and dicussed with him offline. Corrected this and corresponding telemetry.

PR #2039, opened 2025-08-01
fix: add bash command parsing for telemetry metrics

chungjac:
## Problem
We currently lack visibility into which specific commands are being executed through the executeBash tool in our client-side 
metrics

**Important**: We cannot emit the full command strings as they may contain personal information (file paths, usernames, sensitive data, etc.), so we need to extract only the base command names for telemetry.

## Solution
We now extract base commands from complex shell command strings, handling various scenarios including:
- Simple commands: cd /home/user → ["cd"]
- Command chaining: echo "hello" && ls -la → ["echo", "ls"]
- Pipes: cat file.txt | grep pattern → ["cat", "grep"]
- Prefix commands: sudo apt-get install → ["sudo", "apt-get"]
- Path prefixes: /usr/bin/python script.py → ["python"]
- Mixed separators and complex command chains

**Note**: This implementation handles MOST common use cases. While there are more edge cases that could be handled, it covers the majority of real-world scenarios that users will encounter. The goal is to provide useful telemetry data, not to be a 
perfect shell parser.

**Telemetry Approach**: Instead of emitting telemetry at the existing emitInteractWithAgenticChat method level (which would only give us one event per bash execution), we emit telemetry at the most granular level - one event per individual base command. This approach will enable detailed Kibana visualizations showing individual command usage frequency.

For example, `echo "hello" && ls -la && grep pattern file.txt` will generate 3 separate telemetry events: echo, ls, and grep.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
@@ -209,6 +209,7 @@ import { McpEventHandler } from './tools/mcp/mcpEventHandler'
 import { enabledMCP, createNamespacedToolName } from './tools/mcp/mcpUtils'
 import { McpManager } from './tools/mcp/mcpManager'
 import { McpTool } from './tools/mcp/mcpTool'
+import { parseBaseCommands } from './utils/commandParser'
 import {
     freeTierLimitUserMsg,
     onPaidTierLearnMore,
@@ -1755,6 +1756,16 @@ export class AgenticChatController implements ChatHandlers {
                                 )
                             }
                             if (isExecuteBash) {
+                                const fullCommand = (toolUse.input as unknown as ExecuteBashParams).command
+
+                                // Extract just the base commands without arguments or options
+                                const baseCommands = parseBaseCommands(fullCommand)
+
+                                // Emit each seperately for visualization purposes
+                                baseCommands.forEach(command => {
+                                    this.#telemetryController.emitBashCommand(tabId, command)
+                                })
+
                                 this.#telemetryController.emitInteractWithAgenticChat(
                                     'RunCommand',
                                     tabId,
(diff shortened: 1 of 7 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -1751,6 +1752,16 @@ export class AgenticChatController implements ChatHandlers {
                                 )
                             }
                             if (isExecuteBash) {
+                                const fullCommand = (toolUse.input as unknown as ExecuteBashParams).command
```
tsmithsz, line 1755: Could we have integrated this in the `executeBash` tool itself?

chungjac: In terms of the executeBash tool also outputting each of the base commands? If so, this would require modifying tool spec to add this other field, and not sure if it would affect performance of tool
 
I feel that the executeBash tool can focus solely on command execution while controller can handle the post processing and telemetry logic

tsmithsz: We already do some splitting in the `executeBash` tool itself

chungjac: But i need to see if it makes sense to have to pass the telemetryController

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/executeBash.ts
```diff
@@ -146,6 +154,20 @@ export class ExecuteBash {
...
         }
 
+        // Extract individual base commands for telemetry purposes
+        const baseCommands = parseBaseCommands(args)
+        baseCommands.forEach(command => {
+            const metricPayload = {
+                name: ChatTelemetryEventName.BashCommand,
+                data: {
+                    credentialStartUrl: this.credentialsProvider.getConnectionMetadata()?.sso?.startUrl,
+                    result: 'Succeeded',
```
tsmithsz, line 164: Shouldn't we emit this in the invoke method? If validation fails, telemetry will still be emitted for commands that never executed.

tsmithsz (approved):

PR #2046, opened 2025-08-04
fix: Enable Test Flag For AmazonQ UI Testing

laura-codess:
## Problem

Currently, the Mynah-UI elements do not show `data-testid` at the production level of Amazon Q. This makes UI Tests very flakey because we have to rely on CSS Selectors to grab Mynah-UI elements, and these CSS Selectors can be shared across multiple different elements.

## Solution

In `mynahUI.ts` we set `test:true` so `data-testid` will show at production level. Now instead of using CSS Selectors, test writers can grab the `data-testid`

```
const mynahUiProps: MynahUIProps = {
        ...
        config: {
            maxTabs: 10,
            test: true,
        ...
```

Screenshot 2025-08-04 at 2 12 59 PM <br/>

_This is the code in Mynah-UI that turns on test-ids when test config is set to `true`_
https://github.com/aws/mynah-ui/blob/main/src/helper/dom.ts#L264

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -810,6 +810,7 @@ export const createMynahUi = (
         },
         config: {
             maxTabs: 10,
+            test: true,
             dragOverlayIcon: MynahIcons.IMAGE,
             texts: {
                 ...uiComponentsTexts,


chat-client/src/client/mynahUi.ts
```diff
@@ -810,6 +810,7 @@ export const createMynahUi = (
         },
         config: {
             maxTabs: 10,
+            test: true,
```
rli, line 813: should this only be in test mode or is it fine to always have on

laura-codess: I believe it's fine to always have on. The mynah people that I spoke with agreed that setting test = true doesn't do anything aside from adding data-testids.

PR #2058, opened 2025-08-05
refactor: split edit from completion code path

Will-ShaoHua:
## Problem
parent PR #1990

As a new feature introduced recently, "Edits" is the new inline suggestion form however it has few different requirement compared to regular "Completions".
- "Completion" is executed should be executed upfront and in a blocking manner so that requests coming later will not be executed and should return empty result. 
- Whereas "Edits" should be executed in a debounce manner such that the file context we collect is up-to-date and reflect the latest users' editor state.

## Solution

- Separate out "Edits" and "Completions" code path

## License


server/aws-lsp-codewhisperer/src/language-server/inline-completion/session/sessionManager.ts
```diff
@@ -287,17 +288,17 @@ export class SessionManager {
...
+            return (SessionManager._editInstance ??= new SessionManager())
         }
 
-        return SessionManager._instance
+        return (SessionManager._completionInstance ??= new SessionManager())
     }
 
     // For unit tests
     public static reset() {
-        SessionManager._instance = undefined
```
leigaol, line 300: this should reset both of its instances _completionInstance, _editInstance

Will-ShaoHua: good callout, i miss this

server/aws-lsp-codewhisperer/src/language-server/inline-completion/auto-trigger/autoTrigger.ts
```diff
@@ -177,7 +177,7 @@ type AutoTriggerParams = {
     char: string
     triggerType: string // Left as String intentionally to support future and unknown trigger types
     os: string
-    previousDecision: string
+    previousDecision: string | undefined
```
Will-ShaoHua, line 180: will revert this for now, not needed in this pr

server/aws-lsp-codewhisperer/src/language-server/inline-completion/codewhispererController.ts
```diff
@@ -0,0 +1,67 @@
...
+import { AmazonQBaseServiceManager } from '../../shared/amazonQServiceManager/BaseAmazonQServiceManager'
+import { TelemetryService } from '../../shared/telemetry/telemetryService'
+import { DocumentChangedListener } from './documentChangedListener'
+import { SessionManager } from './session/sessionManager'
+import { RecentEditTracker } from './tracker/codeEditTracker'
+import { CursorTracker } from './tracker/cursorTracker'
+import { EditCompletionHandler } from './editCompletionHandler'
+import { RejectedEditTracker } from './tracker/rejectedEditTracker'
+
+export class CodeWhispererController {
```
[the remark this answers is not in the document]
Will-ShaoHua: done https://github.com/aws/language-servers/pull/2058/commits/880bf07e5a92019f4f5a613e28a985b73797486b

server/aws-lsp-codewhisperer/src/language-server/inline-completion/session/sessionManager.ts
Will-ShaoHua: https://github.com/aws/language-servers/pull/2058/commits/880bf07e5a92019f4f5a613e28a985b73797486b

server/aws-lsp-codewhisperer/src/language-server/inline-completion/auto-trigger/autoTrigger.ts
Will-ShaoHua: done

leigaol (approved):

server/aws-lsp-codewhisperer/src/language-server/inline-completion/trigger.ts
```diff
@@ -0,0 +1,62 @@
+import { CodewhispererLanguage } from '../../shared/languageDetection'
+import { SessionManager } from './session/sessionManager'
+import { InlineCompletionWithReferencesParams } from '@aws/language-server-runtimes/protocol'
+import { editPredictionAutoTrigger } from './auto-trigger/editPredictionAutoTrigger'
+import { CursorTracker } from './tracker/cursorTracker'
+import { RecentEditTracker } from './tracker/codeEditTracker'
+import { CodeWhispererServiceBase, CodeWhispererServiceToken } from '../../shared/codeWhispererService'
+
+export class NepTrigger {}
```
Will-ShaoHua, line 9: https://github.com/aws/language-servers/pull/1990/files#diff-564725aca4906036c2f88fcf597bbc4f338ea83087f4b50ae509e36fed624d86R18-R52

server/aws-lsp-codewhisperer/src/shared/codeWhispererService.ts
```diff
@@ -138,6 +142,23 @@ export abstract class CodeWhispererServiceBase {
 
     abstract generateSuggestions(request: GenerateSuggestionsRequest): Promise<GenerateSuggestionsResponse>
 
+    abstract constructSupplementalContext(
```
Will-ShaoHua, line 145: since it will be reused by both "inlineCompletionHandler" and "editCompletionHandler", move to here such that we don't have duplicate

Will-ShaoHua: Also it can leverage the "IAM" vs. "Token" subclass to have their own implementation. 
Currently we don't have supplemental context for IAM and only Token based has such.

server/aws-lsp-codewhisperer/src/language-server/inline-completion/session/sessionManager.ts
```diff
@@ -287,17 +288,18 @@ export class SessionManager {
     /**
      * Singleton SessionManager class
      */
-    public static getInstance(): SessionManager {
-        if (!SessionManager._instance) {
-            SessionManager._instance = new SessionManager()
+    public static getInstance(type: 'COMPLETIONS' | 'EDITS' = 'COMPLETIONS'): SessionManager {
```
Will-ShaoHua, line 291: It's to completely separate out the states of 2 functionalities. 1 for Completion and the other for Edits.

floralph (approved):

PR #2060, opened 2025-08-05
feat: send pinned context button immediately with pending state

constewart9:
## Problem
In large workspaces, indexing can take a long time with no indication of status
## Solution
This triggers an initial pending state for files, folders, code, and workspace which is turned off individually once indexing is complete for that item

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
```diff
@@ -22,6 +25,8 @@ export class ContextCommandsProvider implements Disposable {
         this.registerContextCommandHandler().catch(e =>
             this.logging.error(`Error registering context command handler: ${e}`)
         )
+        //send initial pending state to client immediately
+        void this.processContextCommandUpdate([]).catch(() => {})
```
manodnyab, line 29: why do we need the catch block if we arent doing anything with it?

constewart9: I added just in case there is an error that reaches this we just discard it, but it seems that most other calls to processContextCommandUpdate() are not caught so I will update to match this standard.

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
```diff
@@ -152,10 +168,12 @@ export class ContextCommandsProvider implements Disposable {
             icon: 'image',
             placeholder: 'Select an image file',
         }
-        const workspaceCmd = {
+
+        const workspaceCmd: ContextCommand = {
```
constewart9, line 172: I am not sure why workspaceCmd was not a ContextCommand before. I think it was handled fine by duck typing but this change would bring better type safety. Can revert back if not necessary for this project

PR #2061, opened 2025-08-05
fix(amazonq): bundle corect registry.node for windows

tsmithsz:
## Problem

Related PR: 
- https://github.com/aws/language-server-runtimes/pull/654

On Windows, the LSP crashed when loading [registry-js](https://www.npmjs.com/package/registry-js) because the native module `registry.node` in `servers.zip` was the wrong binary/ABI (seen as `is not a valid Win32 application`). Our webpack step doesn’t bundle native .node files, and the packaging flow could leave an incorrect or missing registry.node in the final Windows servers.zip.

## Solution

Vendored the correct `registry-js` [v1.16.1](https://github.com/desktop/registry-js/releases/tag/v1.16.1) binary and updated `package.sh` to: 

- Copy that vendored `registry.node` into the root of the Windows bundle during zip creation.
- Exclude any other `registry.node` that could be re-added later in the “bundle overlay” step, ensuring a single authoritative binary ends up in `servers.zip`.
- Docs: Noted the dependency, version, and update instructions in `app/aws-lsp-codewhisperer-runtimes/README.md`.

## Testing

- Ran the `npm run ci:generate:agent-standalone -w app/aws-lsp-codewhisperer-runtimes` script and confirmed the generated `registry.node` file in the `servers.zip` hash matched the vendored binary
- Ran this script on mac and windows, then replaced the `aws-lsp-codewhisperer.js` and the `registry.node` files in the plugin cache `servers.zip`, with the generated version from our script. The code worked as expected without any LSP crashes. See the [related PR linked above](https://github.com/aws/language-server-runtimes/pull/654) for the results

## License


app/aws-lsp-codewhisperer-runtimes/scripts/package.sh
```diff
@@ -105,7 +106,7 @@ done
 cd ./build/private/bundle
 for config in "${configs[@]}"; do
     cd ${config}
-    zip -r ../../../../$ARCHIVES_DIR/${config}/win-x64/servers.zip .
+    zip -r ../../../../$ARCHIVES_DIR/${config}/win-x64/servers.zip . -x 'registry.node' '*/registry.node'
```
awschristou, line 109: We shouldn't need this because your change on line 74 already adds the file to the zip. If we add more entries to the `configs` array (not likely) we can bundle if or as needed.

tsmithsz: Removing this line causes the `registry.node` I added to be overwritten by a different copy (I believe it's the one emitted by node-loader under `build/private/bundle/**`)

awschristou: Oh I see, `npm run webpack:prod` places the file into `app\aws-lsp-codewhisperer-runtimes\build\private\bundle\agent-standalone\registry.node`

So what we could do is update `ci:generate:agent-standalone`:
- add a new script into app\aws-lsp-codewhisperer-runtimes\package.json called `copy:native-deps:agent-standalone`
- `copy:native-deps:agent-standalone` will be similar to `copy:resources:agent-standalone`
- `copy:native-deps:agent-standalone` will copy registry.node from _bundle-assets into `app\aws-lsp-codewhisperer-runtimes\build\private\bundle\agent-standalone\registry.node`
- update `ci:generate:agent-standalone` to call `npm run copy:native-deps:agent-standalone` before the call is made to `npm run copy:resources:agent-standalone`

tsmithsz: This makes sense, but won't this also use this `registry.node` file in the `servers.zip` for other platforms instead of the one generated from webpacking? Do we want to do that?

awschristou (approved):

PR #2080, opened 2025-08-07
fix: removing fast glob dependency and using browser compatible logic

suprajaven:
## Problem
`fast-glob` with Node.js-only dependencies was causing browser runtime errors. Replaced that with a. manual implementation using browser-compatible `fdir` and Node.js `fs` APIs

## Solution
- Remove incompatible dependency 
- Replaced with a manual logic to discovering all `.gitignore` and `.npmignore` files in the directory tree and apply proper path prefixing

## License

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16760,74 +16760,6 @@
                 "url": "https://github.com/sponsors/ljharb"
             }
         },
-        "node_modules/globby": {
-            "version": "14.1.0",
-            "resolved": "https://registry.npmjs.org/globby/-/globby-14.1.0.tgz",
-            "integrity": "sha512-0Ia46fDOaT7k4og1PDW4YbodWWr3scS2vAr2lTbsplOt2WkKp0vQbkI9wKis/T5LV/dqPjO3bpS/z6GTJB82LA==",
-            "license": "MIT",
-            "dependencies": {
-                "@sindresorhus/merge-streams": "^2.1.0",
-                "fast-glob": "^3.3.3",
-                "ignore": "^7.0.3",
-                "path-type": "^6.0.0",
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/shared/utils.ts
```diff
@@ -503,65 +502,100 @@ export function hasConnectionExpired(error: any) {
...
-        if (err.code !== 'ENOENT') {
-            console.log('Preindexing walk: gitIgnore file could not be read', err)
-        }
-    }
+        const MAX_FILES = 500_000
+
+        async function findIgnoreFiles(dir: string): Promise<string[]> {
+            const ignoreFiles: string[] = []
+
+            async function walkDir(currentDir: string): Promise<void> {
```
manodnyab, line 515: Manually walking through files adds significant delay when traversing large projects so we cannot add this without significant optimisations.

server/aws-lsp-codewhisperer/src/shared/utils.ts
```diff
@@ -503,65 +502,100 @@ export function hasConnectionExpired(error: any) {
...
+                try {
+                    const entries = await fs.readdir(currentDir, { withFileTypes: true })
+
+                    for (const entry of entries) {
+                        const fullPath = path.join(currentDir, entry.name)
+
+                        if (entry.isFile() && (entry.name === '.gitignore' || entry.name === '.npmignore')) {
```
manodnyab, line 522: we need to add files with common ignore patterns and all the contents in the gitignore and npmignore files

manodnyab (changes requested):

PR #2092, opened 2025-08-08
fix(amazonq): leverage lcs to find the chars added and removed

andrewyuq:
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
@@ -55,8 +55,9 @@ import {
 } from './telemetry'
 import { DocumentChangedListener } from './documentChangedListener'
 import { EditCompletionHandler } from './editCompletionHandler'
-import { EMPTY_RESULT } from './constants'
+import { EMPTY_RESULT, ABAP_EXTENSIONS } from './constants'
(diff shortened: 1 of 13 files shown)


server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts
```diff
@@ -630,22 +630,14 @@ export const CodewhispererServerFactory =
...
-                                session.language,
-                                addedCharactersForEditSuggestion,
-                                true
-                            )
-                            enqueueCodeDiffEntry(session, acceptedSuggestion, addedCharactersForEditSuggestion)
-                        }
+                        const { addedLines, deletedLines } = getAddedAndDeletedLines(acceptedSuggestion.content)
+                        const charDiffResult = getCharacterDifferences(addedLines, deletedLines)
+                        addedCharactersForEditSuggestion = 'a'.repeat(charDiffResult.charactersAdded)
+                        deletedCharactersForEditSuggestion = 'a'.repeat(charDiffResult.charactersRemoved)
```
Will-ShaoHua, line 636: should we just pass the length of it instead of a fake string

andrewyuq: I have though about that. The current countToken() only takes strings since it's inherited from inline suggestions. We will add the countToken() that takes a number later because there are some non-trivial logic in especially countTotalTokens.

PR #2098, opened 2025-08-09
feat(amazonq): add inline suggestion support for iam auth client

parameja1:
## Problem
Inline suggestion currently is not supported for IAM Auth client and throws error message in logs
```Amazon Q service is not signed in```
## Solution
- Changes are to add baseRequest type that can support both IAM based GenerateRecommendations Request along with Token based GenerateSuggestionsRequest and handle the request creation and response parsing properly

## Testing
- Tested in local using the test configuration ``` "configurations": ["CodeWhisperer Server IAM", "Attach to AWS Documents Language Server"]```
- 
image

## License


chungjac:
From slack: 

> Note: This is for testing alone

PR #2113, opened 2025-08-12
feat(amazonq): read tool ui revamp

ctlai95:
## Problem

Previously reverted this change https://github.com/aws/language-servers/pull/2104 due to a UI bug where contextList files were getting added to executeBash header.

## Solution

Revert the revert, and fixed the bug: https://github.com/aws/language-servers/commit/7ceb635bd21686c6664597ef1889114bdfab762e

## License

### DIFF
--- a/chat-client/src/client/mynahUi.ts
+++ b/chat-client/src/client/mynahUi.ts
@@ -1353,10 +1353,15 @@ export const createMynahUi = (
                     fileTreeTitle: '',
                     hideFileCount: true,
                     details: toDetailsWithoutIcon(header.fileList.details),
+                    renderAsPills:
+                        !header.fileList.details ||
+                        (Object.values(header.fileList.details).every(detail => !detail.changes) &&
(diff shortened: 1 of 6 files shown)


jguoamz (approved):

chat-client/src/client/mynahUi.ts
```diff
@@ -1353,10 +1353,15 @@ export const createMynahUi = (
                     fileTreeTitle: '',
                     hideFileCount: true,
                     details: toDetailsWithoutIcon(header.fileList.details),
+                    renderAsPills:
+                        !header.fileList.details ||
```
ashishrp-aws, line 1357: Any better way to do these checks on server side?

ctlai95: IMO it should not be on the server side because this is a UI-specific property

ashishrp-aws (approved):

PR #2122, opened 2025-08-13
fix: handle concurrent calls to init() in localProjectContextController 

constewart9:
## Problem
When a VS Code or JetBrains workspace is opened, the enable indexing option is ignored. Instead, only the default index is created which includes a repomap for use in inline completion but not the full workspace context. 

This is caused by two back to back calls to init -- one with the default config and one with the actual config. Currently, the actual config information is thrown away when the index is built.
## Solution
Add a marker to the default config to differentiate it. Only call init() for localProjectContextController if the actual config is being sent. 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts
@@ -11,6 +11,7 @@ type AgenticChatErrorCode =
     | 'MCPServerInitTimeout' // mcp server failed to start within allowed time
     | 'MCPToolExecTimeout' // mcp tool call failed to complete within allowed time
     | 'MCPServerConnectionFailed' // mcp server failed to connect
+    | 'MCPServerAuthFailed' // mcp server failed to complete auth flow
     | 'RequestAborted' // request was aborted by the user
     | 'RequestThrottled' // request was aborted by the user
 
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -210,7 +218,7 @@ export class LocalProjectContextController {
     // public for test
     async buildIndex(indexingType: string): Promise<void> {
         if (this._isIndexingInProgress) {
-            return
+            await waitUntil(async () => !this._isIndexingInProgress, { interval: 100, timeout: 30000 })
```
manodnyab, line 221: how are we tracking is this is the first indexing in this instance?

manodnyab:
can we skip buildIndex before the actual config is respected?

server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
constewart9: This doesn't track if this is the first indexing. It is more general to handle any concurrent calls to buildIndex while allowing some more leeway while the first call to indexing finishes. It is a problem, though that it will likely time out on larger projects since the build Index takes more than 30 seconds. I will look into addressing this.

constewart9: Also for skip buildIndex before the actual config is respected, is there a case where we want the default config? And therefore not want to skip buildIndex? Such as there is an error getting the actual config? I think if that is the case, then we should change how this works upstream and perhaps not send any config file until we know which is needed. But for init() we can make it slightly more robust to handling updates close together

PR #2145, opened 2025-08-20
fix(amazonq): don't let flare send discard for the still valid suggestion in JB

andrewyuq:
if a valid sessionId is returned back to JB, JB will eventually send a decision for it, but when user types a character to reject the current suggestion, on JB side it will 1) send reject for the current one 2) send a new trigger for the latest context. 2) will happen before 1) which will discard the current active session on flare's end. We don't want flare to do that for JB.

Additonally createSession doesn't need to close the previous session as it should be handled together with telemetry reporting

## License


Will-ShaoHua (approved):

leigaol (approved):
merge after CI

PR #2173, opened 2025-08-27
fix: fix pkce windows url path

bywang56:
## Problem
url path is not correct on Windows
## Solution
Fix it along with some other issues from bug bash

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpOauthClient.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpOauthClient.ts
@@ -10,7 +10,7 @@ import { spawn } from 'child_process'
 import { URL, URLSearchParams } from 'url'
 import * as http from 'http'
 import * as os from 'os'
-import { Logger, Workspace } from '@aws/language-server-runtimes/server-interface'
+import { Logger, Workspace, Lsp } from '@aws/language-server-runtimes/server-interface'
 
 interface Token {
     access_token: string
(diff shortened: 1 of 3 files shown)


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpOauthClient.ts
```diff
@@ -353,17 +354,24 @@ export class OAuthClient {
...
+                      } as const,
                   }
                 : process.platform === 'darwin'
-                  ? { cmd: 'open', args: [authz.toString()] }
-                  : { cmd: 'xdg-open', args: [authz.toString()] }
+                  ? { cmd: 'open', args: [urlStr], opts: { detached: true, stdio: 'ignore' } as const }
+                  : { cmd: 'xdg-open', args: [urlStr], opts: { detached: true, stdio: 'ignore' } as const }
 
-        void spawn(opener.cmd, opener.args, { detached: true, stdio: 'ignore' }).unref()
+        spawn(opener.cmd, opener.args, opener.opts).unref()
```
rli, line 374: should probably be just the `window_showDocument` message since all the clients handle opening external browser correctly

also i do not think this logic as written will not work on VSC remote / JB remote

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpOauthClient.ts
```diff
@@ -95,10 +97,11 @@ export class OAuthClient {
         const savedReg = await this.read<Registration>(regPath)
         if (savedReg) {
             const port = Number(new URL(savedReg.redirect_uri).port)
+            const normalized = `http://127.0.0.1:${port}`
```
ashishrp-aws, line 100: is this a constant port URL? if so can we move it constants?

ashishrp-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpOauthClient.ts
rli: port should be read from the server, not the registration

laileni-aws (approved):

PR #2195, opened 2025-09-02
feat(amazonq): default to diff-based scans

BlakeLazarine:
## Problem
1) Changing default behavior of scans to be diff scanning

2) There was a customer who had enough long findings that it exceeded the limit for the tool return size

## Solution

1) Use prompts to change the default behavior to CODE_DIFF_REVIEW rather than FULL_REVIEW

Also use prompts to inform the customer which type of review is being run. This has proven quite difficult. The best I have been able to do is encourage the start message to take the shape 

```
I'll scan the diff for the following files/folders:

GuruCodeReviewDemo/src/GuruCodeReviewDemo/src/com/amazonaws/guru/demoapp/CsvIterator.java
```

which usually works but not always

The added prompt in the `/review` re-direct makes it very clear that the diff will be scanned rather than the full review. This works consistently

2) set the maximum number of findings to 40 instead of 50. The finding length is limited and 40 will guarantee the messages will fit

## License


rli (approved):
i am wondering where the hardcoded length limit is coming from and if we need to address anything there

PR #2201, opened 2025-09-03
[internal review]feat: add custom_transformation folder support to artifact.zip

Rajanna-Karthik:
## Summary
Adds support for including a `custom_transformation` folder from the solution root into the artifact.zip for .NET transformations.

## Changes
- Add `custom_transformation` folder detection and inclusion in artifact creation
- Update ArtifactManager constructor to accept solution root path
- Update tests to match new constructor signature

## Testing
- Existing tests updated and passing
- Custom transformation folder is only included if it exists in solution root

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
@@ -22,16 +22,19 @@ const zipFileName = 'artifact.zip'
 const sourceCodeFolderName = 'sourceCode'
 const packagesFolderName = 'packages'
 const thirdPartyPackageFolderName = 'thirdpartypackages'
+const customTransformationFolderName = 'custom_transformation'
 
 export class ArtifactManager {
     private workspace: Workspace
     private logging: Logging
     private workspacePath: string
+    private solutionRootPath: string
 
-    constructor(workspace: Workspace, logging: Logging, workspacePath: string) {
+    constructor(workspace: Workspace, logging: Logging, workspacePath: string, solutionRootPath: string) {
         this.workspace = workspace
         this.logging = logging
         this.workspacePath = workspacePath
+        this.solutionRootPath = solutionRootPath
     }
 
     async createZip(request: StartTransformRequest): Promise<string> {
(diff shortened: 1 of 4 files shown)


JiayuW2 (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -282,6 +285,14 @@ export class ArtifactManager {
             this.logging.log('Cannot find artifacts folder')
             return ''
         }
+
+        const customTransformationPath = path.join(this.solutionRootPath, customTransformationFolderName)
+        if (fs.existsSync(customTransformationPath)) {
+            this.logging.log(`Adding custom transformation folder to artifact: ${customTransformationPath}`)
+            const artifactCustomTransformationPath = path.join(folderPath, customTransformationFolderName)
+            fs.cpSync(customTransformationPath, artifactCustomTransformationPath, { recursive: true })
```
jonlouie, line 293: Should we put a try-catch around this? What happens if there's a file writing exception?

pranav-firake, line 293: fs.cpSync will fail for any content >2GB. are we expecting custom_transformation folder bloating to 2gb?

pranav-firake, line 293: do we know what file types we are expecting in this folder? should we only copy files that match expected file types from this folder?

Rajanna-Karthik: We are just using this to test and have some basic files inside it, for now we just copy the entire folder as is and zip it, we will also be giving clear instructions to customer using this to follow the guidelines provided to create this folder.

Rajanna-Karthik: will add it in next revision

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -282,6 +285,19 @@ export class ArtifactManager {
...
+        try {
+            await fs.promises.access(customTransformationPath)
+            try {
+                this.logging.log(`Adding custom transformation folder to artifact: ${customTransformationPath}`)
+                const artifactCustomTransformationPath = path.join(folderPath, customTransformationFolderName)
+                await fs.promises.cp(customTransformationPath, artifactCustomTransformationPath, { recursive: true })
+            } catch (error) {
+                this.logging.warn(`Failed to copy custom transformation folder: ${error}`)
+            }
+        } catch {}
```
pranav-firake, line 299: lets add log in this catch too. If access is not present then we can identify that thru this new log.

jonlouie: +1. Swallowing exceptions is a code smell. If we truly do not need the exception, we should leave a comment explaining why, or even add a log message explaining the benign behavior (e.g. `"Failed to access resources used for custom transformations. Continuing transformation without customizations. ${e}"`).

pranav-firake (approved):

rli (approved):

PR #2207, opened 2025-09-03
sync main

liumofei-amazon:
## License


LiGaCu (approved):

laileni-aws (approved):
Please opt from `Merge commit` before merging this PR or else git history won't align.

laileni-aws:
> Please opt from `Merge commit` before merging this PR or else git history won't align.

Actually this is not enabled in Settings so even though PR gets merged but still we can sync history of both branches.

Only option is to enable the `Merge Commit` option from settings.

PR #2224, opened 2025-09-09
fix: node 24 setup

laileni-aws:
## Notes:
-  Adding node 24 to overrides.json 
- Replacing wget with curl to support with out installing wget.

## License


attribution/overrides.json
```diff
@@ -11,5 +11,12 @@
     },
     "caniuse-lite": {
         "ignore": true
+    },
+    "node": {
```
shruti0085, line 15: out of curiousity: is there a reason why we didnt have node 18 text in this file before?

laileni-aws: Not sure on this! This was not part of this PR: https://github.com/aws/language-servers/pull/2019
Right now, [CI builds requires this](https://github.com/aws/language-servers/actions/runs/17589960484/job/49970907449)

ctlai95 (approved):

PR #2287, opened 2025-09-16
fix: inline UTD empty cases dont differentiate Edit and Completion

Will-ShaoHua:
## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/inline-completion/session/sessionManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/inline-completion/session/sessionManager.ts
@@ -91,6 +91,10 @@ export class CodeWhispererSession {
     customizationArn?: string
     includeImportsWithSuggestions?: boolean
     codewhispererSuggestionImportCount: number = 0
+
+    // Suggestion type specified by the clients, could be either "EDIT" or "COMPLETION"
+    predictionType?: string
+    // Suggestion type "actually" returned from the service, could be either "EDIT" or "COMPLETION"
     suggestionType?: string
     // Track the most recent itemId for paginated Edit suggestions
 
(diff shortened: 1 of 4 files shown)


Will-ShaoHua:
#2288

server/aws-lsp-codewhisperer/src/shared/telemetry/telemetryService.ts
```diff
@@ -258,6 +258,7 @@ export class TelemetryService {
             session.triggerType === 'OnDemand' ? session.timeToFirstRecommendation : timeSinceLastUserModification
         const isInlineEdit = session.suggestionType === SuggestionType.EDIT
```
atonaamz, line 259: I think we can also use `session.predictionType` here and remove the `suggestionType` field. `predictionType` in the request and response should be matching.

server/aws-lsp-codewhisperer/src/language-server/inline-completion/session/sessionManager.ts
```diff
@@ -91,6 +91,10 @@ export class CodeWhispererSession {
     customizationArn?: string
     includeImportsWithSuggestions?: boolean
     codewhispererSuggestionImportCount: number = 0
+
+    // Suggestion type specified by the clients, could be either "EDIT" or "COMPLETION"
+    predictionType?: SuggestionType
+    // Suggestion type "actually" returned from the service, could be either "EDIT" or "COMPLETION"
     suggestionType?: string
```
atonaamz, line 98: I think having one field is enough. But we can follow up later.

Will-ShaoHua: yep you are right. Let me update this rn.

ctlai95 (approved):

atonaamz (approved):

PR #2288, opened 2025-09-16
fix: inline UTD telemetry empty cases dont differentiate Edit and Completion

Will-ShaoHua:
## Problem

suggestionType is inferred by the statements
https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/shared/codeWhispererService.ts#L544-L572

specifically
```
if (apiResponse?.predictions && apiResponse.predictions.length > 0) {
            const suggestionType = apiResponse.predictions[0].edit ? SuggestionType.EDIT : SuggestionType.COMPLETION
            const predictionType = suggestionType === SuggestionType.COMPLETION ? 'completion' : 'edit'

            return {
                suggestions: apiResponse.predictions.map(prediction => ({
                    content: prediction[predictionType]?.content ?? '',
                    references: prediction[predictionType]?.references ?? [],
                    itemId: this.generateItemId(),
                })),
                suggestionType,
                responseContext,
            }
        }
```
so if it's an Edit request and service returns empty suggestion then it will be determined as "Completion".

## Solution
If no suggestion, use client specified predictionType as suggestion type.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/codeWhispererService.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/codeWhispererService.ts
@@ -543,20 +543,37 @@ export class CodeWhispererServiceToken extends CodeWhispererServiceBase {
 
     private mapCodeWhispererApiResponseToSuggestion(
         apiResponse: GenerateCompletionsResponse,
+        request: GenerateTokenSuggestionsRequest,
         responseContext: ResponseContext
     ): GenerateSuggestionsResponse {
-        if (apiResponse?.predictions && apiResponse.predictions.length > 0) {
-            const suggestionType = apiResponse.predictions[0].edit ? SuggestionType.EDIT : SuggestionType.COMPLETION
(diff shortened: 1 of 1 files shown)


server/aws-lsp-codewhisperer/src/shared/codeWhispererService.ts
```diff
@@ -543,20 +543,37 @@ export class CodeWhispererServiceToken extends CodeWhispererServiceBase {
 
     private mapCodeWhispererApiResponseToSuggestion(
         apiResponse: GenerateCompletionsResponse,
+        request: GenerateTokenSuggestionsRequest,
```
ctlai95, line 546: It looks like you only need `request.predictionTypes`? Can we make the argument type `PredictionType[]` instead?

Will-ShaoHua: yea makes sense to me, updated Tai.

PR #2296, opened 2025-09-17
fix(amazonq): fix for mcp permissions read/write inconsistencies

ashishrp-aws:
## Problem

Seeing MCP permission read/write inconsistencies.

## Solution

1. Permission Precedence Logic
• Fixed inconsistent behavior where exact matches weren't properly prioritized over wildcard patterns
• Resolved cases where allowedTools patterns could override explicit tools configurations
• Ensured that exact tool matches take precedence over server-wide or wildcard patterns

2. Enhanced Constructor Dependencies
• Added optional callback functions to AgentPermissionManager constructor:
  • getAvailableTools: Retrieves available tools for a specific server
  • getAllAvailableServers: Gets all available MCP servers
  • getAllBuiltinTools: Returns built-in tool names
• These dependencies enable more accurate permission resolution by providing runtime context

3. Improved Wildcard Handling
• Enhanced pattern matching for server-level wildcards (e.g., @*-mcp)
• Added support for @builtin pattern to match built-in tools specifically
• Fixed global wildcard (*) behavior to work consistently across all scenarios

4. Refined Permission Setting Logic
• Improved setToolPermission method to handle conflicting wildcards more intelligently
• Added proper validation to ensure tools are enabled before being marked as always allowed
• Enhanced conflict resolution when switching between permission types

https://github.com/user-attachments/assets/815bb8fd-4ef2-4298-8f56-513a2860306f

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/agentPermissionManager.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/agentPermissionManager.test.ts
@@ -19,7 +19,12 @@ describe('AgentPermissionManager', () => {
             tools: [],
             allowedTools: [],
         }
-        manager = new AgentPermissionManager(agentConfig)
+        manager = new AgentPermissionManager(
+            agentConfig,
+            undefined, // getAvailableTools
+            undefined, // getAllAvailableServers
+            undefined // getAllBuiltinTools
+        )
     })
 
     describe('matchesPattern', () => {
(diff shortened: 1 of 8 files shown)


tsmithsz (approved):

chungjac (approved):
Nit: Seems like this is used a lot. Maybe we can consider abstracting in a follow up
```
const toolId = serverName ? `@${serverName}/${toolName}` : toolName
const serverPrefix = serverName ? `@${serverName}` : ''
```

PR #2298, opened 2025-09-18
fix(amazonq): fix for delete mcp for mcp config

ashishrp-aws:
## Problem

Fixing bugs for mcp config support

## Solution
- `useLegacyMcpJson` will be set to true by default
- agent config file will be created for mcp config at their level
- delete workflow fix
- disable workflow fix
- refresh for discover mcps from config file at workspace level.

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/mcp/mcpUtils.ts
```diff
@@ -1116,6 +1118,9 @@ export async function migrateAgentConfigToCLIFormat(
             }
         }
 
+        config.useLegacyMcpJson = true
+        updated = true
```
bywang56, line 1122: if always set to true, why we still need it and check below?

PR #2330, opened 2025-09-24
fix(amazonq): handle IAM credentials expiration field to be aws sdk version compatible

parameja1:
## Problem
- IAM credentials type doesnt support expiration field properly for different sdk versions
## Solution
Standardizes IAM credentials expiration field handling across the codebase to support both AWS SDK v2 (`expireTime`) and v3 (`expiration`) formats
- - Added unit tests for credential field validation and normalization

## License


core/aws-lsp-core/src/credentials/credentialsProvider.ts
```diff
@@ -4,6 +4,8 @@ export interface IamCredentials {
     accessKeyId: string
     secretAccessKey: string
     sessionToken?: string
+    expiration?: Date // v3 format
+    expireTime?: Date // v2 format
```
rli, line 8:
```suggestion
```
can we just normalize it directly to v3 where needed

we intend to remove all references to v2 in short order

ashishrp-aws (approved):

core/aws-lsp-core/src/credentials/credentialsProvider.ts
parameja1: Addressed

rli (approved):

PR #2348, opened 2025-09-25
Fix private package mapping during artifact generation

cslong:
## Problem
When mapping private packages from the IDE extension to the reference list, we were trying to match the package id to any part of the reference path. This means that for packages that share a substring, like aws.sdk.core, and aws.sdk.core.extensions, it's possible for the package of aws.sdk.core to map to the reference for extensions.

## Solution
Instead of mapping on the package id only, check for package id + .dll. Also make sure to lowercase since we lowercase all the reference paths earlier in the code.

Also fix issue introduced recently with file filter that was stopping packages from being copied to artifact.zip and set in the requirements.json

This does leave a small gap where there is a package A that contains multiple assembly files that do not match the package name.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
@@ -26,10 +26,7 @@ const customTransformationFolderName = 'customTransformation'
 const filteredExtensions = ['.suo', '.meta', '.user', '.obj', '.tmp', '.log', '.dbmdl', '.jfm', '.pdb']
 const filteredDirectories = ['.git', 'bin', 'obj', '.idea', '.vs', 'artifactworkspace', 'node_modules', 'nuget.config']
 const failedCopies: string[] = []
-const filteredPathRegex: RegExp[] = [
-    /\\users\\[^\\]+\\appdata/i, // IgnoreCase flag with 'i'
-    /.+\.(vspscc|vssscc)$/,
-]
+const filteredPathRegex: RegExp[] = [/.+\.(vspscc|vssscc)$/]
 
 export class ArtifactManager {
     private workspace: Workspace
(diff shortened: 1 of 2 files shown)


server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -261,16 +261,13 @@ export class ArtifactManager {
...
-        reference: ExternalReference,
-        artifactReference: References
-    ): Promise<void> {
+    async processPrivatePackages(request: StartTransformRequest, artifactReference: References): Promise<void> {
         if (!request.PackageReferences) {
             return
         }
         var thirdPartyPackage = request.PackageReferences.find(
-            p => p.IsPrivatePackage && reference.RelativePath.includes(p.Id)
+            // should be toLower because we to lower case the reference paths
```
pranav-firake, line 269: non blocking, :nit I think missing 'need' word here.

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -261,16 +261,13 @@ export class ArtifactManager {
...
+            p => p.IsPrivatePackage && artifactReference.relativePath.includes(p.Id.concat('.dll').toLowerCase())
```
pranav-firake, line 270: do we need to add null check for .relativePath?

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
```diff
@@ -261,16 +261,13 @@ export class ArtifactManager {
         } as RequirementJson
     }
 
-    async processPrivatePackages(
-        request: StartTransformRequest,
-        reference: ExternalReference,
```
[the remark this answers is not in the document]
cslong: the type is still used I believe, it's how it comes from the StartTransformRequest before it's added to the requirements.json

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
cslong: no need for 'need'... just not the most grammatically correct English haha

server/aws-lsp-codewhisperer/src/language-server/netTransform/artifactManager.ts
cslong: I think we always have a relative path, it wouldn't be valid without.

huawenm (approved):

pranav-firake (approved):

huawenm (approved):

PR #2356, opened 2025-09-29
fix(amazonq): send full finding details to plugin, partial to agent

BlakeLazarine:
## Problem
We were hitting the 100,000 tool character limit after only ~35 findings. Because of this, we were sending only the top 30 findings to the Code Issues Panel.

## Solution
With this change, we store full and simplified findings, send the full findings to the Code Issues panel and send the simplified findings to the agent.

Simplified findings have ~10% of the characters of full findings, allowing us to return 300 findings and still maintain a margin of security.

Additionally, now if the agent goes over the 100,000 character limit, the issues would still be sent to code issues panel. This would still be bad UX and I will take an action item to add stricter guard rails to ensure this can never occur.

## License


server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/qCodeAnalysis/codeReview.ts
```diff
@@ -505,8 +519,9 @@ export class CodeReview {
 
         return {
             codeReviewId: jobId,
-            message: `${CODE_REVIEW_TOOL_NAME} tool completed successfully. ${scopeMessage} ${findingsExceededLimit ? ` Inform the user that we are limiting findings to top ${CodeReview.MAX_FINDINGS_COUNT} based on severity.` : ''}`,
+            message: `${CODE_REVIEW_TOOL_NAME} tool completed successfully. Please inform the user to use the explain and fix buttons in the Code Issues Panel to get the best information about particular findings. ${scopeMessage} ${findingsExceededLimit ? ` Inform the user that we are limiting findings to top ${CodeReview.MAX_FINDINGS_COUNT} based on severity.` : ''}`,
```
singhAws, line 522: nit: `Please` is not required

singhAws (approved):

PR #2366, opened 2025-09-30
fix: optimize memory bank token usage and add new tab support

tsmithsz:
## Problem
Amazon Q IDE is encountering issues where conversation history is unexpectedly cleared due to exceeding token limits. This leads to a poor user experience and loss of context in ongoing conversations. We need to implement strategies to prevent hitting these limits and handle cases where we do exceed them more gracefully.

## Solution
- Reduced memory bank file processing from 10 to 5 files with custom fsRead limits (20K per file, 100K total vs standard 200K/400K)
- Memory bank generation now automatically opens in a dedicated new tab, keeping original conversations clean (Note/

### TODO/Callouts: 

- We need to open new tabs when memory bank is triggered by the prompt rather than the explicit button. It still runs in the current chat window for this case)
- We don't currently de-dupe the Memory Bank tabs (similar to \transform). We can take a product call on whether or not we want to do this.

## Testing

- Tested on VSCode repository where it always fails due to context exceeded:

https://github.com/user-attachments/assets/815a530d-b41d-4865-893a-26accbc2473c

- Too many tabs case:

https://github.com/user-attachments/assets/0207b344-0482-4192-9fc8-bcd34f3f5fd1

## License

### DIFF
--- a/chat-client/src/client/features/rules.ts
+++ b/chat-client/src/client/features/rules.ts
@@ -1,4 +1,11 @@
-import { MynahIconsType, MynahUI, DetailedListItem, DetailedListItemGroup, MynahIcons } from '@aws/mynah-ui'
+import {
+    MynahIconsType,
+    MynahUI,
+    DetailedListItem,
+    DetailedListItemGroup,
+    MynahIcons,
+    NotificationType,
+} from '@aws/mynah-ui'
 import { Messager } from '../messager'
 import { ListRulesResult } from '@aws/language-server-runtimes-types'
 import { RulesFolder } from '@aws/language-server-runtimes-types'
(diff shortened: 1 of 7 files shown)


chat-client/src/client/features/rules.ts
```diff
@@ -82,14 +89,32 @@ export class RulesList {
...
+            // Show notification that max tabs reached
+            this.mynahUi.notify({
+                content: 'You can only open ten conversation tabs at a time.',
+                type: NotificationType.WARNING,
+            })
+            return
+        }
+
+        // Create a new tab for the memory bank generation
+        const newTabId = this.mynahUi.updateStore('', { tabTitle: 'Memory Bank' })
```
laileni-aws, line 104: Can we check the existing tab with this title may be and if it does not exist then create a new one ?
WDYT?

tsmithsz: Discussed offline. Will check with product on this

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
```diff
@@ -3458,6 +3467,9 @@ export class AgenticChatController implements ChatHandlers {
             },
         })
 
+        // Reset memory bank flag after completion
+        session.isMemoryBankGeneration = false
```
bywang56, line 3471: Should we handle failure case here as well? maybe move this to a new finally block at line 919

chat-client/src/client/features/rules.ts
```diff
@@ -82,14 +89,32 @@ export class RulesList {
...
+        if (newTabId) {
```
bywang56, line 105: what happen if !newTabId?

server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.ts
tsmithsz: A finally block would reset the flag too early since we need it to stay true throughout the chat flow to apply FS_READ memory bank limits. I added the flag reset in the catch block  and `#handleRequestError` to handle the failure cases properly.

chat-client/src/client/features/rules.ts
tsmithsz: Added a new error notification for this case.

PR #2385, opened 2025-10-03
chore(amazonq): add more tests for inline completion codes

atonaamz:
## Problem

add tests for inline utils files

## Solution

add tests for inline utils files and remove unused files

Coverage should go to 85% after this change: https://app.codecov.io/gh/aws/language-servers/commit/b91f1a0de8b711cd1ce24a7389b64daf15e06f72/tree/server/aws-lsp-codewhisperer/src/language-server?dropdown=coverage

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/inline-completion/utils/diffUtils.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/inline-completion/utils/diffUtils.ts
@@ -131,88 +131,6 @@ export function generateDiffContexts(
     }
 }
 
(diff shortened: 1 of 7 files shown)


server/aws-lsp-codewhisperer/src/language-server/inline-completion/utils/diffUtils.ts
```diff
@@ -131,88 +131,6 @@ export function generateDiffContexts(
     }
 }
 
-/** src: https://github.com/aws/aws-toolkit-vscode/blob/3921457b0a2094b831beea0d66cc2cbd2a833890/packages/amazonq/src/app/inline/EditRendering/diffUtils.ts#L18
```
atonaamz, line 134: We are not using these methods anywhere

PR #2407, opened 2025-10-08
fix: nep

Will-ShaoHua:
## License


server/aws-lsp-codewhisperer/src/language-server/inline-completion/utils/diffUtils.ts
```diff
@@ -275,6 +275,22 @@ export function readUdiff(unifiedDiff: string): UnifiedDiff {
...
+    let prev: number | undefined
+    for (const idx of plusIndexes) {
+        if (!prev || idx === prev + 1) {
+            const removedPlus = relevantLines[idx].substring(1)
+            const removedStartNewline = trimStartNewline(removedPlus)
+            singleLine += removedStartNewline
+        } else {
+            break
+        }
+    }
```
Will-ShaoHua, line 292: not being used, should remove

PR #2441, opened 2025-10-24
chore: add worker thread for processWorkspaceFolders

XiaoxuanLu:
## Problem
Node is single-threaded. Past, we used the main thread for processing everything. For large projects (e.g., Chromium with 184k+ files), workspace indexing blocked the main thread indefinitely and never completed. The synchronous file size checking (fs.statSync()) prevented the event loop from processing other operations, causing both file collection and vector library indexing to fail.

## Solution
Implemented a worker thread to offload file size checking from the main thread:
- Worker Thread - Handles fs.statSync() calls in a separate thread, preventing main thread blocking
- Batch Processing - Sends files in 10k batches to avoid stack overflow
- Parallel Execution - File processing and vector library operations now run simultaneously without blocking each other

## Changes
- Added: server/aws-lsp-codewhisperer/src/shared/fileProcessingWorker.js - Worker thread implementation
- Modified: localProjectContextController.ts - Worker thread integration with fallback to synchronous processing if worker file not found
- Modified: package.sh - Copies worker file into our bundle

Now large projects complete indexing successfully (previously never finished). Main thread remains responsive throughout indexing. Both file collection and vector library operations are faster due to parallel execution. For small project, there is also large improvement for both of the context command received and index built for repomap.

## License


server/aws-lsp-codewhisperer/src/shared/localProjectContextController.ts
```diff
@@ -404,10 +404,119 @@ export class LocalProjectContextController {
...
+
+                worker.on('message', msg => {
+                    if (msg.type === 'ready') {
+                        // Worker initialized, start sending batches
+                        sendBatches().catch(reject)
+                    } else if (msg.type === 'batchComplete') {
+                        batchesInProgress--
+                    } else if (msg.type === 'result') {
+                        clearTimeout(timeout)
+                        void worker.terminate()
```
ashishrp-aws: try and catch for terminate failures/errors

PR #2464, opened 2025-11-05
fix: send a defaultConfig flag to prevent init with default config 

XiaoxuanLu:
## Problem
Currently the LocalProjectContextController does not respect real config setting for indexing, only the default configuration is called and the real configuration is thrown away if the first call is not finished yet. Thus, the vector indexing won't start and the cache indexing is broken. 

## Solution
Add a defaultConfig flag to make LocalProjectContextController prevent init with default config and only execute the real config 

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
@@ -164,6 +164,11 @@ export const LocalProjectContextServer =
         const updateConfigurationHandler = async (updatedConfig: AmazonQWorkspaceConfig) => {
             logging.log('Updating configuration of local context server')
             try {
+                if (updatedConfig.isDefaultConfig === true) {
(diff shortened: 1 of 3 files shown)


manodnyab (approved):

server/aws-lsp-codewhisperer/src/language-server/localProjectContext/localProjectContextServer.ts
```diff
@@ -164,6 +164,11 @@ export const LocalProjectContextServer =
         const updateConfigurationHandler = async (updatedConfig: AmazonQWorkspaceConfig) => {
             logging.log('Updating configuration of local context server')
             try {
+                if (updatedConfig.isDefaultConfig === true) {
+                    logging.log('Skipping init for default config')
```
shruti0085: nit: can be a lil more specific
```suggestion
                    logging.log('Skipping local project context initialization for default config')
```

shruti0085 (approved):

PR #2696, opened 2026-04-13
fix: typed `@folder` context not showing files in collapsible list + fix merge conflicts

chungjac:
### Problem
A pre-existing bug was found where typing @folder in chat did not include the folder's files in the context sent to the LLM — only pinned @folder worked correctly.

### Solution
   - In additionalContextProvider.ts, the ordering logic looked up the folder path in docMap, but docMap is keyed by individual file paths. A typed @folder would never match, so its files were silently dropped from the context.
   - Added a fallback for folder items that matches all child file entries by path prefix.
   - Pinned @folder was unaffected (bypasses docMap lookup).

Screenshot 2026-04-13 at 3 29 09 PM


chat-client/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.1.51](https://github.com/aws/language-servers/compare/chat-client/v0.1.50...chat-client/v0.1.51) (2026-04-07)
+
+
+### Performance Improvements
```
laileni-aws, line 6: FYI
Not part of this PR

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
```diff
@@ -9,9 +15,48 @@ import { LocalProjectContextController } from '../../../shared/localProjectConte
 import { URI } from 'vscode-uri'
 import { activeFileCmd } from './additionalContextProvider'
 
+/**
+ * Maximum items in the initial `sendContextCommands` push.
+ * The client shows these when the user presses `@` before typing.
+ * Server-side filtering (onFilterContextCommands) searches the full set.
+ */
+export const CONTEXT_COMMAND_PAYLOAD_CAP = 2000
```
laileni-aws, line 23: How did we come up with this number?

laileni-aws (approved):

server/aws-lsp-codewhisperer/src/language-server/agenticChat/context/contextCommandsProvider.ts
Will-ShaoHua: it's just some random number we set to restrict the numbers of files/folders being shown when there is no any search term in the @ feature. As in a regular flow when user just type @, not all files can fit in the view port, and 2000 IMO, is large enough and let the cx understand how @ works and they can start typing whatever they want to include.

Will-ShaoHua (approved):

chat-client/CHANGELOG.md
chungjac: Only this commit is for fixing the bug: https://github.com/aws/language-servers/pull/2696/changes/12fcd01f81f1fdb108f9aedbf2483de964d9f4b2

chungjac: All other changes are just pulling in the changes from this PR: https://github.com/aws/language-servers/pull/2682

And resolving those merge conflicts

PR #2702, opened 2026-04-22
feat(amazonq): add consent prompt for workspace-scoped MCP servers

aseemxs:
## Solution

Screenshot 2026-04-22 at 5 02 01 PM

## License


laileni-aws:
A couple of UX clarity questions:

- Is the "Allow" permission a one-time grant or permanent? From the code it looks permanent (persisted to mcp-approvals.json with no TTL), but the prompt text doesn't tell the user this. Should we clarify that in the dialog — e.g., "This choice will be remembered until the server configuration changes"?

- Is the approval scoped to this specific workspace or global across all workspaces? The code scopes it per-workspace (via fingerprintWorkspace), which is the right call, but again the prompt doesn't make this clear to the user.

PR #2706, opened 2026-04-27
fix: surface startTransform errors instead of swallowing them

Rajanna-Karthik:
When ATX Transform commands fail, the LSP server's executeCommand handler catches errors and silently returns undefined. This makes it impossible to diagnose failures from test output or client-side logs — the only visible symptom is expected undefined to exist with no indication of what went wrong.              
                                                                                                                                                                                                                                                                                                                            
  This caused the LSP integ test alarm (V2135436135) to remain open for 47 days. The actual root cause was a ValidationException: max jobs quota reached, but it took multiple hours of log investigation to discover because the error was swallowed.                                                                      
                                                                                                                                                                                                                                                                                                                            
  Changes                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  - `atxNetTransformServer.ts`: Return { error: String(e) } from the catch block instead of implicitly returning undefined. Callers that check for specific fields (e.g., TransformationJobId) are unaffected — the field will still be undefined. Verified safe against the toolkit C# AtxStartTransformResponse model     
  which uses [DataMember] deserialization and ignores unknown fields.                                                                                                                                                                                                                                                       
  - `atxTransformHandler.ts`: Enhanced createJob error logging to include requestId and httpStatusCode from the AWS SDK error metadata. No behavior change — still returns null.                                                                                                                                            
  - `atxTransformInteg.test.ts`: Added console.log of the raw startTransform response before the assertion, so errors are visible in CodeBuild output.                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                            
  Testing                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  - Verified toolkit C# AtxStartTransformResponse model ignores unknown error field

<!---
    REMINDER:
    - Read CONTRIBUTING.md first.
    - Add test coverage for your changes.
    - Link to related issues/commits.
    - Testing: how did you test your changes?
    - Screenshots if applicable
-->

## License

### DIFF
--- a/integration-tests/atx-transform-server/src/tests/atxTransformInteg.test.ts
+++ b/integration-tests/atx-transform-server/src/tests/atxTransformInteg.test.ts
@@ -191,6 +191,8 @@ describe('ATX .NET Transform Integration Tests', () => {
             buildStartTransformRequest('IntegTest-BobsBookstore-' + Date.now(), sourceFiles)
         )
 
+        console.log('StartTransform raw response:', JSON.stringify(result))
+
         transformationJobId = result?.TransformationJobId
         expect(transformationJobId).to.exist
         console.log('TransformationJobId:', transformationJobId)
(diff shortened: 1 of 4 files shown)


laileni-aws:
Closing: this branch is stale and carries a pre-fix GitHub Actions workflow flagged by security scanning; the fix is on main (#2861). Re-open from a fresh branch if the change is still needed.

PR #2713, opened 2026-05-02
fix(amazonq): route inline chat through getChatResponse for correct API selection

laileni-aws:
## Problem

Inline chat fails for Kiro Enterprise subscription users on Eclipse with the error:

> "Your subscription does not support this application. Please contact your administrator."

### Root cause:
The `onInlineChatPrompt` handler in `agenticChatController.ts` was hardcoded to call `client.sendMessage()` directly, bypassing the client-type routing logic in `ChatSessionService.getChatResponse()`. The backend (RTS) Kiro Enterprise subscription check requires both the API and user agent to be allowlisted. `SendMessage` is not in `ALLOWLISTED_KIRO_ENTERPRISE_APIS_WITH_SUBSCRIPTION_CHECK`, while `GenerateAssistantResponse` is.

This only affected Eclipse because it is the only IDE that routes inline chat through the LSP `sendInlineChatPrompt` path. VSCode currently bypasses the LSP for inline chat and calls `GenerateAssistantResponse` directly via its own SDK client (gated behind `amazonqLSPInlineChat` which defaults to `false`). JetBrains has no issue.

When VSCode flips `amazonqLSPInlineChat` to `true`, it would have hit the same issue.

## Solution

- Changed `onInlineChatPrompt` to use `ChatSessionService.getChatResponse()` instead of calling `client.sendMessage()` directly. `getChatResponse()` routes through `#performChatRequest()` which selects `generateAssistantResponse` for token-based (SSO/IdC) clients and `sendMessage` only for IAM clients — matching the behavior of regular chat (`onChatPrompt`).
- Updated `#processSendMessageResponseForInlineChat` to handle both GAR and SendMessage response shapes (`generateAssistantResponseResponse` vs `sendMessageResponse`), using the same pattern as `#processAgenticChatResponse`.

## Testing

- Verified inline chat works with Kiro Enterprise subscription by enabling `amazonqLSPInlineChat` in VSCode (which exercises the same LSP code path as Eclipse).
- TypeScript compilation passes with no errors.
- TODO: Will test in Eclipse once this PR gets in to the feature branch.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/agenticChat/agenticChatController.test.ts
@@ -2159,8 +2159,8 @@ describe('AgenticChatController', () => {
             assert.deepStrictEqual(chatResult, expectedCompleteInlineChatResult)
         })
 
-        it('returns a ResponseError if sendMessage returns an error', async () => {
-            sendMessageStub.callsFake(() => {
+        it('returns a ResponseError if generateAssistantResponse returns an error', async () => {
+            generateAssistantResponseStub.callsFake(() => {
                 throw new Error('Error')
             })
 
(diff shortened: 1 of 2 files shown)


laileni-aws:
Looking into CI failures.

Will-ShaoHua (approved):

ashishrp-aws (approved):

PR #2720, opened 2026-05-08
fix: fixing write lock on worklogs json

pranav-firake:
# Fix worklog file lock 

## Problem

1. **Worklogs stuck/empty on reconnect**: The toolkit couldn't read `worklogs.json` because the LSP held an exclusive file lock during `writeFileSync` on every poll cycle. The toolkit's read attempts failed with "The process cannot access the file" for the entire job duration (~12 minutes of continuous failures).

2. **Checkpoint applied 4x redundantly**: `downloadCompletedStepArtifacts` was called on every poll cycle from `getTransformationPlan`. Multiple concurrent calls raced to apply the same checkpoint before `checkpoints-applied.json` was saved.

### utils.ts — Atomic worklog write
- Write to `.tmp` file then `renameSync` to target — avoids holding exclusive lock on the file the toolkit reads
- Added `EPERM` to retry conditions (Windows rename can fail with this)
- On final failure: log and return gracefully instead of throwing (worklogs are non-critical, will retry next poll)

### atxTransformHandler.ts — Concurrency guard
- Added `_applyingCheckpoints` flag to prevent concurrent execution of `downloadCompletedStepArtifacts`
- Resets in `finally` block to ensure cleanup on error

## Testing

1. Start job → verify worklogs update in real-time in the Worklogs tab
2. Disconnect → reconnect → verify worklogs load after reconnect
3. Verify checkpoint is applied only once (check logs for "Saved applied checkpoint" appearing once per step, not 4x)

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
@@ -80,6 +80,7 @@ export class ATXTransformHandler {
     private cachedHitl: string | null = null
     private cachedStepHitl: string | null = null
     private cachedInteractiveMode: InteractiveMode | null = null
+    private _applyingCheckpoints = false
 
     constructor(serviceManager: AtxTokenServiceManager, workspace: Workspace, logging: Logging, runtime: Runtime) {
         this.serviceManager = serviceManager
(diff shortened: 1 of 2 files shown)


Rajanna-Karthik (approved):

zhenganp (approved):

cslong (approved):
lgtm

chungjac (approved):

PR #2742, opened 2026-05-20
fix: canonicalize paths through filesystem in requiresPathAcceptance

aseemxs:
## Summary
- Replaces string-only `path.resolve` in `requiresPathAcceptance` with `fs.promises.realpath` so the workspace-boundary check is performed against the canonical resolved path rather than the literal input.
- Adds an ENOENT fallback that realpaths the parent directory and joins the basename so the agent can still validate paths for files it is about to create (which don't yet exist on disk).
- Adds a regression test that creates a real on-disk symlink whose target lies outside the workspace and asserts that `requiresPathAcceptance` requires user acceptance.

## Why
`path.resolve` is a string-only operation. When given a path that is itself a symlink whose target is outside the workspace, `path.resolve` returns the symlink path unchanged, so the boundary check could incorrectly conclude the path is in-workspace.

**Impact**: because the boundary check gates the user-acceptance prompt for file writes, an in-workspace symlink whose target is outside the workspace can cause the agent to write to that outside target without the user being asked to approve a path outside their workspace. Files anywhere on disk that the user's process can write to (including SSH/AWS config and other dotfiles) could be modified through what appears, to the agent, to be a routine in-workspace write. Resolving via `realpath` ensures the boundary check operates on the canonical target.

**Threat model**: triggered when a user opens a workspace whose contents are not fully trusted (e.g., a cloned repo from an untrusted source) and the agent subsequently writes to a path within that workspace.

**Reproduction**: the included regression test is a self-contained reproduction — it constructs the symlink, calls `requiresPathAcceptance`, and asserts the expected boundary behavior. It fails on the unfixed code and passes after the fix.

**Out of scope**: this change does not address TOCTOU between canonicalization and the eventual write. That's a separate concern and should be tracked independently if desired.

## Test plan
- [x] New unit test fails on the unfixed code and passes after the fix
- [x] All existing `toolShared.test.ts` tests continue to pass (49 passing)
- [x] `eslint` runs clean on changed files (0 errors)

## Screenshot of verification of fix
 
Screenshot 2026-05-21 at 5 34 13 PM

## Related
- Internal ticket: P414322738


ashishrp-aws (approved):

laileni-aws (approved):
Lets test it before merging to main

aseemxs:
End-to-end verification on VS Code with Amazon Q extension:

Loaded the patched LSP via manifest override (version `1.68.0-path-canonicalization-bugbash.1e87a137`).

Reproduced the original behavior on the unpatched LSP:
- Agent wrote through an in-workspace symlink to a target outside the workspace, with no acceptance prompt.

Confirmed the fix on the patched LSP:
- Same agent action now triggers a "Allow file modification outside of your workspace" prompt naming the path. Denying the prompt blocks the write; the outside file is unchanged.

Combined with the included regression test, this confirms the fix works end-to-end through the agent flow, not just at the `requiresPathAcceptance` unit level.

PR #2755, opened 2026-05-28
fix: surface lbv and checkpoint hitls regardless of job status

Rajanna-Karthik:
mirrors the executing-branch behavior in the non-executing branch so the ide picks up pending hitls when the job reports planning or other non-executing statuses. pre-job mode-selection checkpoint stays filtered.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
@@ -1399,6 +1399,64 @@ export class ATXTransformHandler {
                             TransformationPlan: plan,
                         } as AtxGetTransformInfoResponse
                     }
+                    if (hitl.tag === 'local-build-verification') {
+                        if (!this.jobsPastLocalBuild.has(request.TransformationJobId)) {
+                            this.jobsPastLocalBuild.add(request.TransformationJobId)
+                        }
+                        this.logging.log(
+                            `ATX: ${jobStatus} job has pending LBV HITL — taskId=${hitl.taskId}; surfacing AWAITING_HUMAN_INPUT to IDE`
+                        )
+                        return {
+                            TransformationJob: {
+                                WorkspaceId: request.WorkspaceId,
+                                JobId: request.TransformationJobId,
+                                Status: 'AWAITING_HUMAN_INPUT',
+                            } as AtxTransformationJob,
+                            HitlTag: hitl.tag,
+                            HitlTaskId: hitl.taskId,
+                            TransformationPlan: plan,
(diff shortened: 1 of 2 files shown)


pranav-firake (approved):

zhenganp (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -1403,6 +1403,64 @@ export class ATXTransformHandler {
                             TransformationPlan: plan,
                         } as AtxGetTransformInfoResponse
                     }
+                    if (hitl.tag === 'local-build-verification') {
+                        if (!this.jobsPastLocalBuild.has(request.TransformationJobId)) {
```
laileni-aws: Redundant jobsPastLocalBuild check!

The inner if (!has) guard is unnecessary — Set.add() is idempotent. It doesn't cause a bug, but it's noise

laileni-aws (changes requested):
Can we add tests for this ?

laileni-aws:
## Observations & Concerns

| # | Type | Detail |
|---|------|--------|
| 1 | ⚠️ Logic | The catch-all condition `!(String(hitl.tag).endsWith('-checkpoint') && !this.jobsPastLocalBuild.has(request.TransformationJobId))` is a double-negative that's hard to parse. It effectively means "surface if it's NOT a pre-LBV checkpoint". But given the preceding conditions already handle LBV and post-LBV checkpoints, this case should be unreachable for checkpoints — so the guard is always `true`. Consider simplifying to an unconditional surface (or just removing the `if` wrapper). |
| 2 | ⚠️ Missing reference | The code references `this.jobsPastLocalBuild` — this field doesn't appear anywhere in the current `main` branch. Is it being introduced in a dependency PR? If not, this will fail at runtime. |
| 3 | 🟡 No tests | Codecov reports 0% patch coverage (58 lines uncovered). Given this is HITL-surfacing logic that directly affects IDE UX, unit tests for the three new branches would reduce regression risk. |
| 4 | 🟡 EXECUTING branch divergence | The EXECUTING branch (lines 1301–1325) filters checkpoints differently — it excludes *all* `*-checkpoint` tags via `filter(h => !String(h.tag).endsWith('-checkpoint'))`. This PR's non-executing branch conditionally surfaces post-LBV checkpoints. The PR description says "mirrors the executing-branch behavior" but the logic is subtly different — is this intentional? |

Overall the intent is sound — the IDE should never silently miss a pending HITL. Main ask: clarify `jobsPastLocalBuild` existence and consider simplifying the catch-all guard.

laileni-aws (approved):

ashishrp-aws (approved):

PR #2761, opened 2026-06-09
feat: rejecting atx .ent v1 agent jobs

pranav-firake:
# fix: reject V1 legacy agent startTransform requests with upgrade message

## Summary

Block V1 (legacy) IDE extensions from starting new transformation jobs via the LSP. When an old IDE (< 1.91) sends a `startTransform` request without the `useOrchestratorAgent` flag, the LSP now returns an error message telling the customer to update their extension.

## Problem

V1 agent is being deprecated. Old IDE extensions that don't send `useOrchestratorAgent: true` should no longer be allowed to start transformations. We need to inform customers to update their IDE extension.

## Change

**File:** `server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts`

In the `AtxStartTransformCommand` handler, added a check before job creation:

```typescript
if (!useNew) {
    throw new Error(
        'This version of the AWS Toolkit extension is no longer supported for transformations. ' +
        'Please update to the latest version of the AWS Toolkit extension to continue using AWS Transform.'
    )
}
```

## Behavior

| IDE Version | `useOrchestratorAgent` | Result |
|---|---|---|
| Old (< 1.91) | `undefined` / `false` | Error thrown → shows in output pane |
| New (>= 1.91) | `true` | Routes to V2 handler (no change) |

## What the user sees (old IDE)

1. User clicks "Start Transform"
2. Job fails immediately
3. Generic InfoBar: "Your transformation failed; please review transformation logs for details."
4. Output pane shows: "This version of the AWS Toolkit extension is no longer supported for transformations. Please update to the latest version of the AWS Toolkit extension to continue using AWS Transform."

## Scope

- Only `startTransform` is blocked (new job creation)
- `getTransformInfo` and `uploadPlan` still route to legacy handler for in-progress jobs that were started before the update
- No IDE-side changes required
- No impact to V2 flow

## Test Plan

- Start transform from latest IDE (useOrchestratorAgent=true) → works normally
- Simulate V1 request (useOrchestratorAgent=false/missing) → error with upgrade message
- Verify error appears in output pane
- Verify existing in-progress jobs (getTransformInfo polling) still work on old IDE

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
@@ -79,7 +79,7 @@ export const AtxNetTransformServerToken =
                         if (!useNew) {
                             const msg =
                                 'This version of the AWS Toolkit extension is no longer supported for transformations. Please update to the latest version of the AWS Toolkit extension to continue using AWS Transform.'
-                            void lsp.window.showMessage({ type: MessageType.Error, message: msg })
+                            await lsp.window.showMessage({ type: MessageType.Error, message: msg })
                             throw new Error(msg)
                         }
 
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
@@ -75,12 +75,18 @@ export const AtxNetTransformServerToken =
                             params as AtxStartTransformRequest
                         const useNew = useOrchestratorAgent === true
 
+                        if (!useNew) {
+                            throw new Error(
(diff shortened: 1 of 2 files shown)


Rajanna-Karthik (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
```diff
@@ -263,6 +275,7 @@ export const AtxNetTransformServerToken =
                 }
             } catch (e: any) {
                 logging.error(`ATXTransformServer: Error executing command: ${String(e)}`)
+                return { error: true, message: e.message }
```
laileni-aws, line 278: nit:
```
return { error: true, message: e?.message ?? String(e) }

laileni-aws (approved):

PR #2765, opened 2026-06-19
fix(amazonq): preserve customer edits on checkpoint apply (DealerFx netTransform)

jabouayash:
## Problem

When a .NET transformation is re-run, `applyChanges()` in the netTransform
language server laid every backend checkpoint diff onto the customer's solution
with an unconditional `fs.copyFileSync`, with no check for files the customer
had edited locally since the previous apply. A later sync therefore silently
overwrote manual fixes to an already-transformed project — the data-loss
report where hand-applied NuGet/build fixes were clobbered by a retry.

The watermark needed to detect this already existed
(`getModifiedFilesSinceCheckpoint`, `mtime > manifest.lastAppliedTimestamp`),
but was only consulted on the **uplink** (`updateWorkspace`); the **downlink**
(`applyChanges`) never looked at it. The interactive path
(`downloadCompletedStepArtifacts`) also never wrote the watermark at all, so the
guard would have been a no-op on the exact path edits were lost on.

## Solution

- Thread `jobId` into `applyChanges` (defaulted, back-compatible) and compute
  the customer-modified set once.
- A single `shouldPreserveUserFile` guard runs before all three write loops
  (added / updated / moved). If the destination exists, was edited since the
  last apply, and differs byte-for-byte from the incoming artifact, the
  customer's file is preserved (skip the write — and, for a move, skip the
  unlink) and recorded in `conflictedFiles`; a `.bak` copy is written under
  `{jobId}/checkpoints/conflict-backups/`.
- The byte-equal short-circuit means the agent's own re-emitted (unchanged)
  files still apply normally — only genuine local edits are protected.
- Write the watermark (`saveLastAppliedTimestamp`) on the interactive path too,
  so the guard is effective there.

`filesRemoved` is intentionally left out of scope (it is a product decision
whether a delete should be blocked by a local edit).

**Testing:** 6 new tests in `atxTransformHandler.test.ts` over a real temp
filesystem (preserve+backup, overwrite-when-untouched, byte-equal no-op,
filesMoved skip copy+unlink, no-jobId back-compat, backup does not round-trip
into a new job). Full `applyChanges` suite: 22/22 passing. `prettier --check`
and compile are clean.

## License

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
@@ -3427,19 +3427,14 @@ export class ATXTransformHandler {
                 try {
                     const filePath = path.join(solutionRootPath, relativePath)
                     if (fs.existsSync(filePath)) {
-                        if (
-                            userModifiedFiles.size > 0 &&
-                            userModifiedFiles.has(path.resolve(filePath))
-                        ) {
+                        if (userModifiedFiles.size > 0 && userModifiedFiles.has(path.resolve(filePath))) {
                             try {
                                 const backupPath = this.backupConflictedFile(filePath, jobId, solutionRootPath)
                                 this.logging.log(
                                     `ATX: Preserved customer-modified file from deletion, backed up to ${backupPath}: ${filePath}`
                                 )
                             } catch (e) {
-                                this.logging.error(
-                                    `ATX: Failed to back up conflicted file ${filePath}: ${String(e)}`
-                                )
+                                this.logging.error(`ATX: Failed to back up conflicted file ${filePath}: ${String(e)}`)
                             }
                             conflictedFiles.push(path.resolve(filePath))
                             continue
@@ -3518,23 +3513,14 @@ export class ATXTransformHandler {
                         }
                         // If the customer edited the move SOURCE, preserve it in place —
                         // the rename would delete their edited file from its original path.
-                        if (
-                            userModifiedFiles.size > 0 &&
-                            userModifiedFiles.has(path.resolve(beforePath))
-                        ) {
+                        if (userModifiedFiles.size > 0 && userModifiedFiles.has(path.resolve(beforePath))) {
                             try {
-                                const backupPath = this.backupConflictedFile(
-                                    beforePath,
-                                    jobId,
-                                    solutionRootPath
-                                )
+                                const backupPath = this.backupConflictedFile(beforePath, jobId, solutionRootPath)
                                 this.logging.log(
                                     `ATX: Preserved customer-modified move source, backed up to ${backupPath}: ${beforePath}`
                                 )
                             } catch (e) {
-                                this.logging.error(
-                                    `ATX: Failed to back up conflicted file ${beforePath}: ${String(e)}`
-                                )
+                                this.logging.error(`ATX: Failed to back up conflicted file ${beforePath}: ${String(e)}`)
                             }
                             conflictedFiles.push(path.resolve(beforePath))
                             continue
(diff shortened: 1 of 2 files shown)


JiayuW2:
### Summary

#### What tested
I manually tested PR #2765 (commit `b06a0323c`) which adds user-edit preservation to the `applyChanges` checkpoint flow in the netTransform LSP. Using `ConflictTestApp`, I:

1. Started a transform (net6.0 → net10, interactive mode, checkpoints ON)
2. After the first transformation applied (framework + package upgrades), edited .csproj (changed TargetFramework to net8.0 and `Microsoft.Data.SqlClient` back to `System.Data.SqlClient`)
3. Chat "Retry project"
4. After retry finishes, edits are still on disk. When clicking "show results" in the Job plan next to ConflictTestApp(retry), the before is the applied changes from the 1st transformation, and the after is removing one blank line. (local edits don't appear in either before or after)
5. Confirmed folder conflict-backups is created and the local edited file is saved in the folder
6. Confirmed the conflict detection fired: "Found 3 modified files since last checkpoint", "Preserved (conflicts): 1"
(JobId: `9a15fde8-8e42-4f51-94b0-454d21a63ce2`)
Screenshot 2026-06-24 at 10 16 04 PM
Screenshot 2026-06-24 at 10 16 15 PM


#### Bug found
`applyChanges` was being called twice for the same step in a single polling cycle — once via the `downloadDiffArtifact` path and again via the `downloadCompletedStepArtifacts` path. The second call ran after `saveLastAppliedTimestamp` had already updated the watermark, so it saw 0 modified files and silently overwrote the user's edit.


#### Fix applied
Moved `loadAppliedCheckpoints()` from outside the loop to inside the loop in `downloadCompletedStepArtifacts`. This ensures each iteration reads fresh state from disk, so it sees that the diff-artifact path already applied the step and skips the redundant apply.

JiayuW2 (approved):

JiayuW2:
### Summary of changes `e4ccb83`

#### Code changes in atxTransformHandler.ts:

  1. `filesRemoved` guard — If the customer edited a file since the last apply and the transform wants to delete it, the file is preserved on disk, backed up to conflict-backups/, and recorded as a conflict instead of being silently deleted.
  2. `filesMoved` source guard — If the customer edited the move source file, the move is skipped entirely (source stays in place, target not created), backed up to conflict-backups/, and recorded as a conflict. (The existing target guard remains unchanged.)

  New tests (3):
  - filesMoved: skips move when the move SOURCE is user-modified
  - filesRemoved: preserves a customer-edited file instead of deleting it
  - filesRemoved: deletes file if user did NOT edit it (non-regression)

#### Manual Test Summary: `filesRemoved` and `filesMoved` cases

Setup: Used ConflictTestApp (`D:\TestRepos\ConflictTestAppGroup\sourceCode`) with an active transform job that had already completed step 1 (watermark set).

Method: Synthetic checkpoint injection via a Node.js test script (`test-conflict-cases.js`) that:
  1. Edits `Program.cs` (bumps mtime past the watermark so it's detected as customer-modified)
  2. Creates a fake checkpoint folder with metadata targeting that file
  3. Calls applyChanges directly on the handler
  4. Verifies results and cleans up

`filesRemoved test`:
  - Fake checkpoint with `filesRemoved: ["ConflictTestApp/Program.cs"]`
  - Result: Program.cs stayed on disk, backup created in `conflict-backups/`, `conflictedFiles` populated, `filesRemoved: 0`

 `filesMoved` test:
  - Fake checkpoint with `movedFilesMap: [{before: "ConflictTestApp/Program.cs", after: "ConflictTestApp/App.cs"}]`
  - Result: `Program.cs` stayed at original path, `App.cs` not created, backup created in `conflict-backups/`, `conflictedFiles` populated, `filesMoved: 0`

  Both passed.

Rajanna-Karthik (approved):

laileni-aws (approved):

PR #2766, opened 2026-06-23
feat(amazonq): add client-side chat delivery metrics

araarush:
## Problem

  The chat UI runs in a webview and receives inbound messages from the IDE host via postMessage. If handleInboundMessage drops an inbound message — for example a strict origin check
  rejecting it on a host whose webview reports an opaque/empty origin — the chat response never renders, yet the backend still returns HTTP 200. There is currently no client-side signal
  distinguishing "response received by the language server" from "response rendered in the UI," so this class of failure is invisible to server-side metrics. See aws/amazon-q-eclipse#555.

  ## Solution

  Adds two client-side telemetry metrics, emitted from chat-client and re-emitted by Flare's ChatTelemetryController (so they flow to every IDE host through the existing shared telemetry
  path — no per-client definitions needed):

  - amazonq_chatMessageRendered — positive signal, emitted when a chat response is delivered to mynah-ui (gated to the terminal, non-partial chunk to avoid per-chunk volume).
  - amazonq_chatPostMessageRejected — emitted when handleInboundMessage drops an inbound message (reason: untrustedOrigin | undefinedData | unknownCommand).

  product and os are stamped automatically by each host's telemetry envelope; languageServerVersion is included in the metric data.

  ## Testing

  - Unit tests added on both sides (Flare chatTelemetryController.test.ts, including a guard that a single inbound event emits exactly once; chat-client/chat.test.ts). Full suites pass.
  - Verified end-to-end to telemetry on VS Code and Eclipse — both metrics, with correct product / os / languageServerVersion.
  - Not yet validated on JetBrains or Visual Studio. The chat-client emit code is shared across hosts, but per CONTRIBUTING's cross-host requirement I'm flagging that the per-host
  telemetry forwarding/envelope hasn't been confirmed on those two.

  ## License

### DIFF
--- a/chat-client/src/client/chat.test.ts
+++ b/chat-client/src/client/chat.test.ts
@@ -18,6 +18,8 @@ import { createChat } from './chat'
 import * as sinon from 'sinon'
 import { TELEMETRY } from '../contracts/serverContracts'
 import {
+    CHAT_MESSAGE_RENDERED_TELEMETRY_EVENT,
+    CHAT_POST_MESSAGE_REJECTED_TELEMETRY_EVENT,
     ERROR_MESSAGE_TELEMETRY_EVENT,
     SEND_TO_PROMPT_TELEMETRY_EVENT,
     TAB_ADD_TELEMETRY_EVENT,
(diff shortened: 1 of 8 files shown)


araarush:
Tested metrics successfully on Visual Studio Code, Jetbrains, Eclipse, and Visual Studio. End to end metric flow is functional.

ashishrp-aws (approved):

chungjac (approved):

PR #2789, opened 2026-07-20
chore(amazonq): authenticate agentic-chat E2E tests via a dedicated OIDC role

laileni-aws:
## What
Re-point the `q-agentic-chat-server` integration tests (`.github/workflows/integration-tests.yml`) at a team-owned SSO bearer token, authenticated via a dedicated least-privilege GitHub-OIDC role, replacing the previous token-refresher role + lambda that has been removed.

## Changes
- **Configure AWS Credentials** now assumes the read-only OIDC role `arn:aws:iam::550160095699:role/LanguageServersGitHubActionsReadRole` (us-east-1), session name `lsp-integ-ci`.
- **Removed** the `Refresh Token` step (`aws lambda invoke ... TokenRefresherLambda`): the token secret is refreshed out-of-band every ~5 min, and the new role intentionally has no `lambda:InvokeFunction`.
- **Get SSO Token** now does a raw `secretsmanager:GetSecretValue` + `jq` extraction of the access token into `TEST_SSO_TOKEN`, and masks it with `::add-mask::` (public repo).
- Tags traffic with `AWS_SDK_UA_APP_ID=lsp-integ-ci` so the token provider's dashboards can attribute/exclude CI traffic.

## Security
- OIDC trust is scoped to this repo with a tightened `sub` (branch refs, not `:*`) and `aud=sts.amazonaws.com`.
- Least-privilege: `secretsmanager:GetSecretValue` on one secret + `kms:Decrypt` on its key. No write, no lambda.
- Token masked in logs; only non-secret identifiers (account id, role ARN, secret name) appear, consistent with the prior setup.

## Testing
`integration-tests.yml` runs post-merge via `workflow_run` (not a PR gate). Once unblocked, we'll trigger a post-merge run and confirm the OIDC assume succeeds, the token is masked, and `q-agentic-chat-server` passes across the OS matrix. For `workflow_run` events the OIDC `sub` ref will be verified against the trust policy on that first run.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -81,8 +81,8 @@ jobs:
                   npm run test-integ -w integration-tests/q-agentic-chat-server
               env:
                   TEST_SSO_TOKEN: ${{ env.TEST_SSO_TOKEN }}
-                  TEST_SSO_START_URL: ${{ secrets.TEST_SSO_START_URL }}
-                  TEST_PROFILE_ARN: ${{ secrets.TEST_PROFILE_ARN }}
+                  TEST_SSO_START_URL: ${{ secrets.INTEG_TEST_SSO_START_URL }}
+                  TEST_PROFILE_ARN: ${{ secrets.INTEG_TEST_PROFILE_ARN }}
                   TEST_RUNTIME_FILE: ${{ github.workspace }}/downloaded-artifacts/aws-lsp-codewhisperer.js
                   # Tags LSP CI traffic so RTS ServiceMetrics/dashboards can segment or exclude it (BI).
                   AWS_SDK_UA_APP_ID: lsp-integ-ci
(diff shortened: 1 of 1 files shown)


xinyiww1 (approved):

.github/workflows/integration-tests.yml
```diff
@@ -52,26 +52,37 @@ jobs:
             - name: Configure AWS Credentials
               uses: aws-actions/configure-aws-credentials@v4
               with:
-                  role-to-assume: arn:aws:iam::964765661569:role/GitHubActionsTokenRefresherRole
-                  role-session-name: language-servers-github
+                  # DEXP-owned GitHub-OIDC read role in the gamma-IAD canary account (550160095699):
```
ashishrp-aws, line 55: Don't think we should mention accounts on open source PRs

ashishrp-aws (changes requested):

.github/workflows/integration-tests.yml
laileni-aws: Its fine we do mention this even before!

kmmcclai (approved):

ashishrp-aws (approved):

PR #2796, opened 2026-08-11
fix(amazonq): pass notification feature to the Q service manager

ashishrp-aws:
Targets `feature/qdev-signup-message` (not `main`), so a new flare pre-release can be cut from that branch for IDE testing.

## Problem

The access-blocked notification added in #2794 never reaches the client. `AmazonQServiceServerFactory` destructures the features it forwards to the service manager, and `notification` was not among them:

```ts
({ credentialsProvider, lsp, workspace, logging, runtime, sdkInitializator }) => {
```

So `features.notification` was always `undefined`, the guard in `serviceFactory` never passed, `onAccessBlocked` was never assigned, and the notifier could not run. Confirmed empirically: signing in with a Builder ID that RTS blocks produced the denial (the service message came back through chat) but no notification, no sign-out, and no blocked screen.

`notification` is optional on `QServiceManagerFeatures` so existing constructions and test fixtures keep compiling — which is also why omitting it here did not fail the build. It silently disabled client-facing reporting instead. Commented at the call site so the next person adding a feature there doesn't repeat it.

## Also: stable notification id

The notification now carries `id: 'qDevPluginAccessBlocked'`.

Clients need to recognise it without inspecting its text. The message is the service's own copy and is expected to change, and `FEATURE_NOT_SUPPORTED` is reused across several RTS gates (`KIRO_USER_BLOCKED_EXCEPTION_MESSAGE`, `WCS_USER_BLOCKED_EXCEPTION_MESSAGE`), so the reason alone doesn't identify this gate either. Both IDE clients already prefer the id when present and only fall back to matching the title because the released server doesn't send one yet — this lets that fallback be removed.

## Testing

- `tsc --noEmit` clean, prettier clean, 6/6 notifier tests pass (one now asserts the id)
- A server bundle built from this branch (`webpack.config.prod.js`, entry `agent-standalone`) contains `qDevPluginAccessBlocked` and `isQDevPluginAccessBlockedError`; a bundle built from the previous branch head contains neither
- Verified end-to-end in VS Code by pointing `aws.dev.amazonqLsp.path` at that bundle

## Follow-up not included here

The observer is on the token client only; `streamingClientService.ts` has none. In practice sign-in detection still fires, because the gate denies every operation and the A/B config fetch goes through the token client moments after credentials arrive. But chat-time denials — which is where the message surfaces to the user today — are not observed. Happy to add that here or separately.

### DIFF
--- a/server/aws-lsp-codewhisperer/src/shared/amazonQServer.ts
+++ b/server/aws-lsp-codewhisperer/src/shared/amazonQServer.ts
@@ -19,7 +19,7 @@ const LOGGING_PREFIX = '[AMAZON Q SERVER]: '
 
 export const AmazonQServiceServerFactory =
     (serviceManager: (features: QServiceManagerFeatures) => AmazonQBaseServiceManager): Server =>
-    ({ credentialsProvider, lsp, workspace, logging, runtime, sdkInitializator }) => {
+    ({ credentialsProvider, lsp, workspace, logging, runtime, sdkInitializator, notification }) => {
         let amazonQServiceManager: AmazonQBaseServiceManager
 
         const log = (message: string) => {
@@ -39,6 +39,11 @@ export const AmazonQServiceServerFactory =
                 logging,
                 runtime,
                 sdkInitializator,
+                // Required for the service manager to be able to surface anything to the client. It is
+                // optional on QServiceManagerFeatures so that existing constructions (including test
+                // fixtures) keep compiling, which means omitting it here does not fail the build -- it
+                // just silently disables client-facing reporting.
+                notification,
             })
 
             /*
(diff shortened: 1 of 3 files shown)


chungjac (approved):

hezelin-work (approved):

ashishrp-aws:
Correction to this PR's description, for anyone reading back.

I justified the `id` on the grounds that clients could match `id === 'qDevPluginAccessBlocked'`. That is not true. `RouterByServerName.send()` replaces the id before it reaches the client:

```js
const idObject = { serverName: this.serverName, id: params.id }
const id = this.encoding.encode(JSON.stringify(idObject))   // base64
```

So the client receives base64 of `{"serverName":...,"id":"qDevPluginAccessBlocked"}`, never the raw string. `id` is the runtime's followup-routing token, not a semantic identifier.

The id is still worth keeping — it is what makes notification followups routable back to this server — but the stated client-matching rationale was wrong, and both IDE clients currently identify the notification by title instead. I'll follow up on a durable way to identify it.

Separately, the wiring in this PR was necessary but not sufficient: the notification was still dropped because the server never declared `serverInfo`, so the runtime built no notification router. Fixed in #2797.

PR #2798, opened 2026-08-12
feat: scope local build verification to the loaded beamed repo and add beam discovery/download

Rajanna-Karthik:
# feat: scope local build verification to the loaded beamed repo and add beam discovery/download

## What this does

Adds the LSP support for **Beam to IDE** (AWS Transform .NET): the language server mediates between the IDE and the platform so a developer can pull a web-transformed repository into Visual Studio and run a real local build. Scoped entirely to the `netTransform` handler.

## What changed

- **Scoped build-verification HITL selection** (`getHitlAgentArtifact` + the EXECUTING path): when the IDE passes the loaded repo's scope, the handler returns **only that repo's** Local Build Verification request. Critically, when scope is set but no in-scope LBV matches, an out-of-scope (sibling) LBV is **excluded from the fallback** — so on a multi-repo beam the IDE is never handed a sibling repo's build-verification request (the root of "false green"). Includes the mixed pending-HITL case (sibling LBV coexisting with a non-LBV HITL).
- **Beam discovery + artifact download**: `listBeamedRepos` (derives beamed repos from the per-repo beam-status artifacts, tolerant of the status-label naming) and `downloadBeamArtifact`.
- **Chat scoping marker**: prepends the routing token to outbound chat so the orchestrator scopes context to the loaded repo (the IDE strips it at display).
- Removed a vestigial, uncalled `createBeamJob` command to keep the surface clean.

## Files

Exactly three files, all under `server/aws-lsp-codewhisperer/src/language-server/netTransform/`:
- `atxTransformHandler.ts` — scoping guard, discovery, download, marker.
- `atxNetTransformServer.ts` — command registration/routing.
- `atxModels.ts` — new request/response types (additive).

## Additive by design

New request/response fields are optional; the scoping logic is a **no-op when no beam scope is provided** — a normal (non-beam) transform selects HITLs exactly as before (verified: with no scope, the new fallback collapses to the original selection). One shared helper (`createArtifactDownloadUrl`) gained throttle-retry resilience — behavior-preserving on success and on missing-URL, only adds a retry on transient throttling.

## Testing

- **Beam, live (multi-repo):** discovery returns the beamed repos; per-repo LBV isolation held; no sibling leak; builds completed cleanly.
- **Non-beam, live:** normal transform HITL selection unchanged (no scope sent → original path).

## Transparency note for reviewers

The mixed-HITL exclusion (sibling LBV + a non-LBV HITL pending simultaneously) is covered by unit tests and closed in code, but did not naturally occur in the live runs — the IDE-side deferral independently covers the same case and was exercised live. It's defense-in-depth: either layer prevents a false green.

## Related

Coordinated with the IDE change (Beam to IDE feature) and the web orchestrator / runtime changes for the shared-job beam flow.

## License

### DIFF
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
@@ -47,6 +47,9 @@ const AtxGetJobDashboardCommand = 'aws/atxTransform/getJobDashboard'
 const AtxGetJobReportCommand = 'aws/atxTransform/getJobReport'
 const AtxUploadCustomPlanCommand = 'aws/atxTransform/uploadCustomPlan'
 const AtxLoadOlderWorklogsCommand = 'aws/atxTransform/loadOlderWorklogs'
+// Beam to IDE
+const AtxListBeamedReposCommand = 'aws/atxTransform/listBeamedRepos'
+const AtxDownloadBeamArtifactCommand = 'aws/atxTransform/downloadBeamArtifact'
 
 export const AtxNetTransformServerToken =
     (): Server =>
@@ -170,12 +173,13 @@ export const AtxNetTransformServerToken =
                         return await atxTransformHandler.uploadPackages(request)
                     }
                     case AtxSendMessageCommand: {
-                        const { workspaceId, jobId, text, skipPolling } = params as any
+                        const { workspaceId, jobId, text, skipPolling, beamStepId } = params as any
                         return await atxTransformHandler.sendMessage({
                             workspaceId,
                             jobId,
                             text,
                             skipPolling,
+                            beamStepId,
                         })
                     }
                     case AtxListMessagesCommand: {
@@ -269,6 +273,32 @@ export const AtxNetTransformServerToken =
                             request.nextToken
                         )
                     }
+                    // ---- Beam to IDE ----
+                    case AtxListBeamedReposCommand: {
+                        // PascalCase params/return to match the C# ExecuteCommandParams convention.
+                        const { WorkspaceId, ParentJobId } = params as any
+                        if (!WorkspaceId || !ParentJobId) {
+                            throw new Error('WorkspaceId and ParentJobId are required for listBeamedRepos')
+                        }
+                        const repos = await atxTransformHandler.listBeamedRepos(WorkspaceId, ParentJobId)
+                        return { Repos: repos }
+                    }
+                    case AtxDownloadBeamArtifactCommand: {
+                        // PascalCase params to match the C# ExecuteCommandParams serialization
+                        // convention used by every other C#-backed ATX command (e.g. downloadArtifact).
+                        const { WorkspaceId, ParentJobId, BeamArtifactId, DestDir } = params as any
+                        if (!WorkspaceId || !ParentJobId || !BeamArtifactId || !DestDir) {
+                            throw new Error(
+                                'WorkspaceId, ParentJobId, BeamArtifactId and DestDir are required for downloadBeamArtifact'
+                            )
+                        }
+                        return await atxTransformHandler.downloadBeamArtifact(
+                            WorkspaceId,
+                            ParentJobId,
+                            BeamArtifactId,
+                            DestDir
+                        )
+                    }
                     default: {
                         throw new Error(`Unknown ATX FES command: ${params.command}`)
                     }
@@ -311,6 +341,8 @@ export const AtxNetTransformServerToken =
                             AtxGetJobReportCommand,
                             AtxUploadCustomPlanCommand,
                             AtxLoadOlderWorklogsCommand,
+                            AtxListBeamedReposCommand,
+                            AtxDownloadBeamArtifactCommand,
                         ],
                     },
                 },
--- a/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
+++ b/server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
@@ -530,6 +530,349 @@ export class ATXTransformHandler {
     }
 
     /**
+     * Beam to IDE — read the beam-map artifact from the parent web job and return
+     * the list of beamed repos. The web orchestrator uploads a JSON beam-map
+     * (category HITL_FROM_AGENT) listing repos the user chose to beam. We find it
+     * by listing the parent job's artifacts and picking the beam-map JSON.
+     */
+    async listBeamedRepos(workspaceId: string, parentJobId: string): Promise<any[]> {
+        try {
+            this.logging.log(`ATX: listBeamedRepos for parentJobId=${parentJobId}`)
+            if (!this.atxClient && !(await this.initializeAtxClient())) {
+                throw new Error('ATX client not initialized')
+            }
+
+            // Find the beam-map artifact under the parent job.
+            // NOTE: beam artifacts must be uploaded by the web orchestrator as
+            // CUSTOMER_OUTPUT — the FES client CategoryType enum does not expose
+            // HITL_FROM_AGENT as a listable filter value.
+            const command = new ListArtifactsCommand({
+                workspaceId: workspaceId,
+                jobFilter: {
+                    jobId: parentJobId,
+                    categoryType: 'CUSTOMER_OUTPUT',
+                },
+            })
+            await this.addAuthToCommand(command)
+            const result = (await this.atxClient!.send(command)) as any
+            const artifacts: any[] = result.artifacts || []
+
+            // DIAGNOSTIC: dump exactly what ListArtifacts returned so we can tell
+            // "artifact not in the list" from "filtered out by fileType/.json". Both
+            // otherwise produce the same silent 'no beam-map' verdict.
+            this.logging.log(
+                `[BEAM-PKG] listBeamedRepos: ListArtifacts returned ${artifacts.length} CUSTOMER_OUTPUT artifact(s) for job=${parentJobId}`
+            )
+            for (const a of artifacts) {
+                this.logging.log(
+                    `[BEAM-PKG]   artifact | id=${a.artifactId} fileType=${a.fileType} ` +
+                        `name=${a.name ?? a.fileName ?? ''} path=${a.fileMetadata?.path ?? ''} category=${a.categoryType ?? a.category ?? ''}`
+                )
+            }
+
+            // Enumerate transformed-code ZIP artifacts. Each transformed bundle has a path
+            // like "<repo>_<hash>/<repo>.zip"; the repo name is the path's first segment
+            // (minus the _<hash> suffix) and the artifactId is what the IDE downloads+extracts
+            // to open the solution. NOTE: in a multi-repo job, EVERY repo that transformed on
+            // web has such a zip — so this listing alone is NOT the set of *beamed* repos.
+            const ZIP_RE = /(^|\/)([^/]+?)_[0-9a-f]{6,}\/[^/]+\.zip$/i
+            const transformedZips = artifacts.filter(a => {
+                const p = (a.fileMetadata?.path || '').toLowerCase()
+                return p.endsWith('.zip')
+            })
+
+            // Build repoName -> {artifactId, path} from the transformed zips (deduped).
+            const zipByRepo = new Map<string, { artifactId: string; path: string }>()
+            for (const a of transformedZips) {
+                const path = a.fileMetadata?.path || ''
+                const m = path.match(ZIP_RE)
+                const repoName = m ? m[2] : (path.split('/')[0] || '').replace(/_[0-9a-f]{6,}$/i, '')
+                if (!repoName || !a.artifactId) continue
+                const key = repoName.toLowerCase()
+                if (!zipByRepo.has(key)) {
+                    zipByRepo.set(key, { artifactId: a.artifactId, path })
+                }
+            }
+
+            // Read the beam-map JSON. It is the AUTHORITATIVE list of which repos the user
+            // EXPLICITLY beamed (its repos[] holds only beamed repos, each with a stepId).
+            // We use it to FILTER the transformed zips down to just the beamed repos — NOT
+            // merely to enrich. Previously discovery returned every transformed zip and used
+            // the beam-map only for enrichment, so in a multi-repo job the IDE showed the
+            // Load button on repos that transformed but were never beamed.
+            // The beam-map artifact often has NO usable name/path/category in the ListArtifacts
+            // response (all empty — observed), so we cannot identify it by label. We instead
+            // scan candidate (non-.zip) artifacts and pick the one whose JSON has a `repos`
+            // array. CRITICAL: each download/parse is wrapped in its OWN try/catch — a single
+            // artifact that throws (or isn't JSON, e.g. an unrelated metadata.json / binary)
+            // must NOT abort the scan, or we miss the real beam-map that comes later in the list
+            // and wrongly fall back to zip-discovery (empty stepId). We scan ALL candidates and
+            // keep the LAST valid repos[] (beam-map is re-written on each beam; latest wins).
+            // Candidate beam-map artifacts = everything EXCEPT the transformed-source
+            // bundles. The beam-map is itself stored as a ZIP by FES (downloadJsonArtifact
+            // unzips + parses it), so we MUST NOT exclude by ".zip" extension — that would
+            // filter out the beam-map itself (the earlier bug: the beam-map 3b1f9c00 has a
+            // .zip path, so a plain !endsWith('.zip') filter skipped it entirely → fell back
+            // to zip-discovery with empty stepId). Instead, exclude only the transformed-
+            // source bundles, whose paths match "<repo>_<hash>/<repo>.zip" (ZIP_RE). Anything
+            // else (including an unnamed/empty-path zip = the beam-map) is a candidate.
+            let beamMapRepos: any[] | null = null
+            const candidates = artifacts.filter(a => {
+                const p = (a.fileMetadata?.path || '').toLowerCase()
+                return !ZIP_RE.test(p) // keep non-transformed-bundle artifacts (incl. the beam-map zip)
+            })
+            this.logging.log(
+                `[BEAM-PKG] beam-map scan: ${candidates.length} candidate artifact(s) (excluded transformed-source bundles)`
+            )
+            for (const a of candidates) {
+                if (!a.artifactId) continue
+                try {
+                    const map = await this.downloadJsonArtifact(workspaceId, parentJobId, a.artifactId)
+                    if (map && Array.isArray(map.repos)) {
+                        beamMapRepos = map.repos as any[]
+                        this.logging.log(
+                            `[BEAM-PKG] beam-map found (${beamMapRepos.length} repo(s)) artifact=${a.artifactId} — filtering discovery by beam-map membership`
+                        )
+                        // Do NOT break — keep scanning; a later artifact could be a newer beam-map.
+                        // (Cost is bounded: only empty-path/non-zip artifacts, a handful per job.)
+                    }
+                } catch (e) {
+                    // Non-fatal: this artifact wasn't the beam-map (not JSON, wrong shape, or
+                    // download hiccup). Keep scanning the rest.
+                    this.logging.log(`[BEAM-PKG] beam-map scan: skipped artifact ${a.artifactId}: ${String(e)}`)
+                }
+            }
+
+            const beamed: any[] = []
+
+            // Show-set: fetch the parent job's plan tree ONCE so we can stamp each beamed repo
+            // with whether its Local Build Verification is still open (the IDE shows only repos
+            // still awaiting LBV). Best-effort: if the plan can't be fetched, leave IsLbvOpen
+            // undefined and the IDE defaults it true (never hides a fresh transfer).
+            let planRoot: AtxPlanStep | null = null
+            try {
+                const plan = await this.getTransformationPlan(workspaceId, parentJobId, '', true)
+                planRoot = plan?.Root ?? null
+            } catch (e) {
+                this.logging.log(`[BEAM-PKG] listBeamedRepos: could not fetch plan for LBV-open status: ${String(e)}`)
+            }
+
+            if (beamMapRepos) {
+                // Beam-map present: include ONLY repos listed in it (the beamed ones). Pull the
+                // artifactId from the matching transformed zip; carry stepId/scenario/tfm from
+                // the beam-map entry.
+                for (const r of beamMapRepos) {
+                    const nr = this.normalizeBeamRepo(r)
+                    if (!nr.RepositoryName) continue
+                    const zip = zipByRepo.get(nr.RepositoryName.toLowerCase())
+                    // PREFER the transformed-source zip artifactId (stable + downloadable). The
+                    // beam-map's own BeamArtifactId is NOT reliable: web-orc re-writes the beam-map
+                    // each poll and the id it carries can point at a NON-source artifact with no
+                    // presigned download URL, which made the IDE's BeamArtifactId thrash and Load
+                    // fail with "No presigned URL for beam artifact". The source zip stays constant
+                    // and is always downloadable; fall back to the beam-map id only when there is
+                    // no matching transformed zip.
+                    const artifactId = zip?.artifactId || nr.BeamArtifactId || ''
+                    if (!artifactId) {
+                        this.logging.log(
+                            `[BEAM-PKG] beam-map repo '${nr.RepositoryName}' has no artifact (transformed zip or beam-map) — skipping`
+                        )
+                        continue
+                    }
+                    if (!zip?.artifactId && nr.BeamArtifactId) {
+                        this.logging.log(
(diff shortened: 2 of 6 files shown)


pranav-firake (approved):

finncas (approved):

pranav-firake (approved):

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -529,6 +529,349 @@ export class ATXTransformHandler {
...
+                this.logging.log(`[BEAM-PKG] listBeamedRepos: could not fetch plan for LBV-open status: ${String(e)}`)
+            }
+
+            if (beamMapRepos) {
+                // Beam-map present: include ONLY repos listed in it (the beamed ones). Pull the
+                // artifactId from the matching transformed zip; carry stepId/scenario/tfm from
+                // the beam-map entry.
+                for (const r of beamMapRepos) {
+                    const nr = this.normalizeBeamRepo(r)
+                    if (!nr.RepositoryName) continue
```
ashishrp-aws, line 736: **[BLOCKER] Regex injection / ReDoS risk**

`parentJobId` is interpolated directly into `new RegExp()` without escaping. If a job ID ever contains regex metacharacters (`.`, `*`, `+`, `(`, `)`, etc.), this will either throw, silently mismatch, or in pathological cases cause catastrophic backtracking.

```ts
const statusRe = new RegExp(`beam-status-${parentJobId}-(.+)\\.json$`, 'i')
```

**Fix:** escape `parentJobId` before interpolating, e.g.:
```ts
const escapedJobId = parentJobId.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
const statusRe = new RegExp(`beam-status-${escapedJobId}-(.+)\\.json$`, 'i')
```

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -529,6 +529,349 @@ export class ATXTransformHandler {
...
+                const path = a.fileMetadata?.path || ''
+                const m = path.match(ZIP_RE)
+                const repoName = m ? m[2] : (path.split('/')[0] || '').replace(/_[0-9a-f]{6,}$/i, '')
+                if (!repoName || !a.artifactId) continue
+                const key = repoName.toLowerCase()
+                if (!zipByRepo.has(key)) {
+                    zipByRepo.set(key, { artifactId: a.artifactId, path })
+                }
+            }
+
```
ashishrp-aws, line 633: **[BLOCKER] Unvalidated downloaded JSON drives downstream artifact selection**

`downloadJsonArtifact` result is used directly as `beamMapRepos` with zero schema validation. This content comes from an artifact uploaded by the web orchestrator — if it's malformed, unexpectedly shaped, or (in a compromised/misbehaving orchestrator scenario) attacker-influenced, it flows straight into `normalizeBeamRepo` and drives which artifact IDs get downloaded and which paths get written to.

**Fix:** validate the parsed JSON has the expected shape (e.g. `repos` is an array of objects with string fields) before using it, and reject/log-and-skip anything that doesn't conform.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -529,6 +529,349 @@ export class ATXTransformHandler {
         }
     }
 
+    /**
```
ashishrp-aws, line 569: **[BLOCKER] Test coverage claim contradicted by CI**

The PR description states the mixed-HITL exclusion case is "covered by unit tests," but:
- A code search of this repo found **zero** test files referencing `listBeamedRepos` or any other new function added here.
- Codecov's own bot comment on this PR reports **26.13% patch coverage** (551 of 694 new lines uncovered) and overall project coverage regressed **-1.35%**.

This function (`listBeamedRepos`) and the scoped-HITL selection logic below are exactly the kind of branchy, edge-case-heavy code that most needs test coverage, and currently has none in this diff.

**Ask:** either add unit tests actually exercising these paths (empty beam-map, beam-map present, beam-status fallback, mixed sibling/self LBV, regex edge cases) or correct the PR description.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -529,6 +529,349 @@ export class ATXTransformHandler {
...
+                dl.s3PresignedUrl,
+                dl.requestHeaders,
+                destDir,
+                'beam-artifact.zip',
+                this.logging
+            )
+            this.logging.log(`[BEAM-PKG] downloadBeamArtifact extracted -> ${extractedPath}`)
+            return { Success: true, ExtractedPath: extractedPath }
+        } catch (error) {
+            this.logging.error(`[BEAM-PKG] downloadBeamArtifact error: ${String(error)}`)
```
ashishrp-aws, line 975: **[BLOCKER] Possible zip-slip via `downloadAndExtractArchive`**

This extracts a remotely-sourced zip (from a presigned URL controlled by artifact content, ultimately originating from the web orchestrator) into `destDir`, which itself is caller/IDE-supplied with only a truthiness check in `atxNetTransformServer.ts` (no path validation).

If `Utils.downloadAndExtractArchive` doesn't canonicalize/validate zip entry paths against `destDir` (reject `../` traversal, absolute paths, symlinks), a malicious or corrupted archive could write outside the intended extraction directory.

**Fix:** confirm `Utils.downloadAndExtractArchive` guards against zip-slip; if not, add entry-path validation there. Also validate `destDir` resolves within the expected workspace root before extracting.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -798,48 +1141,91 @@ export class ATXTransformHandler {
...
 
-                const normalizedHeaders: Record<string, string> = {}
-                if (result.requestHeaders) {
-                    for (const [key, value] of Object.entries(result.requestHeaders)) {
-                        normalizedHeaders[key] = Array.isArray(value) ? value[0] : value
+                    return {
+                        s3PresignedUrl: result.s3PreSignedUrl,
+                        requestHeaders: normalizedHeaders,
                     }
+                } else {
```
ashishrp-aws, line 1299: **[MAJOR] Retry backoff has no jitter**

Fixed exponential backoff (250/500/1000ms) with no randomization. Under concurrent throttling (multiple LSP instances or parallel calls hitting FES at once), all callers will retry in lockstep, amplifying the thundering-herd effect that caused the throttling in the first place.

**Fix:** add jitter, e.g. `backoffMs * (0.5 + Math.random())`.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -529,6 +529,349 @@ export class ATXTransformHandler {
...
     }
 
+    /**
+     * Beam to IDE — read the beam-map artifact from the parent web job and return
+     * the list of beamed repos. The web orchestrator uploads a JSON beam-map
+     * (category HITL_FROM_AGENT) listing repos the user chose to beam. We find it
+     * by listing the parent job's artifacts and picking the beam-map JSON.
+     */
+    async listBeamedRepos(workspaceId: string, parentJobId: string): Promise<any[]> {
+        try {
```
ashishrp-aws, line 576: **[MAJOR] Pervasive `any` typing defeats compile-time safety**

`listBeamedRepos` returns `Promise<any[]>`, and internally `artifacts: any[]`, `beamMapRepos: any[] | null`, `beamed: any[]` are all untyped. Combined with `normalizeBeamRepo`'s tolerant `??` chains reading arbitrary keys off `any`, there's no compiler check that the shape callers (the IDE via `atxNetTransformServer.ts`) expect actually matches what's produced.

**Fix:** define interfaces for the FES artifact shape, the beam-map JSON shape, and the returned `BeamedRepoInfo` type, and use them instead of `any` throughout this function and `normalizeBeamRepo`.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -529,6 +529,349 @@ export class ATXTransformHandler {
...
+                if (!zipByRepo.has(key)) {
+                    zipByRepo.set(key, { artifactId: a.artifactId, path })
+                }
+            }
+
+            // Read the beam-map JSON. It is the AUTHORITATIVE list of which repos the user
+            // EXPLICITLY beamed (its repos[] holds only beamed repos, each with a stepId).
+            // We use it to FILTER the transformed zips down to just the beamed repos — NOT
+            // merely to enrich. Previously discovery returned every transformed zip and used
+            // the beam-map only for enrichment, so in a multi-repo job the IDE showed the
```
ashishrp-aws, line 638: **[MAJOR] Sequential, unbounded artifact downloads with no timeout**

Each candidate artifact is downloaded and parsed one at a time, awaited in a `for` loop, with no cap on `candidates.length` and no overall timeout. If a job has many non-zip artifacts (metadata files, logs, etc.), this call could take a long time serially, blocking the response to `listBeamedRepos` (and thus the IDE's discovery UI).

**Fix:** cap the number of candidates scanned (e.g. stop after N or after finding a valid beam-map), and/or parallelize with `Promise.all` plus an overall timeout.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -1090,10 +1478,58 @@ export class ATXTransformHandler {
...
+            // spelling (stepId/planStepId/parentStepId), same as the forwarding path.
+            const scopeStepIds = (beamScopeStepIds || '')
+                .split(',')
+                .map(s => s.trim())
+                .filter(s => s.length > 0)
+            const lbvHitls = hitls.filter(h => h.tag === 'local-build-verification')
+            let scopedLbv: any = null
+            if (scopeStepIds.length > 0 && lbvHitls.length > 0) {
+                scopedLbv = lbvHitls.find(h =>
+                    scopeStepIds.includes(String(h.stepId ?? h.planStepId ?? h.parentStepId ?? ''))
```
ashishrp-aws: **[MAJOR] Scoped-HITL selection logic duplicated between here and the EXECUTING path**

This `scopedLbv`/`fallbackPool` selection block is near-verbatim duplicated below in `_getTransformInfoInternal`'s EXECUTING branch (`execScopeStepIds`/`scopedExecLbv`/`execFallbackPool`). Any future fix to the selection rule (e.g. a new HITL tag, a different tie-break) has to be applied in both places, and it's easy to update one and miss the other.

**Fix:** extract a shared helper, e.g. `selectScopedHitl(hitls: Hitl[], scopeStepIds: string[]): { selected: Hitl | null; allOutOfScope: boolean }`, and call it from both sites.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -1335,15 +1782,87 @@ export class ATXTransformHandler {
...
+                    // path too since a beam job surfaces its LBV HITL here.
+                    const execScopeStepIds = (request.beamScopeStepIds || '')
+                        .split(',')
+                        .map(s => s.trim())
+                        .filter(s => s.length > 0)
+                    const scopedExecLbv =
+                        execScopeStepIds.length > 0
+                            ? blockingHitls.find(
+                                  h =>
+                                      h.tag === 'local-build-verification' &&
```
ashishrp-aws: **[MAJOR] This is the second copy of the scoped-HITL selection logic**

See the corresponding comment on `getHitlAgentArtifact` above — this block (`execScopeStepIds`/`scopedExecLbv`/`execFallbackPool`) is a near-duplicate of that logic. Extracting a shared `selectScopedHitl` helper would let both call sites stay in sync and reduce this method's size (this file is already ~4800 lines).

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
```diff
@@ -269,6 +273,32 @@ export const AtxNetTransformServerToken =
...
+                    case AtxDownloadBeamArtifactCommand: {
+                        // PascalCase params to match the C# ExecuteCommandParams serialization
+                        // convention used by every other C#-backed ATX command (e.g. downloadArtifact).
+                        const { WorkspaceId, ParentJobId, BeamArtifactId, DestDir } = params as any
+                        if (!WorkspaceId || !ParentJobId || !BeamArtifactId || !DestDir) {
+                            throw new Error(
+                                'WorkspaceId, ParentJobId, BeamArtifactId and DestDir are required for downloadBeamArtifact'
+                            )
+                        }
+                        return await atxTransformHandler.downloadBeamArtifact(
```
ashishrp-aws, line 295: **[MAJOR] `DestDir` is only truthiness-checked, not path-validated**

`WorkspaceId`, `ParentJobId`, `BeamArtifactId`, `DestDir` are all checked only for presence (`!DestDir`), not for whether `DestDir` is a sane path within the expected workspace root. This feeds directly into `downloadBeamArtifact`'s zip extraction (see corresponding comment in `atxTransformHandler.ts`).

**Fix:** validate `DestDir` resolves within the workspace root before passing it through.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -529,6 +529,349 @@ export class ATXTransformHandler {
...
+            // response (all empty — observed), so we cannot identify it by label. We instead
+            // scan candidate (non-.zip) artifacts and pick the one whose JSON has a `repos`
+            // array. CRITICAL: each download/parse is wrapped in its OWN try/catch — a single
+            // artifact that throws (or isn't JSON, e.g. an unrelated metadata.json / binary)
+            // must NOT abort the scan, or we miss the real beam-map that comes later in the list
+            // and wrongly fall back to zip-discovery (empty stepId). We scan ALL candidates and
+            // keep the LAST valid repos[] (beam-map is re-written on each beam; latest wins).
+            // Candidate beam-map artifacts = everything EXCEPT the transformed-source
+            // bundles. The beam-map is itself stored as a ZIP by FES (downloadJsonArtifact
+            // unzips + parses it), so we MUST NOT exclude by ".zip" extension — that would
```
ashishrp-aws, line 650: **[MAJOR] Silent error swallowing makes failure modes indistinguishable**

Errors here (and in several other catch blocks added by this PR — `downloadJsonArtifact`, the outer `listBeamedRepos` catch) are logged and treated as "not this one, try the next" or "return empty." This means a transient network failure looks identical to "this artifact genuinely isn't the beam-map," and "FES is down" looks identical to "job was never beamed" (0 repos returned either way).

**Fix:** at minimum, track whether any candidate download failed vs. cleanly didn't match, and surface that distinction to the caller (e.g. `{ repos: [], hadErrors: true }`) so callers/observability can tell "confirmed not beamed" from "couldn't determine."

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
ashishrp-aws, line 569: **[MAJOR] This adds a second, largely independent responsibility to an already very large class**

`ATXTransformHandler` was already sizable before this PR; this diff adds ~700 lines of "Beam to IDE" logic (discovery, artifact download, LBV-open detection, plan-tree node matching) that has minimal coupling to the rest of the class beyond a few shared private helpers (`createArtifactDownloadUrl`, `getTransformationPlan`). This is a good candidate for extraction into a dedicated `BeamRepoService`/`atxBeamHandler.ts`, both for testability (unit tests could construct it with mocked dependencies instead of the whole handler) and to keep this file's growth in check.

Not necessarily a blocker for this PR, but worth tracking as an immediate follow-up given the class is now pushing 4800+ lines.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
```diff
@@ -529,6 +529,349 @@ export class ATXTransformHandler {
...
+                } catch (e) {
+                    // Non-fatal: this artifact wasn't the beam-map (not JSON, wrong shape, or
+                    // download hiccup). Keep scanning the rest.
+                    this.logging.log(`[BEAM-PKG] beam-map scan: skipped artifact ${a.artifactId}: ${String(e)}`)
+                }
+            }
+
+            const beamed: any[] = []
+
+            // Show-set: fetch the parent job's plan tree ONCE so we can stamp each beamed repo
```
ashishrp-aws, line 718: **[MAJOR] Unversioned, unstable contract with the web orchestrator**

`normalizeBeamRepo` tolerates 4 different spellings for each field (`repoName`/`repositoryName`/`repo`/`name`, `stepId`/`planStepId`/`parentStepId`, etc.), and the same `stepId ?? planStepId ?? parentStepId` pattern is repeated at least 4 more times elsewhere in this file for HITL objects. This signals the actual wire contract with the web orchestrator/FES is not pinned down — if it changes again, there's no single place to update, and no way to detect at compile time that a field was renamed upstream.

**Fix:** centralize this into one `getStepId(obj: unknown): string | undefined` helper (and similarly for repo-name fields), and raise this with the web orchestrator team as something worth a versioned schema.

ashishrp-aws (changes requested):
Multi-persona review (Senior SDE, Principal SDE, Security, QA) surfaced several blocker and major issues, posted inline at the relevant lines below.

**Summary of blockers:**
- PR description claims the mixed-HITL case is "covered by unit tests," but Codecov shows 26.13% patch coverage (551/694 lines uncovered) and no test file references any function added here.
- Regex injection risk: `parentJobId` interpolated unescaped into `new RegExp(...)`.
- Downloaded beam-map JSON used without schema validation.
- Possible zip-slip in beam artifact extraction to a caller-supplied `destDir`.

**Summary of majors:** duplicated HITL-scoping logic between `getHitlAgentArtifact` and the EXECUTING path, pervasive `any` typing, unbounded sequential artifact downloads with no timeout, no backoff jitter, silent error swallowing, an unversioned field-name-tolerant contract with the web orchestrator, and general god-class growth of `ATXTransformHandler`.

The core feature intent (preventing a "false green" on multi-repo beam LBV) is sound — these are implementation/robustness/test-coverage concerns to address before merge.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed. `parentJobId` is now escaped before interpolation: `parentJobId.replace(/[.*+?^${}()|[\]\]/g, '\$&')`. Test: `listBeamedRepos › does not throw or mismatch when the jobId contains regex metacharacters`.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed. `map.repos` is shape-validated before use — only entries that are objects yielding a non-empty repo name are kept; malformed / `null` / non-object entries are rejected and logged. Test: `listBeamedRepos › rejects malformed beam-map entries`.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed. Added a `Beam to IDE` unit suite: `getStepId`, `normalizeBeamRepo`, and `listBeamedRepos` (empty / beam-map / beam-status fallback / hash-suffix / regex-edge / send-failure), plus `downloadBeamArtifact` destDir rejection and a zip-slip regression test in `utils.test.ts`. Also corrected the PR description's coverage claim.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed at the extraction util: `extractAllEntriesTo` resolves each entry against the extraction root and skips + logs any that escape it (`../` / absolute). Added defensive `destDir` validation in `downloadBeamArtifact` (rejects relative / `..` / empty). Tests: `utils › should block zip-slip entries that escape the extraction directory` and `downloadBeamArtifact › rejects a destDir containing ".." traversal`.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed. Backoff is now `base * (0.5 + Math.random())` over the 250/500/1000ms base.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed. Added `BeamMapRepo` and `BeamedRepoInfo` interfaces; `listBeamedRepos` returns `Promise<BeamedRepoInfo[]>` and `normalizeBeamRepo` is typed against them. No `any` remains in the beam discovery path.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed. The candidate scan is capped (`BEAM_MAP_SCAN_MAX_CANDIDATES = 25`) and bounded by an overall deadline (`BEAM_MAP_SCAN_BUDGET_MS = 15000`); the loop breaks when either is reached.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed. Extracted a shared `selectScopedLbvHitl(hitls, scopeStepIds)` returning `{ scopedLbv, allOutOfScopeLbv }`; `getHitlAgentArtifact` now calls it.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed. The EXECUTING path now calls the same `selectScopedLbvHitl` helper, so the selection rule lives in one place.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxNetTransformServer.ts
Rajanna-Karthik: Fixed. `downloadBeamArtifact` rejects a non-absolute, `..`-containing, or empty `destDir` before extraction, complementing the entry-level zip-slip guard.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Agreed this is worth doing, but returning `{ repos, hadErrors }` changes `listBeamedRepos`'s return shape, which the C# IDE client consumes over the wire — it needs a coordinated IDE-side change and can't land safely in this PR. Tracking as a fast-follow.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Agreed — tracking as the immediate follow-up. Holding the ~700-line extraction out of this PR to avoid pre-release regression risk on the shared handler; will do it as a dedicated refactor.

server/aws-lsp-codewhisperer/src/language-server/netTransform/atxTransformHandler.ts
Rajanna-Karthik: Fixed the code half: centralized the step-id coalescing into a single `getStepId(obj)` helper across all call sites, and the shapes are now typed via `BeamMapRepo` / `BeamedRepoInfo`. Left a TODO to pin a versioned schema with the web-orchestrator team, tracked as a follow-up.

Rajanna-Karthik:
Thanks for the thorough review. Addressed all 4 blockers plus 5 majors in code — regex escaping, beam-map JSON validation, zip-slip guard + destDir validation, backoff jitter, typed interfaces (no `any` in the beam path), bounded candidate scan, shared `selectScopedLbvHitl` helper, and the `getStepId` centralization — each with unit tests. Pushed as `4d00fff2`.

Two majors deferred as tracked follow-ups: the `{ repos, hadErrors }` return-shape change (needs a coordinated IDE-client change) and the `BeamRepoService` extraction (~700-line refactor, non-blocking per your note). Details in the per-thread replies above.

ashishrp-aws (approved):

